PR #11, opened 2018-11-30
feat: Add annoated tags by default

HazAT:
Sorry for the reformat of readme and changelog, some autoformatter 🤷‍♂️ 

This change still needs to be tested in the wild, but it should create annotated tags from now on.


tonyo:
@HazAT Thanks for kicking this off  💟

PR #13, opened 2018-12-27
feat: Environment files

tonyo:
`craft` will now look into `~/.craft.env` and `$(pwd)/.craft.env` files to read additional environment variables.


src/config.ts
```diff
@@ -164,3 +180,65 @@ export function getGitTagPrefix(): string {
...
+    );
+  } else {
+    logger.debug(
+      `No environment file found in the home directory: ${homedirEnvFile}`
+    );
+  }
+
+  // Read from current dir
+  const currentDirEnvFile = join(process.cwd(), ENV_FILE_NAME);
+  if (existsSync(currentDirEnvFile)) {
```
jan-auer: This check is prone to race conditions. You could probably just invoke `nvar` and then catch down the expected "file doesn't exist" exception. Also, this would allow to explicitly handle the case when the file is not readable.

tonyo: TBH, I don't really want to worry too much about what can go wrong here (file not readable, it's a directory, etc.)

PR #15, opened 2019-02-11
Allow empty preReleaseCommand

HazAT:
This will no longer print the error
```
⚠ warn Not running the pre-release command: no command specified
✖ error Error: Nothing to commit: has the pre-release command done its job?
```
if the `preReleaseCommand` is defined as empty string.


tonyo:
@HazAT ❤️

PR #18, opened 2019-03-14
feat(crates): Clone with submodules before publishing

jan-auer:
Instead of downloading a source tarball from GitHub, clone the release branch with submodules instead. This allows to publish crates like `symbolic` that depend on shipping code from submodules.

Noteworthy points:
 - This performs a **shallow clone** (`--depth=1`) to reduce the time this takes.
 - There is no straight-forward option to directly shallow clone an arbitrary commit SHA, so instead this clones the release branch. Since the branch is checked against Zeus immediately beforehand, we can assume that the branch head revision matches.
 - The clone is performed with `--recurse`, which initializes submodules. If the repository does not specify shallow submodules, this might take quite a while, but that is out of scope of this PR. We could pass `--shalllow-submodules`, but that comes with a bunch of other implications, so I'd rather keep that simple. `symbolic` only has shallow references, for instance.

### DIFF
--- a/src/targets/crates.ts
+++ b/src/targets/crates.ts
@@ -197,22 +197,24 @@ export class CratesTarget extends BaseTarget {
    * Clones a repository and its submodules.
    *
    * @param config Git configuration specifying the repository to clone.
-   * @param version The version specifying the release branch to clone.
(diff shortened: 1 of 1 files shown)


tonyo:
Thank you, mister

PR #20, opened 2019-03-22
feat/fix: Add tag next for npm prereleases

HazAT:
We published `@sentry/*` with `5.0.0-beta1` & `5.0.0-beta.2` to the npm registry.
Apparently, npm needs a tag with `--tag=next` to not make it `latest` which means it will be the default when doing `npm install @sentry/browser` for example.

Without `--tag=next` you would get the highest version, in our case `5.0.0-beta.2`.
We actually never ever want this to happen (imho) that's why I added this check to make sure we always add `--tag=next` in case it's a pre-release version.

We had to unpublish both version from npm 😞


HazAT:
image

It works, I let @tonyo merge it if he is fine with it.

tonyo:
Oopsie 😊 
Thanks @HazAT !

PR #36, opened 2020-01-23
feat: Optional artifact check

HazAT:
If `statusProvider` is github, we do not by default ask Zeus for artifacts.
Also for each target, if `includeNames` is not defined we skip artifacts check on Zeus as well.

### DIFF
--- a/src/commands/publish.ts
+++ b/src/commands/publish.ts
@@ -202,12 +202,10 @@ async function publishToTargets(
  *
  * @param zeus Zeus store object
  * @param revision Git revision SHA
- * @param targetNames A list of target names to publish
  */
 async function printRevisionSummary(
   zeus: ZeusStore,
-  revision: string,
-  targetNames: string[]
+  revision: string
 ): Promise<void> {
   const artifacts = await zeus.listArtifactsForRevision(revision);
   if (artifacts.length > 0) {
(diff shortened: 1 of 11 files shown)


src/targets/base.ts
```diff
@@ -62,6 +62,12 @@ export class BaseTarget {
       ...defaultFilterOptions,
       ...this.filterOptions,
     };
+    if (!filterOptions.includeNames) {
```
tonyo: I'm pretty sure this will break some repos, because `includeNames` is usually undefined.
We can probably do a `filterOptions.includeNames ===  []` comparison, so `[]` in `includeNames` can have an explicit meaning of "do not include anything".

src/commands/publish.ts
```diff
@@ -505,13 +497,20 @@ export async function publishMain(argv: PublishOptions): Promise<any> {
...
-    await printRevisionSummary(
-      zeus,
-      revision,
-      targetConfigList.map(t => t.name || '__undefined__')
-    );
+    // TODO: Handle new Artifacts stores
+    // We only ask Zeus now for artifacts
+    if (statusProviderName === StatusProviderName.Zeus) {
+      await printRevisionSummary(zeus, revision);
+      await checkRequiredArtifacts(zeus, revision, config.requireNames);
```
tonyo: Just a note: we can do this for now, but in the future we want to allow arbitrary statusProvider/artifactProvider combinations.

PR #43, opened 2020-02-05
feat: Support for custom remote names

rhcarvalho:
This enables users to use craft to prepare releases targeting a remote
different than origin. For example, some people clone repositories such
that the remote 'origin' points to a fork and the remote 'upstream'
points to the upstream project repository.

Note that the string 'origin' is also found in ghPages.ts and
registry.ts, but those usages refer to a remote in a temporary clone
owned by craft and thus takes no customization.

### DIFF
--- a/src/commands/prepare.ts
+++ b/src/commands/prepare.ts
@@ -126,9 +126,10 @@ async function createReleaseBranch(
   }
   if (branchHead) {
     let errorMsg = `Branch already exists: ${branchName}. `;
+    const remoteName = getRemoteName();
     errorMsg +=
       'Run the following commands to delete the branch, and then rerun "prepare":\n';
-    errorMsg += `    git branch -D ${branchName}; git push origin --delete ${branchName}\n`;
+    errorMsg += `    git branch -D ${branchName}; git push ${remoteName} --delete ${branchName}\n`;
     reportError(errorMsg, logger);
   }
 
@@ -153,11 +154,12 @@ async function pushReleaseBranch(
   branchName: string,
   pushFlag: boolean = true
 ): Promise<any> {
+  const remoteName = getRemoteName();
   if (pushFlag) {
     logger.info(`Pushing the release branch "${branchName}"...`);
     // TODO check remote somehow
     if (shouldPerform()) {
-      await git.push('origin', branchName, { '--set-upstream': true });
+      await git.push(remoteName, branchName, { '--set-upstream': true });
     } else {
       logger.info('[dry-run] Not pushing the release branch.');
     }
@@ -165,7 +167,7 @@ async function pushReleaseBranch(
     logger.info('Not pushing the release branch.');
     logger.info(
       'You can push this branch later using the following command:',
-      `  $ git push -u origin "${branchName}"`
+      `  $ git push -u ${remoteName} "${branchName}"`
     );
   }
 }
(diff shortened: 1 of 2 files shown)


rhcarvalho:
Note: was a pity that prepare and publish get the remote name differently, open for suggestions. Otherwise this is the ugly solution that works.

rhcarvalho:
Applied linter suggestion `let` -> `const`.

src/commands/prepare.ts
```diff
@@ -126,9 +126,10 @@ async function createReleaseBranch(
   }
   if (branchHead) {
     let errorMsg = `Branch already exists: ${branchName}. `;
+    const remoteName = process.env.CRAFT_REMOTE || 'origin';
```
tonyo: Could we perhaps move this get-or-default logic to a separate function? Otherwise it's duplicated across different files.

rhcarvalho: Within the same file ok, updated; among the two files I didn't see a clear way forward, suggestions welcome.

rhcarvalho:
Linter asking for docs, I'm not sure if it is worth it in this case... See how it is for `getGithubApiToken`: 

```javascript
/**
 * Gets GitHub API token from environment
 *
 * @returns Github authentication token if found
 */
```

What's the value? All that repetitive text for a oneliner :confused:

src/utils/githubApi.ts
```diff
@@ -62,11 +64,13 @@ export class GithubRemote {
   public constructor(
     owner: string,
     repo: string,
+    name?: string,
     username?: string,
     apiToken?: string
   ) {
     this.owner = owner;
     this.repo = repo;
+    this.name = name || 'origin';
```
rhcarvalho: We could have `'origin'` be a constant somewhere... but then I don't understand why `prepare.ts` and `publish.ts` doesn't use the same facilities to use git?

src/commands/prepare.ts
```diff
@@ -504,6 +506,15 @@ export async function releaseMain(argv: ReleaseOptions): Promise<any> {
   await switchToDefaultBranch(git, defaultBranch);
 }
 
+/**
+ * Gets Git remote name to use
+ *
+ * @returns Git remote name from environment or default
+ */
+function getRemoteName(): string {
```
tonyo: I was thinking of moving this to `utils/githubApi.ts`, so it can be reused by all the different files.

src/utils/githubApi.ts
tonyo: They both use `utils/githubApi`, or wdym?

rhcarvalho: I enumerated the files incorrectly, sorry.

Second try:

- `src/targets/ghPages.ts` and `src/targets/registry.ts` both use `GithubRemote` from `src/utils/githubApi.ts`
- `src/commands/prepare.ts` doesn't use `GithubRemote`

---

IMHO the constructor here should not read nor otherwise depend on `process.env`. And since there is a use of this class that doesn't provide all arguments to the constructor, we need a default value.

https://github.com/getsentry/craft/blob/21ec582914e08ae8eb1f9b1aa812d28d1d7eac20/src/targets/registry.ts#L36-L39

src/commands/prepare.ts
rhcarvalho: This does `process.env.CRAFT_REMOTE || 'origin'` and in `GithubRemote` we need `name || 'origin'`.
This is looking overcomplicated in my head, would you like to give it a shot with code?

src/utils/githubApi.ts
rhcarvalho: Talked offline with @tonyo, realized this was a mistake since this class is only used for repo clones "owned" by craft in a temp dir.

src/commands/prepare.ts
rhcarvalho: The updated patch only touches `prepare.ts`. @tonyo and I decided it is of little value to create a const = 'origin` somewhere just to import it in two places.

Therefore this small func stays here and is used only in this file.

rhcarvalho:
Talked offline with @tonyo, realized that the original patch was unnecessarily changing remote names for temporary clones made by craft during the publish step. Code (and commit message) updated.

PR #52, opened 2020-02-20
feat: Artifact provider abstraction

tonyo:
Includes the first implementation of the Zeus artifact provider.

Adds `craft artifacts` command group with the following commands:
* list
* download
* upload (TODO)

Some code is currently duplicated in zeus store and zeus artifact provider, but the store will be eventually removed.


tonyo:
Gonna merge it, shouldn't break anything (yet :)).

PR #57, opened 2020-02-25
ref(docs): Clarify artifact provider docstrings

lobsterkatie:
This makes explicit things which weren't obvious to me, as a newcomer to the codebase. Specifically:

1) I tried to differentiate between `XXX` methods, which rely on caching when they can, and `doXXX` methods, which always get their data from the provider.

2) I made sure every docstring included all parameters and the return value (if there was one).

3) Where paths are local, I tried to call them out as such.

4) When possible, I let subclass implementations of abstract methods inherit their docstrings from the base class.

### DIFF
--- a/src/artifact_providers/base.ts
+++ b/src/artifact_providers/base.ts
@@ -136,7 +136,7 @@ export abstract class BaseArtifactProvider {
   ): Promise<string>;
 
   /**
-   * Given an arry of artifacts, returns an arry of local paths to the those
+   * Given an array of artifacts, returns an array of local paths to those
    * artifacts, downloading (and then caching) each file first if necessary.
    *
    * The cache persists for the lifetime of the ArtifactProvider instance, so
--- a/src/artifact_providers/base.ts
+++ b/src/artifact_providers/base.ts
@@ -38,7 +38,7 @@ export interface FilterOptions {
  * Base interface for artifact providers.
  */
 export abstract class BaseArtifactProvider {
-  /** URL cache for downloaded files */
+  /** Cache for local paths to downloaded files */
   protected readonly downloadCache: {
     [key: string]: Promise<string> | undefined;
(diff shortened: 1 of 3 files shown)


src/artifact_providers/base.ts
```diff
@@ -118,15 +121,31 @@ export abstract class BaseArtifactProvider {
...
+   * @returns Absolute path to the saved file
    */
   protected abstract async doDownloadArtifact(
     artifact: CraftArtifact,
     downloadDirectory: string
   ): Promise<string>;
 
   /**
-   * Downloads multiple artifacts to the given directory
+   * Given an arry of artifacts, returns an arry of local paths to the those
```
tonyo:
```suggestion
   * Given an array of artifacts, returns an array of local paths to those
```

src/artifact_providers/zeus.ts
```diff
@@ -34,14 +34,7 @@ export class ZeusArtifactProvider extends BaseArtifactProvider {
...
   /**
-   * Downloads the given artifact file.
-   *
-   * Downloaded URL are cached during the instance's lifetime, so the same
-   * file is downloaded only once.
-   *
-   * @param artifact An artifact object to download
-   * @param downloadDirectory Directory where downloaded artifact is stored
-   * @returns Absolute path to the saved file
+   * @inheritDoc
```
tonyo: Nice, didn't know about this one 👍

src/artifact_providers/base.ts
lobsterkatie: Damn, and I proofread and everything. Thanks for catching!

src/artifact_providers/zeus.ts
lobsterkatie: You can thank your friends Kamil and Daniel for that one :-)

image

PR #60, opened 2020-03-10
ref(config): Create environment utils module

lobsterkatie:
With the introduction of the CGS artifact provider, we now have need for a function like [this](https://github.com/getsentry/craft/blob/2f623facdf151fde646e7746e93a8f6b3e902a60/src/commands/publish.ts#L100-L118) in multiple places, so this PR factors it out into its own module and makes it more generic. (It also moves all other env-related stuff to the new module.)

~I've marked this WIP until I can resolve two outstanding questions:~

~1) I've included unit tests for the function itself, but not any integration/acceptance/etc tests. So, in particular, though I've manually tested the fact that this doesn't break publishing (in other words, that that the generic function with specific parameters correctly mimics the behavior of the single-use function it's replacing), I have not written tests which confirm this. Not sure if these should live in a new `craft/src/commands/__tests__/publish.test.ts ` file or elsewhere. (I didn't find integration-y tests anywhere else, but please point me to them if I missed them.)~

~2) The line between what's env and what's config feels pretty blurry, but I wonder if there's anything else that ought to live under the heading of "env," now that I've separated them.~

### DIFF
--- a/src/__tests__/config.test.ts
+++ b/src/__tests__/config.test.ts
@@ -1,29 +1,14 @@
 /**
- * Tests of our ability to read craft config files and env vars (NOT general
- * test configuration).
+ * Tests of our ability to read craft config files. (This is NOT general test
+ * configuration).
  */
 
-import { readFileSync, writeFileSync } from 'fs';
-import { homedir } from 'os';
+import { readFileSync } from 'fs';
 import { dirname, join } from 'path';
 
 import { compile } from 'json-schema-to-typescript';
 
-import {
-  CONFIG_FILE_NAME,
-  ENV_FILE_NAME,
-  getProjectConfigSchema,
-  readEnvironmentConfig,
-  validateConfiguration,
-} from '../config';
-import { withTempDir } from '../utils/files';
-
-jest.mock('os', () => ({ ...require.requireActual('os'), homedir: jest.fn() }));
-
-// Mock getConfigFileDir
-// tslint:disable-next-line:no-var-requires
-const configModule = require('../config');
-const getConfigFileDirMock = (configModule.getConfigFileDir = jest.fn());
+import { getProjectConfigSchema, validateConfiguration } from '../config';
 
 const configSchemaDir = join(dirname(__dirname), 'schemas');
 const configGeneratedTypes = join(configSchemaDir, 'project_config.ts');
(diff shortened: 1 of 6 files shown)


tonyo:
> I've included unit tests for the function itself, but not any integration/acceptance/etc tests. 

There are no real integration tests (yet) :( 
Not saying it's impossible, just wasn't super straightforward to me how to approach this.

> The line between what's env and what's config feels pretty blurry, but I wonder if there's anything else that ought to live under the heading of "env," now that I've separated them.

So to be super specific here, we have:
* YAML configuration (`.craft.yml`) -- there we keep project-specific options, i.e. those that define what to do for the specific project.
* Environment variables -- user-specific and env-specific options, runtime configuration (e.g. secrets)
* Env files -- these are a just another way to prepare the environment, so I don't think we should handle them separately anywhere. We just assume that those were loaded in the environment early enough.

src/utils/env.ts
```diff
@@ -0,0 +1,80 @@
...
+ * Checks the environment for a single variable, taking into account its legacy
+ * name, if app.
+ *
+ * Copies value over to live under current name if only found under legacy name.
+ *
+ * @param envVar a RequiredConfigVar object to check for
+ * @returns true if variable was found, under either current or legacy name,
+ * false otherwise
+ */
+const envHasVar = (envVar: RequiredConfigVar): boolean => {
```
tonyo: Any specific reason why this is not a function?

lobsterkatie: 🤔 It is a function... ?

Or do you mean, why didn’t I write it as `function envHasVar(...) { ... }`?

lobsterkatie:
> There are no real integration tests (yet) :(
> Not saying it's impossible, just wasn't super straightforward to me how to approach this.

Fair. For simplicity, then, I'm going to leave that question for another PR to solve.

src/utils/env.ts
tonyo: Yep, I'm curious why this specific function is not a `function (...)`.
Is it because it's not exported?

lobsterkatie:
> So to be super specific here, we have:
> 
> * YAML configuration (`.craft.yml`) -- there we keep project-specific options, i.e. those that define what to do for the specific project.
> * Environment variables -- user-specific and env-specific options, runtime configuration (e.g. secrets)
> * Env files -- these are a just another way to prepare the environment, so I don't think we should handle them separately anywhere. We just assume that those were loaded in the environment early enough.

Yup - good with all that (but that does lay it out very clearly and helps me think about my own question more with more clarity, so thanks!). I was indeed talking about the difference between your first category and the second two, as a unit (so, as you say, project-specific config vs user/env-specific config). Right now those two different purposes have code which is intermingled in `config.ts`, and so my temptation is to take the env stuff and move it from `config.ts` to the new `env.ts`. Does that sound reasonable? (Would probably do it as a separate PR, though, just to keep things digestible.)

src/utils/env.ts
lobsterkatie: Nope - nothing other than muscle memory (though I can't promise I'm 100% consistent about that, even so). Happy to switch it back.

lobsterkatie: (I think with the advent of arrow functions I've also been trying to train myself out of using the `function` keyword in callbacks and such, and perhaps that gets over-applied.)

lobsterkatie: Oh, yeah - I see that I was't even consistent within this file. D'oh. Fixing.

src/commands/publish.ts
```diff
@@ -399,9 +375,15 @@ async function handleReleaseBranch(
...
   logger.debug('Argv:', JSON.stringify(argv));
   checkMinimalConfigVersion();
-  checkPrerequisites();
+  // TODO (kmclb): pull the names of the necessary env vars out of config once
+  // there are more options than just Zeus
+  checkEnvForPrerequisite({ name: 'ZEUS_API_TOKEN', legacyName: 'ZEUS_TOKEN' });
+  // We currently need ZEUS_TOKEN set for zeus-sdk to work properly
+  if (!process.env.ZEUS_TOKEN) {
+    process.env.ZEUS_TOKEN = process.env.ZEUS_API_TOKEN;
+  }
```
tonyo: We can probably move this block to the Zeus status/artifact provider constructors already.

src/utils/__tests__/env.test.ts
```diff
@@ -0,0 +1,136 @@
...
+        )
+      ).not.toThrowError(ConfigurationError);
+      expect(logger.warn).toHaveBeenCalledWith(
+        `Usage of OPAL is deprecated, and will be removed in later versions. ` +
+          `Please use CHARLIE instead.`
+      );
+      expect(process.env.CHARLIE).toEqual('GOOD PUPPY');
+    });
+  }); // end describe('multiple variables')
+}); // end describe('checkEnvForPrerequisites')
```
tonyo: That's... more than an average number of tests for this project, nice :)

src/commands/publish.ts
lobsterkatie: Agreed, but will do that in an artifact-provider-themed PR in the near future.

src/utils/__tests__/env.test.ts
lobsterkatie: Thanks! 🙂

I'll give it credit - once I realized I needed to refactor the function, the tests actually did help me catch a bug or two!

src/__tests__/config.test.ts
```diff
@@ -50,15 +35,16 @@ describe('compile-json-schema-to-typescript', () => {
 });
 
 describe('validateConfiguration', () => {
-  test('parses minimal configuration', async () => {
+  test('parses minimal configuration', () => {
```
lobsterkatie: Removed this `async` and others because the linter was complaining and nothing in the tests I removed it from was, in fact, async. LMK if it was there for some reason I haven't thought of, and if it needs to go back.

src/__tests__/config.test.ts
```diff
@@ -70,130 +56,10 @@ describe('validateConfiguration', () => {
 });
 
 describe('getProjectConfigSchema', () => {
-  test('returns non-empty object', async () => {
+  test('returns non-empty object', () => {
     const projectConfigSchema = getProjectConfigSchema();
 
-    expect(projectConfigSchema).toBeTruthy();
```
lobsterkatie: Took this out because even an empty object is truthy. (Why, JS, why??) Also, if the next to assertions pass, it's clearly not empty in any case.

src/__tests__/config.test.ts
```diff
@@ -50,15 +35,16 @@ describe('compile-json-schema-to-typescript', () => {
 });
 
 describe('validateConfiguration', () => {
-  test('parses minimal configuration', async () => {
```
lobsterkatie: Removed this `async` and others because the linter was complaining and nothing in the tests I removed it from was, in fact, async. LMK if it was there for some reason I haven't thought of, and if it needs to go back.

lobsterkatie:
Okay, I think we're good to go!

(FYI, one thing I had to do was re-solve the mocking-a-single-function riddle, and I managed to get it down to something pretty manageable. The minimal diff is [here](https://github.com/getsentry/craft/commit/81c1d1aa5478bcbfc3efa1eadf939354d1b8c363) in case you're curious.)

src/__tests__/config.test.ts
tonyo: Probably a remnant from the times when it was copied from some async test, and the linter didn't complain about it 🤷‍♂

lobsterkatie:
One last question occurred to me - is it weird that we have config stuff at the top level of `src` and env stuff in `utils`? Given that they're basically two halves of the same coin (I'm messing up that metaphor, but you get my point), it seems like they ought to be siblings.

My gut would be to move config stuff into `utils` (rather than move env stuff out), but happy to go with whichever of those options makes more sense to you.

tonyo:
> My gut would be to move config stuff into utils (rather than move env stuff out)
I'm fine with that, I'd just do it in a separate PR, because combining changes and file renames doesn't always look pretty diff-wise.

lobsterkatie:
> > My gut would be to move config stuff into utils (rather than move env stuff out)
> 
> I'm fine with that, I'd just do it in a separate PR, because combining changes and file renames doesn't always look pretty diff-wise.

Great. And ah, good point. Will do.

PR #63, opened 2020-03-13
ref(targets): Create GCS API module

lobsterkatie:
In preparation for the upcoming CGS artifact provider, this pulls all code from the GCS target module which might also useful to the artifact provider, collects it in a new GCS API module, and refactors the GCS target to use it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -597,12 +597,14 @@ The bucket paths (`paths`) can be interpolated using Mustache syntax (`{{ variab
 
 Google Cloud credentials can be provided using either of the following two environment variables.
 
-| Name                         | Description                                                              |
-| ---------------------------- | ------------------------------------------------------------------------ |
-| `CRAFT_GCS_CREDENTIALS_PATH` | Local filesystem path to Google Cloud credentials (service account file) |
-| `CRAFT_GCS_CREDENTIALS_JSON` | Full service account file contents, as a JSON string                     |
+| Name                          | Description                                                              |
+| ----------------------------- | ------------------------------------------------------------------------ |
+| `CRAFT_GCS_TARGET_CREDS_PATH` | Local filesystem path to Google Cloud credentials (service account file) |
+| `CRAFT_GCS_TARGET_CREDS_JSON` | Full service account file contents, as a JSON string                     |
 
-If defined, `CRAFT_GCS_CREDENTIALS_JSON` will be preferred over `CRAFT_GCS_CREDENTIALS_PATH`.
+If defined, `CRAFT_GCS_TARGET_CREDS_JSON` will be preferred over `CRAFT_GCS_TARGET_CREDS_PATH`.
+
+_Note:_ `CRAFT_GCS_TARGET_CREDS_JSON` and `CRAFT_GCS_TARGET_CREDS_PATH` were formerly called `CRAFT_GCS_CREDENTIALS_JSON` and `CRAFT_GCS_CREDENTIALS_PATH`, respectively. While those names will continue to work for the foreseeable future, you'll receive a warning encouraging you to switch to the new names.
 
 **Configuration**
 
--- a/src/utils/__tests__/gcsAPI.test.ts
+++ b/src/utils/__tests__/gcsAPI.test.ts
@@ -1,7 +1,13 @@
 import * as fs from 'fs';
 
-import { getGCSCredsFromEnv } from '../gcsApi';
+import { getGCSCredsFromEnv, CraftGCSClient } from '../gcsApi';
 import { withTempFile } from '../files';
+
+const mockGCSUpload = jest.fn();
+jest.mock('@google-cloud/storage', () => ({
+  Bucket: jest.fn(() => ({ upload: mockGCSUpload })),
+  Storage: jest.fn(() => ({})),
+}));
 
 describe('getGCSCredsFromEnv', () => {
   const cleanEnv = { ...process.env };
@@ -31,6 +37,10 @@ describe('getGCSCredsFromEnv', () => {
   });
 
   it('pulls filepath creds from env', async () => {
+    // ensure that the assertions below actually happen, since they in an async
+    // function
+    expect.assertions(3);
+
     await withTempFile(tempFilepath => {
       fs.writeFileSync(
         tempFilepath,
@@ -78,6 +88,17 @@ describe('getGCSCredsFromEnv', () => {
     }).toThrowError('Error parsing JSON credentials');
   });
 
+  it('errors if creds file missing from given path', () => {
+    process.env.DOG_CREDS_PATH = './iDontExist.json';
+
+    expect(() => {
+      getGCSCredsFromEnv(
+        { name: 'DOG_CREDS_JSON' },
+        { name: 'DOG_CREDS_PATH' }
+      );
+    }).toThrowError('File does not exist: `./iDontExist.json`!');
+  });
+
   it('errors if necessary field missing', () => {
     process.env.DOG_CREDS_JSON = `{
       "project_id": "squirrel-chasing",
@@ -92,3 +113,70 @@ describe('getGCSCredsFromEnv', () => {
     }).toThrowError('GCS credentials missing `client_email`!');
   });
 }); // end describe('getGCSCredsFromEnv')
+
+describe('CraftGCSClient class', () => {
+  const client = new CraftGCSClient({
+    bucketName: 'captured-squirrels',
+    credentials: {
+      client_email: 'might_huntress@dogs.com',
+      private_key: 'DoGsArEgReAtSoMeSeCrEtStUfFhErE',
+    },
+    projectId: 'squirrel-chasing',
+  });
+
+  beforeEach(() => {
+    jest.resetAllMocks();
+  });
+
+  it('calls the GCS library upload method with the right parameters', async () => {
+    await client.uploadArtifacts(['./dist/someFile'], {
+      path: '/some/destination/spot/',
+    });
+
+    expect(mockGCSUpload).toHaveBeenCalledWith('./dist/someFile', {
+      // contentType: 'application/javascript; charset=utf-8',
+      destination: '/some/destination/spot/someFile',
+      gzip: true,
+      metadata: { cacheControl: `public, max-age=300` },
+    });
+  });
+
+  it('detects content type correctly', async () => {
+    await client.uploadArtifacts(['./dist/bundle.js'], {
+      path: '/some/destination/spot/',
+    });
+
+    expect(mockGCSUpload).toHaveBeenCalledWith(
+      './dist/bundle.js',
+      expect.objectContaining({
+        contentType: 'application/javascript; charset=utf-8',
+      })
+    );
+  });
+
+  it('errors if destination path not specified', async () => {
+    await expect(
+      client.uploadArtifacts(['./dogs'], undefined as any)
+    ).rejects.toThrowError('no destination path specified!');
+
+    await expect(
+      client.uploadArtifacts(['./dogs'], {
+        path: undefined,
+      } as any)
+    ).rejects.toThrowError('no destination path specified!');
+  });
+
+  it('errors if GCS upload goes sideways', async () => {
+    mockGCSUpload.mockImplementation(() => {
+      throw new Error('whoops');
+    });
+
+    await expect(
+      client.uploadArtifacts(['./dist/someFile'], {
+        path: '/some/destination/spot/',
+      })
+    ).rejects.toThrowError(
+      'Error uploading `someFile` to `/some/destination/spot/someFile`'
+    );
+  });
+}); // end describe('CraftGCSClient class')
--- a/src/utils/__tests__/gcsAPI.test.ts
+++ b/src/utils/__tests__/gcsAPI.test.ts
@@ -1,0 +1,94 @@
+import * as fs from 'fs';
+
+import { getGCSCredsFromEnv } from '../gcsApi';
+import { withTempFile } from '../files';
+
+describe('getGCSCredsFromEnv', () => {
+  const cleanEnv = { ...process.env };
+
+  beforeEach(() => {
+    process.env = { ...cleanEnv };
+  });
+
+  it('pulls JSON creds from env', () => {
+    process.env.DOG_CREDS_JSON = `{
+      "project_id": "squirrel-chasing",
+      "private_key": "DoGsArEgReAtSoMeSeCrEtStUfFhErE",
+      "client_email": "might_huntress@dogs.com",
+      "other_stuff": "can be anything",
+      "tail_wagging": "true",
+      "barking": "also VERY true"
+    }`;
+
+    const { project_id, client_email, private_key } = getGCSCredsFromEnv(
+      { name: 'DOG_CREDS_JSON' },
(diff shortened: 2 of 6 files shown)


src/utils/gcsApi.ts
```diff
@@ -0,0 +1,321 @@
...
+   *
+   * @param localFilePath Location of the file to be uploaded
+   * @param uploadConfig Configuration for the upload including destination path
+   * and metadata
+   */
+  private async uploadArtifactFromPath(
+    localFilePath: string,
+    uploadConfig: Pick<
+      Required<GCSUploadOptions>,
+      'destination' | 'metadata' | 'gzip'
```
tonyo: To clarify, you want to require those three attributes, correct?

src/utils/gcsApi.ts
```diff
@@ -0,0 +1,321 @@
...
+ */
+const CONTENT_TYPES_EXT: Array<[RegExp, string]> = [
+  [/\.js$/, 'application/javascript; charset=utf-8'],
+  [/\.js\.map$/, 'application/json; charset=utf-8'],
+];
+
+/**
+ * Is this bucket being used as an artifact store or a target?
+ */
+export const enum BucketRole {
```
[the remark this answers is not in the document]
lobsterkatie: Yeah, it's a valid point - should the utility class be aware of how it's being used or not? I did that in order to enable it to behave differently depending on what it was being used for.

In the end, the only place where that came into play was tagging the logger appropriately (in the constructor), which is why I don't actually store the value past that point. (Which answers your second question as well - it's in the config object because that's what gets passed to the constructor.)

I guess I'm not entirely sure how this is different from something like the credentials, which are also in the config and also discarded as soon as we've set up the bucket.

src/utils/gcsApi.ts
```diff
@@ -0,0 +1,321 @@
...
+
+  /**
+   * Pulls GCS redentials out of the environment, where they can be stored either
+   * as a path to a JSON file or as a JSON string.
+   *
+   * @returns An object containing the credentials
+   */
+  public static getGCSCredsFromEnv = (
+    gcsRole: BucketRole
+  ): { [key: string]: string } => {
```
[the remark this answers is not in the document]
lobsterkatie: Sure, I can do that.

Being new to typescript, I'm not sure where to strike the balance between the two extremes of no interfaces (easy! fewer lines of code the write or read! but also: why even use TS if that's what you're doing to do?) and making every one-off collection of properties in an object into an interface (ultimate type safety! every single object is documented! but also: a lot of work, and the strong likelihood of making your code so busy the reader can't tell what's important.).

Intuitively, I think I've been going with a heuristic something along the lines of:

- if it's used multiple times, maybe it should be an interface, so that you know all of those instances have the same structure
- if it's used as a method of passing data between modules - literally an _inter_face - (as is the case with `GCSBucketConfig`, for instance), almost certainly make it an interface, so you know the two modules are speaking the same language
- if it's only used once, internally, probably don't make it an interface, because you risk drowning out the interfaces from the first two cases (or at least making it difficult for the reader to tell whether it's _one of those_ or _one of these_)

but I'm sure there's more nuance to it. Any tips?

src/utils/gcsApi.ts
lobsterkatie: Yeah - I needed a way to translate the sentence

_You have to follow the `GCSUploadOptions` interface but also, even though I know these things are optional there, I need you to include `destination`, `metadata`, and `gip` in your object_.

into Typescript. If there's a more elegant way to do that, I'd definitely be interested to know.

src/utils/gcsApi.ts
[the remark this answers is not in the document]
lobsterkatie: Oh, wait - scratch that. Setting the logger tag is not the only place `bucketRole` is used (I gave it a different name for some reason, which is why my search didn't find it; fixed now). It's also used to know which env variables to look for (in `getGCSCredsFromEnv`), and I'm not immediately seeing a way around that.

src/utils/gcsApi.ts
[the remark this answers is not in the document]
tonyo: What you wrote makes sense, but since you return an object and not a primitive type, I think it already makes sense to add a type for it: e.g., it helps you to catch things like misnamed or missing fields.
So I'd personally find the code easier to read if we had that interface.
You don't have to export it, of course.

src/utils/gcsApi.ts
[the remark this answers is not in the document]
tonyo: > I guess I'm not entirely sure how this is different from something like the credentials, which are also in the config and also discarded as soon as we've set up the bucket.

I think what bothers me here is that the user of the function doesn't really care about that logger prefix, but we require it as part of `GCSBucketConfig`.
I don't mind keeping the enum for picking up the right env variables, but not sure that the logger prefix is worth it.

README.md
```diff
@@ -597,12 +597,14 @@ The bucket paths (`paths`) can be interpolated using Mustache syntax (`{{ variab
...
-| `CRAFT_GCS_CREDENTIALS_JSON` | Full service account file contents, as a JSON string                     |
+| Name                          | Description                                                              |
+| ----------------------------- | ------------------------------------------------------------------------ |
+| `CRAFT_GCS_TARGET_CREDS_PATH` | Local filesystem path to Google Cloud credentials (service account file) |
+| `CRAFT_GCS_TARGET_CREDS_JSON` | Full service account file contents, as a JSON string                     |
 
-If defined, `CRAFT_GCS_CREDENTIALS_JSON` will be preferred over `CRAFT_GCS_CREDENTIALS_PATH`.
+If defined, `CRAFT_GCS_TARGET_CREDS_JSON` will be preferred over `CRAFT_GCS_TARGET_CREDS_PATH`.
+
+_Note:_ `CRAFT_GCS_TARGET_CREDS_JSON` and `CRAFT_GCS_TARGET_CREDS_PATH` were formally called `CRAFT_GCS_CREDENTIALS_JSON` and `CRAFT_GCS_CREDENTIALS_PATH`, respectively. While those names will continue to work for the foreseeable future, you'll receive a warning encouraging you to switch to the new names.
```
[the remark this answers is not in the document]
lobsterkatie: Drat. Yes. Thanks. Will fix.

src/utils/gcsApi.ts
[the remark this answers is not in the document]
lobsterkatie: > but since you return an object and not a primitive type, I think it already makes sense to add a type for it: e.g., it helps you to catch things like misnamed or missing fields.

This seems to imply that any function/method which returns an object should have an associated interface. Is that generally the rule you go by?

In any case, added one here. Had to tweak the code a bit to make the compiler happy.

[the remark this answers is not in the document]
tonyo: I haven't written a lot of Typescript myself, but I feel like when we know the shape of the data we return (i.e. all the object attributes), why not having a type for it?

> because you risk drowning out the interfaces from the first two cases

I honestly have never experienced that. Do you have examples of this in the codebase?

src/utils/gcsApi.ts
```diff
@@ -0,0 +1,301 @@
...
+    const destinationPath = path.dirname(destinationFilePath);
+    const filename = path.basename(localFilePath);
+    // const fileUploadConfig: GCSUploadOptions = { ...uploadConfig };
+    // const contentType = this.detectContentType(filename);
+    // if (contentType) {
+    //   fileUploadConfig.contentType = contentType;
+    // }
+    const fileUploadConfig: GCSUploadOptions = {
+      ...uploadConfig,
+      contentType: this.detectContentType(filename),
```
tonyo: Don't remember the exact behavior, but there's a chance that `contentType: undefined` might cause problems (hence the previous, commented-out implementation).

src/utils/gcsApi.ts
```diff
@@ -0,0 +1,301 @@
...
+    );
+  }
+
+  /**
+   * Pulls GCS redentials out of the environment, where they can be stored either
+   * as a path to a JSON file or as a JSON string.
+   *
+   * @returns An object containing the credentials
+   */
+  public static getGCSCredsFromEnv = (bucketRole: BucketRole): GCSCreds => {
```
tonyo: Let's also make this a "regular" function :)

src/utils/gcsApi.ts
tonyo: 🤔 
Can't think of anything better.

src/utils/gcsApi.ts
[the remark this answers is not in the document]
lobsterkatie: > > because you risk drowning out the interfaces from the first two cases
> 
> I honestly have never experienced that. Do you have examples of this in the codebase?

Not one in particular, no. More just an instinct to keep things tidy and following the idea of creating an interface for every one-off object that gets returned to its logical conclusion. Who knows - perhaps I'm worrying when I don't need to be, and in practice it's fine. ¯\_(ツ)_/¯

src/utils/gcsApi.ts
lobsterkatie: 👍🏻

src/utils/gcsApi.ts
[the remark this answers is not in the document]
lobsterkatie: > Let's also make this a "regular" function :)

Okay, so to confirm:

- Move `getGCSCredsFromEnv` to be a function in the module (rather than a static method on the class)
- Pass `bucketRole` to that function, in order to get the right creds from the env
- Don't pass `bucketRole` to the class constructor, because we don't care so much about labeling the logger

I can do that - I agree that the logger tag is a nice-to-have rather than a necessity. I guess I'm still confused why passing it to the function is fine, but passing it to the constructor isn't. The "user" you're talking about is the same in both cases (namely, either the GCS target module or the soon-to-exist GCS artifact store module) and in fact the function will be called only about six lines above the constructor, so... 🤔

And I apologize, I'm not meaning to seem like I'm arguing, nor do I mean to belabor the point, but I'd really like to understand the underlying principle here, so I can make good choices going forward.

src/utils/gcsApi.ts
lobsterkatie: Yes, thanks for the reminder. The commented out part in fact is still there specifically so I remember to test out whether or not that's the case (and so I can easily switch it back if the current implementation does cause problems), but I should have put a `TODO` also.

Will resolve this before finalizing the PR.

src/utils/gcsApi.ts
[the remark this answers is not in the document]
tonyo: I guess my point is, passing the role (or, basically, an indicator what env vars we want to read) to `getGCSCredsFromEnv` mostly makes sense to me. It's not how I would do this (I'd prefer to have two separate functions that maybe reuse some utility function?), and probably that's why I'm not fully happy with the current implementation, but it doesn't seem wrong.

But passing the role to the constructor just seems... unnecessary? We could theoretically pass lots of different parameters to utility functions to give them more "context" (for loggers and whatnot), but it seems to me like a line we don't want to cross, because it pollutes the API.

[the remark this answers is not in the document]
tonyo: > Let's also make this a "regular" function :)

I actually meant rewriting arrow function to `function()` function, but moving it outside the class also makes sense to me 👍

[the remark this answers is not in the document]
lobsterkatie: Ah, okay. Well, in the process of doing the latter I ended up doing the former anyway, so, great! 🙂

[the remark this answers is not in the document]
lobsterkatie: > I'd prefer to have two separate functions that maybe reuse some utility function

> but it doesn't seem wrong

Okay, so whether or not I change that part, just so I see all the options, you'd do something like 

`getCredsFromEnv = (jsonVarName, pathVarName) => credsObject` 

and then, in the `gcsTarget` code, 

`creds = getCredsFromEnv('CRAFT_GCS_TARGET_CREDENTIALS_JSON', 'CRAFT_GCS_TARGET_CREDENTIALS_PATH')`

?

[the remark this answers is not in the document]
tonyo: Yes, something like that!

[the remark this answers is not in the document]
lobsterkatie: Okay, done. See what you think.

src/utils/gcsApi.ts
lobsterkatie: You were right - it wasn't what I wanted, but I finally remembered the syntax I was looking for:

```
    const contentType = this.detectContentType(filename);
    const fileUploadConfig: GCSUploadOptions = {
      ...uploadConfig,
      contentType: this.detectContentType(filename),
      ...(contentType && { contentType }),
    };
```

Tested and working both when `contentType` is defined and when it's not.

src/utils/__tests__/gcsAPI.test.ts
```diff
@@ -0,0 +1,182 @@
...
+
+    await expect(
+      client.uploadArtifacts(['./dist/someFile'], {
+        path: '/some/destination/spot/',
+      })
+    ).rejects.toThrowError(
+      'Error uploading `someFile` to `/some/destination/spot/someFile`'
+    );
+  });
+}); // end describe('CraftGCSClient class')
```
tonyo: 👍 👍

src/utils/gcsApi.ts
```diff
@@ -0,0 +1,273 @@
...
+ * @returns An object containing the credentials
+ */
+export function getGCSCredsFromEnv(
+  jsonVar: RequiredConfigVar,
+  filepathVar: RequiredConfigVar
+): GCSCreds {
+  // make sure we have at least one of the necessary variables
+  try {
+    checkEnvForPrerequisite(jsonVar, filepathVar);
+  } catch (oO) {
```
tonyo: Tiny thing: everywhere else we use `e` as a var name in error handlers (i.e. `catch (e)...`), let's keep using that for consistency.

lobsterkatie: Okay, though in this case I was specifically trying to indicate, "I'm ignoring this thing." When it's something I'm actually using, I've been calling it `err`, though I can switch to `e` if you'd rather.

(The `oO` is something I picked up from @kamilogorek. 🙂Could switch it to `_` if you think that'd make the "this is getting thrown away" hint clearer.)

kamilogorek: I love my `o_O` :D

lobsterkatie: @tonyo - Should I make the change to `catch(_)`, or merge as is? Happy to do either.

tonyo: I guess consistency wouldn't hurt here (hence my suggestion about `e` that's used everywhere else), but if you want to explicitly mark it as unused, we can use `_` or `_e`.
`_` may potentially conflict with `lodash` (which is kinda stupid, I think we should replace all of those usages), but I don't have a strong opinion, your call.

kamilogorek: You should be using `lodash-es` anyway,  as it doesnt have to be treeshaken.

lobsterkatie: > I think we should replace all of those usages

I'll put it on my todo list 😊

lobsterkatie: As for the `catch` variable, I settled for calling it `e` and just leaving a comment about how we're ditching it (and why).

PR #66, opened 2020-03-20
ref(GCS client): Always pass full `CraftArtifact` objects

lobsterkatie:
For consistency, when communicating between modules, always pass full `CraftArtifact` objects.

(Previously, there was one place where we were passing just local filepaths, while in all others we passed around `CraftArtifact` objects.)

### DIFF
--- a/src/utils/gcsApi.ts
+++ b/src/utils/gcsApi.ts
@@ -11,6 +11,7 @@ import { isDryRun } from './helpers';
 import { logger as loggerRaw } from '../logger';
 import { reportError, ConfigurationError } from './errors';
 import { checkEnvForPrerequisite, RequiredConfigVar } from './env';
+import { CraftArtifact } from '../artifact_providers/base';
 
 const DEFAULT_MAX_RETRIES = 5;
 const DEFAULT_UPLOAD_METADATA = { cacheControl: `public, max-age=300` };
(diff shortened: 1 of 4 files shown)


src/utils/gcsApi.ts
```diff
@@ -172,24 +173,30 @@ export class CraftGCSClient {
...
    *
-   * @param localFilePath Location of the file to be uploaded
+   * @param artifact The artifact to be uploaded
    * @param uploadConfig Configuration for the upload including destination path
    * and metadata
    */
-  private async uploadArtifactFromPath(
-    localFilePath: string,
+  private async uploadArtifact(
+    artifact: CraftArtifact,
```
tonyo: Does this mean we have to provided the required "storedFile" attribute in `artifact` now?

tonyo:
Could you clarify why we need this "consistency"? `CraftArtifact` initially meant to represent a "remote" artifact (hence all the attributes), and looks like now we're forcing users to create a valid `CraftArtifact` object before uploading, is that really necessary?

src/utils/gcsApi.ts
lobsterkatie: Shoot. Good point. Not a problem for the target, but would be a problem for the artifact provider, before the files have been uploaded. Thanks for catching that. Will fix.

lobsterkatie:
> `CraftArtifact` initially meant to represent a "remote" artifact

> Could you clarify why we need this "consistency"?

> now we're forcing users to create a valid `CraftArtifact` object before uploading, is that really necessary?

So here was my thinking:

1) This client is meant to be like (and eventually take the place of) the zeus client, and it deals in `Artifact` objects, which I was seeing as the analog to `CraftArtifact` objects.

lobsterkatie:
Okay, have no idea how I typo'd my way into closing this... 🤔

tonyo:
It generally makes sense, my point is that it theoretically can bring in some confusion regarding what are local attributes, what are remote, which of them are mutable, which are required, etc.
I don't have any good examples to back up this claim, so we can try and see what happens.

src/utils/gcsApi.ts
tonyo: Haven't checked in details, but probably we can just make `storedFile` optional.

lobsterkatie:
No - upon further reflection and futzing and rubberducking w some helpful folks in Slack and futzing some more... I will say (not for the first time, and, I'm guessing, not the last): you're right, and I finally get what you were saying. Futzed until I came to a solution (a good one, too, one that was demonstrably better than my original), all on my own... and it was basically exactly what you had done originally. 🤦🏻‍♀️

Never mind. Sorry for leading you in circles.

lobsterkatie:
BTW, here 

https://github.com/getsentry/craft/commit/1d0132a5ca06f03b0f73614524b8a2fcb175074f

is the core of that solution (which just proves your point that `CraftArtifact` really means "artifact on the artifact provider," as every use of `CraftArtifact` became `CloudArtifact` - which is the point at which the lightbulb finally went off for me). Though functionally it's no different (and if I don't do ^^^^^, the `LocalArtifact` interface is currently unused), 

a) it makes your point above glaringly obvious, and 
b) it gives us more flexibility in case we ever do revisit either the question of passing whole artifact objects or the question of where in the callstack we want to switch from `upload/downloadAll` to `upload/downloadOne` (as we've talked about elsewhere),

so I'd kinda like to PR it. Before I litter the repo with more aborted PRs, though, I'd love to know what you think of that plan.

tonyo:
@lobsterkatie Sorry, just noticed the request for comments.
I think it's a good approach 👍 
I'd just rename CloudArtifact to RemoteArtifact, probably.

lobsterkatie:
Can do - thanks.

lobsterkatie:
Okay, done (along with some other cleanup) in https://github.com/getsentry/craft/pull/68.

PR #71, opened 2020-04-02
fix(dry-run): Fix reporting of required artifact check in dry-run mode

lobsterkatie:
In dry-run mode, errors don't actually stop iteration of a for loop (since they're just logged, not thrown), so it's possible to make it all the way through a fail-fast loop, even having failed on at least one iteration. Therefore, we can't assume that just because we made it past the for loop, the test succeeded.

This adds a flag and a check of that flag, so that we report the correct result in dry-run mode.

### DIFF
--- a/src/commands/publish.ts
+++ b/src/commands/publish.ts
@@ -257,12 +257,17 @@ async function checkRequiredArtifacts(
     throw new Error(`Revision ${revision} not found!`);
   }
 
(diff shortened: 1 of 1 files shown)


lobsterkatie:
Sure. Definitely not going out and looking for them - this one just jumped out at me and was an easy fix.

PR #72, opened 2020-04-02
ref(zeus): Make zeus handle its own special cases

lobsterkatie:
Currently, all of the `listArtifactsForRevision` methods on artifact providers (the base and its descendants) have a signature indicating that they return one of three things: an array full of artifacts, an empty array, or `undefined`. The reason for this is historical, as these three possible return values correspond to the three possible responses when asking zeus for a commit's artifacts, respectively

1. "Yup, got that commit, and here are the associated artifacts."
2. "Yup, got that commit, but there are no artifacts associated with it."
3. "Wait, what commit do you want stuff for? Never heard of it."

The only reason response number two is a possibility, though, is because zeus is both a status provider _and_ an artifact provider, the former role being the one which would let it learn of a commit independent of any associated artifacts; any service serving solely as an artifact provider would have no way of landing in situation two.

Further, zeus - unlike other artifact providers - will store multiple copies of the same file attached to the same commit, rather than overwriting the file when a newer version is uploaded to a commit which already has an older version. Therefore, when retrieving the file list, we need to group them by filename and sort them by update date, to mimic the overwriting behavior.

This PR makes it so that zeus, rather than the base class, has to deal with its own differing behavior, thereby simplifying the base class, its other descendants, and any files which use methods from same. Specifically, it lets us cut `undefined` out of the signature for all of these methods, since we only need to represent situations one and three above, which we can do by returning either a full array or an empty array. (Zeus still handles all three cases, but now does so with log statements rather than return values, since functionally cases two and three are equivalent.)

### DIFF
--- a/src/artifact_providers/base.ts
+++ b/src/artifact_providers/base.ts
@@ -225,19 +225,15 @@ export abstract class BaseArtifactProvider {
   }
 
   /**
-   * Gets a list of all recent artifacts for the given revision, either from the
-   * cache or from the provider's API.
-   *
-   * If there are several artifacts with the same name, returns the most recent
-   * of them.
+   * Gets a list artifacts for the given revision, either from the cache or from
+   * the provider's API.
    *
    * @param revision Git commit id
-   * @returns Filtered list of artifacts, or "undefined" if the revision cannot
-   * be found
+   * @returns List of artifacts associated with that commit
    */
   public async listArtifactsForRevision(
     revision: string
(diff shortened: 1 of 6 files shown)


lobsterkatie:
> > "zeus is the only artifact provider which..."
>
> I just hope it will stay the only exception 😅

LOL - fair. I suppose I'm hoping that for a decent while at least, we're really only going to be talking about the two options.

PR #73, opened 2020-04-03
feat(artifacts): Implement GCS artifact provider

lobsterkatie:
Adds another option besides zeus for artifact storage.

Things happening in this PR (commits should be relatively self-contained) :

- All config provided to artifact constructors is now passed in a `config` object, which is then stored on the provider instance. This config now also includes any options specified for the artifact provider in the project config file.
- New `GCSArtifactProvider` class, implementing artifact listing and downloading (and underlying methods implemented in GCS client).
- Fix for GCS creds env variable name, which didn't match README.
- Tests for new GCS client methods. Includes pulling fixtures and the logger mock each into their own file.

Things to happen in further PRs:

- General clean up of code and logging associated with the above changes.
- Creation of an `upload` CLI command.
- Hopefully some testing of the artifact providers themselves.

### DIFF
--- a/src/artifact_providers/zeus.ts
+++ b/src/artifact_providers/zeus.ts
@@ -114,14 +114,15 @@ export class ZeusArtifactProvider extends BaseArtifactProvider {
   protected async doListArtifactsForRevision(
     revision: string
   ): Promise<RemoteArtifact[]> {
+    const { repoName, repoOwner } = this.config;
     logger.debug(
-      `Fetching Zeus artifacts for ${this.repoOwner}/${this.repoName}, revision ${revision}`
+      `Fetching Zeus artifacts for ${repoOwner}/${repoName}, revision ${revision}`
     );
     let artifacts;
     try {
       artifacts = await this.client.listArtifactsForRevision(
-        this.repoOwner,
-        this.repoName,
+        repoOwner,
+        repoName,
         revision
       );
     } catch (e) {
--- a/src/commands/publish.ts
+++ b/src/commands/publish.ts
@@ -426,9 +426,11 @@ export async function publishMain(argv: PublishOptions): Promise<any> {
   logger.debug('Revision to publish: ', revision);
 
   const statusProvider = getStatusProviderFromConfig();
-  logger.info(`Using "${statusProvider.constructor.name}" for status checks`);
   const artifactProvider = getArtifactProviderFromConfig();
+  logger.info(' ');
+  logger.info(`Using "${statusProvider.constructor.name}" for status checks`);
   logger.info(`Using "${artifactProvider.constructor.name}" for artifacts`);
+  logger.info(' ');
 
   // Check status of all CI builds linked to the revision
   await checkRevisionStatus(statusProvider, revision, argv.noStatusCheck);
--- a/src/utils/__fixtures__/gcsFileObj.ts
+++ b/src/utils/__fixtures__/gcsFileObj.ts
@@ -1,0 +1,179 @@
+// tslint:disable: object-literal-sort-keys
+
+// Note: All but the last two objects in this file (the ones which are exported)
+// exist purely because real-life GCS File objects are enormous, deeply nested,
+// and full of repetition, so breaking them up is the only way to have any idea
+// of what's in them.
+
+const acl = {
+  owners: {},
+  readers: {},
+  writers: {},
+  pathPrefix: '/acl',
+};
+
+const defaultAcl = {
+  ...acl,
+  pathPrefix: '/defaultObjectAcl',
+};
+
+const aclWithDefault = {
+  ...acl,
+  default: defaultAcl,
+};
+
+const aclRoles = {
+  OWNER_ROLE: 'OWNER',
+  READER_ROLE: 'READER',
+  WRITER_ROLE: 'WRITER',
+};
+
+const scopes = [
+  'https://www.googleapis.com/auth/iam',
+  'https://www.googleapis.com/auth/cloud-platform',
+  'https://www.googleapis.com/auth/devstorage.full_control',
+];
+
+const authClient = {
+  jsonContent: {
+    client_email: 'mighty_huntress@dogs.com',
+    private_key: 'DoGsArEgReAtSoMeSeCrEtStUfFhErE',
+  },
+  cachedCredential: {
+    domain: null,
+    _events: {},
+    _eventsCount: 0,
+    transporter: {},
+    credentials: {
+      access_token: 'IaMaGoOdDoGpLeAsElEtMeIn',
+      token_type: 'Bearer',
+      expiry_date: 1585600265000,
+      refresh_token: 'jwt-placeholder',
+    },
+    certificateExpiry: null,
+    refreshTokenPromises: {},
+    eagerRefreshThresholdMillis: 300000,
+    email: 'mighty_huntress@dogs.com',
+    key: 'DoGsArEgReAtSoMeSeCrEtStUfFhErE',
+    scopes,
+    gtoken: {
+      token: 'IaMaGoOdDoGpLeAsElEtMeIn',
+      expiresAt: 1585600265000,
+      rawToken: {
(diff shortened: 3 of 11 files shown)


src/commands/publish.ts
```diff
@@ -426,9 +426,17 @@ export async function publishMain(argv: PublishOptions): Promise<any> {
...
+  const statusProviderName = statusProvider.constructor.name;
   const artifactProvider = getArtifactProviderFromConfig();
-  logger.info(`Using "${artifactProvider.constructor.name}" for artifacts`);
+  const artifactProviderName = artifactProvider.constructor.name;
+  logger.info(' ');
+  logger.info(`Using "${statusProviderName}" for status checks`);
+  logger.info(`Using "${artifactProviderName}" for artifacts`);
+  if (statusProviderName !== 'zeus' && artifactProviderName !== 'zeus') {
+    logger.info(' ');
+    logger.info("    Zeus? Who's Zeus?");
```
tonyo, line 437: Haha
You sure you want to display that when providers are _not_ zeus, though?

src/artifact_providers/zeus.ts
```diff
@@ -114,14 +114,15 @@ export class ZeusArtifactProvider extends BaseArtifactProvider {
...
     logger.debug(
-      `Fetching Zeus artifacts for ${this.repoOwner}/${this.repoName}, revision ${revision}`
+      `Fetching Zeus artifacts for ${repoOwner}/${repoName}, revision ${revision}`
     );
     let artifacts;
     try {
       artifacts = await this.client.listArtifactsForRevision(
-        this.repoOwner,
-        this.repoName,
+        this.config.repoOwner,
```
tonyo, line 124: I guess we can remove `this.config` from these lines, then.

src/utils/env.ts
```diff
@@ -109,6 +109,7 @@ export function readEnvironmentConfig(
   // Read from home dir
   const homedirEnvFile = join(homedir(), ENV_FILE_NAME);
   if (existsSync(homedirEnvFile)) {
+    console.log('here!!!!');
```
tonyo, line 112: 🔍

src/utils/gcsApi.ts
```diff
@@ -234,4 +236,104 @@ export class CraftGCSClient {
...
+
+  /**
+   * Downloads a file stored on the artifact provider
+   *
+   * @param artifact The artifact to download
+   * @param destinationDirectory Path to directory into which to download the
+   * file
+   * @returns Path to the downloaded file
+   */
+  public async downloadArtifact(
```
tonyo, line 248: Since this is a lower-level util function, it doesn't really have to know about artifacts.
Looks like we only need the remote path and the resulting filename from the artifact, perhaps we can just pass those values directly, what do you think?

src/utils/__tests__/gcsAPI.test.ts
```diff
@@ -7,235 +8,324 @@ import {
...
+
+    describe('listArtifactsForRevision', () => {
+      it('calls the GCS library getFiles method with the right parameters', async () => {
+        mockGCSGetFiles.mockReturnValue([[]]);
+
+        await client.listArtifactsForRevision(
+          dogsGHOrg,
+          squirrelRepo,
+          squirrelSimulatorCommit
+        );
```
tonyo, line 297: Perhaps you want an assert here that it actually returns an empty list?

src/utils/__tests__/env.test.ts
```diff
@@ -157,7 +157,7 @@ describe('env utils functions', () => {
       writeFileSync(outConfigFile, '');
     }
 
-    test('calls homedir/findConfigFile', async () => {
+    test.only('calls homedir/findConfigFile', async () => {
```
tonyo, line 160: Let's get rid of this.

src/utils/__fixtures__/gcsFileObj.ts
```diff
@@ -0,0 +1,1950 @@
+// tslint:disable: object-literal-sort-keys
+// tslint:disable: max-file-line-count
+
+export const squirrelStatsGCSFileObj = {
```
tonyo, line 4: Good job with setting up all these fixtures 👍 
Do you think we need this object as a whole, though?
Can we perhaps kill (or add a stub) packageJson attributes at least? They are useless for our tests and make the fixture less readable.

src/commands/publish.ts
lobsterkatie: That was kind of the little joke. We're not reliant on Zeus anymore, so we can pretend we’ve never heard of him. But if it takes explaining, it's not funny. (This is what happens when nerds try to make jokes.) Can take it out.

src/utils/env.ts
lobsterkatie: This is an artifact of me trying to debug why travis was failing (see my thread w Billy in #discuss-javascript - seems like there might be an actual bug in either travis or jest). That's why I hadn't actually requested a review yet - plan was to debug that then get rid of that last commit `testing the test`, which introduced this. I should have been more communicative on that point, though, sorry.

src/utils/__tests__/env.test.ts
lobsterkatie: See above re: debugging the tests.

src/utils/env.ts
tonyo: Ok sure, feel free to request another review when everything is ready.

src/utils/__fixtures__/gcsFileObj.ts
lobsterkatie: Sure. Can do. I originally had pulled out everything we didn't use, but then had second thoughts - maybe making it too clean would put us in danger of things passing which wouldn't otherwise, with a more realistic fixture? So then I just left everything in. (You're right, though, that the `packageJson` entries shouldn't be what makes the difference.)

Any tips on where the draw the line on questions like ^^^^ ?

tonyo: 🤷‍♂ 
No fixture is ideal, but stripping down highly unrelated attributes while preserving the overall structure seems ok to me.

src/artifact_providers/zeus.ts
lobsterkatie: Oh, good catch, thanks.

src/utils/gcsApi.ts
lobsterkatie: 👍🏻

src/utils/__tests__/gcsAPI.test.ts
lobsterkatie: Isn't that just testing jest at that point, though? I've told jest, "Please return an empty list when I call this method," so asserting that it does so only tests whether jest is going to do what I asked, not anything to do with my code (since mocked return values ignore what happens inside the function entirely).

src/utils/__fixtures__/gcsFileObj.ts
lobsterkatie: Okay, got rid of duplicated stuff (of which there was a LOT) and `packageJson`. Cut the filesize by a factor of more than 10! 😮

src/utils/__tests__/gcsAPI.test.ts
tonyo: Yeah, but you're mocking `getFiles` and then calling `listArtifactsForRevision`, that may ignore the result of `getFiles` in case of a bug or something.
So I guess what we also might want to test is that `listArtifactsForRevision` in this case returns the (processed) output of `getFiles`.

lobsterkatie: Ohhhh - I had misunderstood who "it" was. Since my existing assertion is on the GCS method, I thought you were saying I should do

```javascript
mockGCSGetFiles.mockReturnValue([[]]);
...
expect(mockGCSGetFiles).toReturn([[]]);
```

Okay, sure, now I get what you're saying. Will add it.

PR #78, opened 2020-04-22
fix(GCS): Fix location of `contentType` upload option

lobsterkatie:
According to the GCS docs ([1](https://googleapis.dev/nodejs/storage/latest/global.html#UploadOptions) and [2](https://cloud.google.com/storage/docs/json_api/v1/objects/insert#request_properties_JSON)), `contentType` is an attribute _inside of_ (rather than as a sibling to) `metadata` in the `UploadOptions` object. However, if you do put it in the wrong place, it appears that the underlying GCS library tries to be clever, and moves it for you... thereby mutating an object (in this case, `DEFAULT_UPLOAD_METADATA`) whose value would otherwise never change. As a result of that mutation, the `contentType` value from one file can leak into the upload options for other files, causing incorrect content types.

This PR fixes that by 1) including `contentType` in the right spot, an 2) cloning the `DEFAULT_UPLOAD_METADATA` object rather than just pointing to it.

Before:

image

After:

image

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1845,11 +1845,6 @@ dot-prop@^4.1.0:
   dependencies:
     is-obj "^1.0.0"
 
-dryrun@1.0.2:
-  version "1.0.2"
-  resolved "https://registry.yarnpkg.com/dryrun/-/dryrun-1.0.2.tgz#bf59e1193efea0de5a140ee89f6a89af81cee305"
-  integrity sha1-v1nhGT7+oN5aFA7on2qJr4HO4wU=
-
 duplexer2@~0.1.4:
   version "0.1.4"
   resolved "https://registry.yarnpkg.com/duplexer2/-/duplexer2-0.1.4.tgz#8b12dab878c0d69e3e7891051662a32fc6bddcc1"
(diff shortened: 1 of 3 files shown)


yarn.lock
```diff
@@ -1845,11 +1845,6 @@ dot-prop@^4.1.0:
   dependencies:
     is-obj "^1.0.0"
 
-dryrun@1.0.2:
-  version "1.0.2"
-  resolved "https://registry.yarnpkg.com/dryrun/-/dryrun-1.0.2.tgz#bf59e1193efea0de5a140ee89f6a89af81cee305"
-  integrity sha1-v1nhGT7+oN5aFA7on2qJr4HO4wU=
```
rhcarvalho, line 1851: How is this related to the `contentType` changes?

tonyo: It's not.
Artifact of not running `yarn` after getting rid of that package a few PRs back.

lobsterkatie: Oh, yeah, didn't see that you'd added that in there. Come to think about it, it did disappear from my general cleanup branch when I rebased yesterday...

PR #79, opened 2020-04-24
chore(general): Post-GCS artifact provider cleanup

lobsterkatie:
As I was working on the GCS artifact provider, there were all sorts of tiny things which I fixed or tweaked or otherwise mucked with, which had nothing to do with the actual feature - just stuff I happened upon while wandering around the codebase as I built the provider. I kept them out of that PR (and ones leading up to it) so those PRs would be easier to review, but now I'd like to clear the decks before starting in on the `upload` (and eventual `download` and `list`) command(s).

As always, I tried to keep each commit scoped to one idea, so they each ought to be independently reviewable, and hopefully each easier to digest than the whole. The big themes addressed (in roughly the same order as the commits [_EDIT: for no reason that I can explain, the commits below are showing up in a different order here than in my local `git log`, but pretend that the last two listed are actually first and the order should be right_]):

- typos, formatting, etc
- the use of `async` in tests of non-async code, which makes the linter mad, and more consistent use of `expect.assertions(n)` in tests that actually are async
- a small handful of renamed variables for clarity
- better logging (added messages, more consistent scope markers, addition of `contentType` to artifact table, other small fixes)
- a bunch of todos
- a few added comments


src/utils/__tests__/githubApi.test.ts
```diff
@@ -161,6 +162,7 @@ describe('retryHttp', () => {
         ...params,
         cleanupFn: async () => {
           cleanupCalled += 1;
+          return Promise.resolve();
```
tonyo, line 165: Isn't this done implicitly?

src/commands/publish.ts
```diff
@@ -185,19 +186,25 @@ async function printRevisionSummary(
 ): Promise<void> {
   const artifacts = await artifactProvider.listArtifactsForRevision(revision);
   if (artifacts.length > 0) {
-    const artifactData = artifacts.map(ar => [
-      ar.filename,
-      formatSize(ar.storedFile.size),
-      ar.storedFile.lastUpdated || '',
-    ]);
+    const artifactData = artifacts.map(ar => {
+      const contentType = detectContentType(ar.filename);
```
tonyo, line 190: To clarify, this content type guessing was initially meant to be done for GCS target only.

Artifact providers technically can store content types for files as part of metadata, but honestly, I don't really see a consistent way of showing this information: that content type we're showing, was it guessed? Is it part of artifact metadata? Also, we don't really attach the contentType we detected to the artifact itself, so later in the pipeline (e.g. inside GCS target) we have to detect it again.

src/utils/__tests__/githubApi.test.ts
lobsterkatie: Yes - but the linter was mad, so instead of ignoring it, I f figured, why not be explicit? ¯\_(ツ)_/¯ Happy to switch to a linter-calming comment if you'd like.

src/commands/publish.ts
lobsterkatie: > To clarify, this content type guessing was initially meant to be done for GCS target only.

I see. Okay, so currently it's part of the underlying `gcsApi.uploadArtifact()` method, meaning it applies to the GCS artifact provider as well. Do you anticipate that causing problems, such that I should extract it and stick it back into the target only?

In any case, I take your point that at this point in the process (when we're just logging the list of files) we don't actually record content type on the artifact , so if the detected type differs from the stored type, and we're targeting something other than GCS, this could give a false impression of what we're uploading elsewhere. Two options I see here:

1) Take out the detection at this stage and just report the real data we have (attached to the artifact object) on content type. In conjunction, make it more obvious when, as part of GCS target upload, we change the type.

2) Take out this reporting all together.

I lean towards 1) since I feel like, within reason, more information is better than less information. WDYT?

tonyo: I'm fine with 1), but let's keep in mind that metadata provided by GCS artifact provider will also be "guessed" by GCS client in the "upload" command (if we don't do anything smart there, I mean).

lobsterkatie: Sorry - does this mean you think I should take it out of `gcsApi` and put it only in `gcsTarget`, or not? (Separate from the logging.)

tonyo: "it" -- you mean `detectContentType`? It can stay in `files`

lobsterkatie: No, sorry - I don't mean where the code for the method lives. I mean, currently we're doing the detection-and-content-type-setting in the generic GCS upload method, so any `.js` or `.map` file which gets uploaded either by the GCS artifact provider or by the GCS target has its content type set manually. My question is, should this only happen for the target, and not the artifact provider?

tonyo: Ah, hm, good question 🤔 
Let's leave the detection where it is for now, and we can separately discuss if we want to allow setting custom content types from command line as part  of `craft artifact upload`, because it's related.

lobsterkatie: > if we want to allow setting custom content types from command line as part of craft artifact upload

Oh, good point. I actually did have that in my mental spec for the upload command (since it's something zeus can do and which I've seen used by SDKs when they upload to zeus), but hadn't gotten as far as thinking about how that and this question would interact. Will add a note there about this and we can revisit at that point.

src/utils/__tests__/githubApi.test.ts
tonyo: It seems that we use different linters, then 😅

lobsterkatie: 🤔

image

vs

image

tonyo: Don't know why our `yarn lint` command doesn't catch this now.
And I think `return undefined` should also work.

lobsterkatie: > Don't know why our yarn lint command doesn't catch this now.

Yeah, that is odd.

> And I think return undefined should also work.

It doesn't like `return undefined`, but it's okay with `return`. But at that point, if we're bothering to return anything, why not return the thing the signature says we should return?

src/commands/publish.ts
lobsterkatie: Okay, this turns out to be more complicated than I realized. Took the detection out of the listing and left a TODO for the next round.

src/utils/__tests__/githubApi.test.ts
tonyo: Return values of async functions are wrapped in Promises automatically, so normally you don't have to return `Promise.resolve(...)`, value is enough.
In any case, please let's try to unify the linting configuration, it seems that your VS code does more than mine and than `yarn lint`.

lobsterkatie: > Return values of async functions are wrapped in Promises automatically

True, and fair point. Maybe it's the remnants of the Python dev still in me, LOL?

image

As for linting, if we're going to work on it at all we should probably use it as an excuse to migrate to`eslint`. (See https://medium.com/palantir/tslint-in-2019-1a144c2317a9.) I'll make an issue.

PR #92, opened 2020-06-01
test(windows): Fix tests on Windows

BYK:
Makes tests assume less about the platform :)

### DIFF
--- a/src/utils/__tests__/system.test.ts
+++ b/src/utils/__tests__/system.test.ts
@@ -166,7 +166,7 @@ describe('calculateChecksum', () => {
 
 describe('isExecutableInPath', () => {
   test('checks for existing executable', () => {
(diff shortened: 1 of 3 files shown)


src/utils/__tests__/system.test.ts
```diff
@@ -159,29 +165,27 @@ describe('calculateChecksum', () => {
 });
 
 describe('isExecutableInPath', () => {
-  test('checks for existing executable', async () => {
-    expect(hasExecutable('ls')).toBe(true);
+  test('checks for existing executable', () => {
+    expect(hasExecutable('git')).toBe(true);
```
jan-auer, line 169: It's probably safe to assume, but should we maybe even go safer and assume `node` here?

PR #93, opened 2020-06-04
feat(changelog): Add "auto" changeset policy

BYK:
This patch adds the "auto" changeset policy, which works as follows:

1. If there is already a changeset entry for the new version, use that (same as simple)
2. Else if there is an "Unreleased" changeset, rename that to the new version
3. Else, create a new version section with a default text
~~4. In all cases, create an empty "Unreleased" section at the top~~

Decided to remove the creation of an empty "Unreleased" section as that would end up in the final commit/tag which is pointless: you don't expect to see an "unreleased" section on your version commit/tag.

TODO:

1. ~~Add tests~~
2. Make "Unreleased title" configurable (follow up)
3. ~~Respect existing changelog files indentation and header style~~
4. Add a bunch more tests for `prepare` command, especially for dry run (follow up)

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -2,7 +2,7 @@ minVersion: "0.8.4"
 github:
   owner: getsentry
   repo: craft
-changelogPolicy: auto
+changelogPolicy: simple
 requireNames:
   - /^sentry-craft.*\.tgz$/
 targets:
--- a/.craft.yml
+++ b/.craft.yml
@@ -2,7 +2,7 @@ minVersion: "0.8.4"
 github:
   owner: getsentry
   repo: craft
-changelogPolicy: simple
+changelogPolicy: auto
 requireNames:
   - /^sentry-craft.*\.tgz$/
 targets:
--- a/src/commands/prepare.ts
+++ b/src/commands/prepare.ts
@@ -1,4 +1,4 @@
-import { existsSync, readFileSync } from 'fs';
+import { existsSync, promises as fsPromises } from 'fs';
 import { dirname, join, relative } from 'path';
 import * as shellQuote from 'shell-quote';
 // tslint:disable-next-line:no-submodule-imports
@@ -13,7 +13,13 @@ import {
 } from '../config';
 import { logger } from '../logger';
 import { ChangelogPolicy } from '../schemas/project_config';
-import { DEFAULT_CHANGELOG_PATH, findChangeset } from '../utils/changes';
+import {
+  DEFAULT_CHANGELOG_PATH,
+  DEFAULT_UNRELEASED_TITLE,
+  findChangeset,
+  removeChangeset,
+  prependChangeset,
+} from '../utils/changes';
 import {
   ConfigurationError,
   handleGlobalError,
@@ -360,7 +366,7 @@ async function checkForExistingTag(
  * @param changelogPolicy One of the changelog policies, such as "none", "simple", etc.
  * @param changelogPath Path to the changelog file
  */
-async function checkChangelog(
+async function prepareChangelog(
   newVersion: string,
   changelogPolicy: ChangelogPolicy = ChangelogPolicy.None,
   changelogPath: string = DEFAULT_CHANGELOG_PATH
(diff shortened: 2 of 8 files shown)


tonyo:
Looks good so far, thanks!

Another quick TODO would be to update README.md

src/utils/changes.ts
```diff
@@ -4,6 +4,9 @@ import { getVersion } from './version';
  * Path to the changelog file in the target repository
  */
 export const DEFAULT_CHANGELOG_PATH = 'CHANGELOG.md';
+export const DEFAULT_UNRELEASED_TITLE = 'Unreleased';
+export const DEFAULT_CHANGESET_BODY = '- No documented changes.';
```
jan-auer, line 8: At a first glance, I was thinking: Can these two be private?

Since you're planning to make those configurable later, however, would it make sense to move them into `prepare.ts` and have the defaulting there? Prepare has access to the configuration, and so most likely that logic will have to move anyway.

src/utils/changes.ts
```diff
@@ -24,22 +35,57 @@ export interface Changeset {
...
+ * @param predicate A callback that takes the found title and the expected title
+ *                  and returns true if they match, false otherwise
+ * @returns A ChangesetLoc object where "start" has the matche for the header,
+ *          and "end" has the match for the next header so the contents
+ *          inbetween can be extracted
+ */
+export function locateChangeset(
+  markdown: string,
+  header: string,
+  predicate: (match: string, expected: string) => boolean = (a, b) => a === b
```
jan-auer, line 70: Passing `header` as second parameter seems a bit redundant -- can you just pass the match? You might need to have the default explicitly in the function body, but as a nice side effect that would also trim down the function signature significantly.

src/utils/changes.ts
```diff
@@ -4,6 +4,9 @@ import { getVersion } from './version';
  * Path to the changelog file in the target repository
  */
 export const DEFAULT_CHANGELOG_PATH = 'CHANGELOG.md';
+export const DEFAULT_UNRELEASED_TITLE = 'Unreleased';
+export const DEFAULT_CHANGESET_BODY = '- No documented changes.';
+const HEADER_REGEX = /^(?:( *)## *([^\n]+?) *#*|^([^\n]+)\n *(?:-){2,}) *(?:\n+|$)/gm;
```
jan-auer, line 9: Sharing regexes with `/g` can be dangerous since they remember state between executions. This is why it was previously instantiated in a function body rather than gobally. Are you sure that this does not misbehave?

src/commands/prepare.ts
```diff
@@ -370,35 +376,81 @@ async function checkChangelog(
...
+    throw new ConfigurationError(`Invalid changelog path: "${changelogPath}"`);
+  }
+
+  if (!existsSync(relativePath)) {
+    throw new ConfigurationError(
+      `Changelog does not exist: "${changelogPath}"`
+    );
+  }
+
+  let changelogString = (await fsPromises.readFile(relativePath)).toString();
```
jan-auer, line 402: Just out of curiosity - is there a reason to switch to the promise version over `readFileSync`?

src/utils/changes.ts
BYK: Made them private for now. Will do the actual moving when we implement the configuration options.

src/utils/changes.ts
BYK: Eh, I'll give this another pass as the addition of `predicate` was closer to the end so it can probably use some pruning.

src/utils/changes.ts
BYK: I did explicitly set its `lastIndex` property before each run that said there's no reason for this to be a module global. In an earlier iteration I was reusing this but then I abstracted that logic out into `locateChangeset`. Will move the regex back into the function.

src/commands/prepare.ts
BYK: Just want to keep the good hygiene of doing all I/O async.

BYK:
@tonyo @jan-auer there's a slight logic change so see my notes in the PR description please.

jan-auer:
Random thought for the future: We could add a stage to `craft prepare` where it shows the changelog that it selected or created on the command line and offer to optionally edit it (similar to `gh pr create`). Even without editing, this could be a useful validation step in the publishing process.

BYK:
>Random thought for the future: We could add a stage to craft prepare where it shows the changelog that it selected or created on the command line and offer to optionally edit it (similar to gh pr create). Even without editing, this could be a useful validation step in the publishing process.

Love the idea, will definitely give this a shot after this PR lands.

>Wouldn't it be safer to assume that with "auto" policy we always want a non-empty changelog (either in the Unreleased section, or in the existing version entry)? And that not having a changelog is more of an exception, in which case one can just create a new entry with the default text manually?

The point of adding `auto` was to prepare Relay releases for automated CalVer versioning where a new version may not have any changes at all (as they are regularly and automatically released every month). That said I think in the broader context, your concern is valid.

Proposal: add a flag or a `strict-auto` mode to enforce the behavior you mentioned. Or make `auto` behave like you described and add a flag `--alow-empty-changelog` to keep the behavior in this PR.

jan-auer:
I agree with @BYK that it's important to also allow for releases without changes. 

It seems that there are two actual policies:
- Changelog is optional and will be filled in with "no changes".
- Changelog is required and an error is thrown when it's missing.

What if we just have two policies and allow for the "unreleased" logic in both of them. The only difference being what happens when there's no changelog?

.craft.yml
```diff
@@ -2,7 +2,7 @@ minVersion: "0.8.4"
 github:
   owner: getsentry
   repo: craft
-changelogPolicy: simple
+changelogPolicy: auto
```
tonyo, line 5:
```suggestion
changelogPolicy: simple
```

tonyo: I'd like to keep it "simple" for this repo for now.

BYK:
>The only difference being what happens when there's no changelog?

So I think you are advocating for adding an `--allow-empty-changelog` flag, but effective on both `simple` and `auto` policies?

I think this makes sense. I imagine it would work as follows:

For `simple`:

1. If there is no changelog entry matching the version passed, stop - throw exception
2. If there is a matching non-empty changelog entry for the version, continue
3. If there is a matching changelog entry for the version but it is empty:
  a. Stop, and throw and exception (default)
  b. Automatically add the default body (?) to the entry and continue (`--allow-empty-changelog` set)

For `auto`:

1. Item 1 above in `simple` does not throw an exception when `--allow-empty-changelog` but creates an entry automatically.

Makes sense? We just need to figure out whether we want to populate the changelog entry with the default text automatically in `simple` mode.

jan-auer:
I was actually advocating for not having `simple` and `auto` but rather `required` and `optional` instead of an additional flag.  Both with the same logic regarding "Unreleased".  `simple` is currently a strict subset of `auto` and its functionality is fully backward compatible.

Adding a separate toggle for this also makes sense to me. If we do this, then let's please add it to the config, as this should be enforceable.

PR #95, opened 2020-06-08
feat(target): Add "docker" target

BYK:
This change adds a "docker" target which pulls an already built
source image tagged with the revision being released, tags it with
the new version, and pushes to a new target.

### DIFF
--- a/README.md
+++ b/README.md
@@ -776,6 +776,39 @@ targets:
     specPath: MyProject.podspec
 ```
 
+### Docker (`docker`)
+
+Pulls an existing source image tagged with the revision SHA, and then pushed it
+to a new target tagged with the released version. No release
+assets are required for this target except for the source image at the provided
+source image location.
+
+**Environment**
+
+`docker` executable (or something equivalent) must be installed on the system.
+
+| Name                    | Description                               |
+| ----------------------- | ----------------------------------------- |
+| `DOCKER_USERNAME`       | The username for the Docker registry      |
+| `DOCKER_PASSWORD`       | The password for the Docker registry      |
+| `DOCKER_BIN`            | **optional**. Path to `pod` executable.   |
+
+**Configuration**
+
+| Option     | Description                                  |
+| ---------- | -------------------------------------------- |
+| `source`   | Path to the source Docker image to be pulled |
+| `target`   | Path to the target Docker image to be pushed |
+
+**Example**
+
+```yaml
+targets:
+  - name: cocoapods
+    source: us.gcr.io/sentryio/craft
+    target: getsentry/craft
+```
+
 ## Integrating Your Project with `craft`
 
 Here is how you can integrate your GitHub project with `craft`:
--- a/README.md
+++ b/README.md
@@ -781,7 +781,10 @@ targets:
 Pulls an existing source image tagged with the revision SHA, and then pushed it
 to a new target tagged with the released version. No release
 assets are required for this target except for the source image at the provided
-source image location.
+source image location so it would be a good idea to add a status chcek that
+ensures the source image exists, otherwise `craft publish` will fail at the
+`docker pull` step, causing an interrupted publish. This is an issue for other,
+non-idempotent targets, not for the Docker target.
 
 **Environment**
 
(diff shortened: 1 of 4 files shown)


BYK:
Not adding tests as there are no tests for most of the targets and the test would essentially replicate the implementation (as there are no edge cases to cover, or even too much complexity)

src/targets/docker.ts
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * Command to launch docker
+ */
+const DOCKER_BIN = process.env.DOCKER_BIN || DEFAULT_DOCKER_BIN;
+
+/** Options for "docker" target */
+export interface DockerTargetOptions extends TargetConfig {
+  username: string;
+  password: string;
+  /** Source image path, like `us.gcr.io/sentryio/carft` */
```
tonyo, line 21:
```suggestion
  /** Source image path, like `us.gcr.io/sentryio/craft` */
```
🙈

src/targets/docker.ts
```diff
@@ -0,0 +1,121 @@
...
+    return spawnProcess(DOCKER_BIN, ['push', target], {}, { showStdout: true });
+  }
+
+  /**
+   * Pushes a source image to Docker Hub
+   *
+   * @param version The new version
+   * @param revision The SHA revision of the new version
+   */
+  public async publish(_version: string, revision: string): Promise<any> {
```
tonyo, line 114:
```suggestion
  public async publish(version: string, revision: string): Promise<any> {
```

src/targets/docker.ts
```diff
@@ -0,0 +1,121 @@
...
+    await this.login();
+    await this.pull(revision);
+    await this.push(revision, _version);
```
tonyo, line 117:
```suggestion
    await this.push(revision, version);
```

src/targets/docker.ts
tonyo: `_version` was there just when it was unused, IIRC.

BYK: Yup, oversight. Thanks for the heads up :)

BYK:
>Probably stating the obvious, but for this target to work, the system that builds Docker images (GCB, Docker Hub, Travis, etc.) should report its status to GitHub as a commit status check (or whatever it's called now), so that craft publish (with GitHub as status provider configured) can pick it up and wait until the image is ready.

I mean, that would be _ideal_ but if the image doesn't exist, I just expect Craft to fail as `docker pull` will error out. Makes sense?

tonyo:
> I mean, that would be ideal but if the image doesn't exist, I just expect Craft to fail as docker pull will error out. Makes sense?

Yes, it will fail, it's just not all the targets in craft are currently idempotent, meaning that if the publishing process fails on this target, it'll require some human involvement to apply the remaining targets. 
So ideally, one should make sure that all required artifacts (be it files or docker images) are present before running `publish`. For files, craft can do that, but in this case there's no artifact we upload, so the idea is to check that Docker build finished successfully.

src/targets/docker.ts
jan-auer: Random spam: Let's rename this repo to carft. I really like the sound of it. So quirky :)

README.md
```diff
@@ -776,6 +776,48 @@ targets:
     specPath: MyProject.podspec
 ```
 
+### Docker (`docker`)
+
+Pulls an existing source image tagged with the revision SHA, and then pushed it
+to a new target tagged with the released version. No release
+assets are required for this target except for the source image at the provided
+source image location so it would be a good idea to add a status chcek that
```
tonyo, line 784:
```suggestion
source image location so it would be a good idea to add a status check that
```

README.md
```diff
@@ -776,6 +776,48 @@ targets:
...
+
+**Environment**
+
+`docker` executable (or something equivalent) must be installed on the system.
+
+| Name                    | Description                               |
+| ----------------------- | ----------------------------------------- |
+| `DOCKER_USERNAME`       | The username for the Docker registry      |
+| `DOCKER_PASSWORD`       | The password for the Docker registry      |
+| `DOCKER_BIN`            | **optional**. Path to `pod` executable.   |
```
tonyo, line 797:
```suggestion
| `DOCKER_BIN`            | **optional**. Path to `docker` executable.   |
```

README.md
```diff
@@ -776,6 +776,48 @@ targets:
...
+| Option     | Description                                  |
+| ---------- | -------------------------------------------- |
+| `source`   | Path to the source Docker image to be pulled |
+| `target`   | Path to the target Docker image to be pushed |
+
+**Example**
+
+```yaml
+targets:
+  - name: cocoapods
```
tonyo, line 810:
```suggestion
  - name: docker
```

PR #96, opened 2020-06-08
fix(logger): Logger should respect log level from env file

### LINKED ISSUE #82: CRAFT_LOG_LEVEL is not picked from environment files
`CRAFT_LOG_LEVEL` only applies when it's added to the environment beforehand, i.e. this line is not picked from Craft env files:

`export CRAFT_LOG_LEVEL=debug`

BYK:
Fixes #82.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,7 +2,7 @@
 import * as once from 'once';
 import * as yargs from 'yargs';
 
-import { logger } from './logger';
+import { logger, init as initLogger } from './logger';
 import { readEnvironmentConfig } from './utils/env';
 import { isDryRun } from './utils/helpers';
 import { hasNoInput, setNoInput } from './utils/noInput';
@@ -61,6 +61,8 @@ function main(): void {
   printVersion();
 
   readEnvironmentConfig();
+
+  initLogger();
 
   initSentrySdk();
 
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -74,13 +74,20 @@ export function setLogLevel(logLevel: LOG_LEVEL): void {
   consola.level = logLevel;
 }
 
+let initialized = false;
+
 /**
  * Initialize and return the logger
  * @param [logLevel] The desired logging level
  */
 export function init(logLevel?: LOG_LEVEL): Logger {
+  if (initialized) {
+    consola.warn('Logger already initialized, ignoring duplicate init.');
+  }
+
   setLogLevel(logLevel !== undefined ? logLevel : getLogLevelFromEnv());
   consola.reporters.push(new SentryBreadcrumbReporter());
+  initialized = true;
   return consola;
 }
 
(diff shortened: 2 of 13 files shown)


src/logger.ts
```diff
@@ -56,15 +57,25 @@ class SentryBreadcrumbReporter {
...
 
-consola.level = getLogLevel();
-consola.reporters.push(new SentryBreadcrumbReporter());
-
 // tslint:disable-next-line: variable-name
 export type Logger = typeof consola;
+
+/**
+ * Initialize and return the logger
+ * @param [logLevel] The desired loggin level
```
tonyo, line 70:
```suggestion
 * @param [logLevel] The desired logging level
```

src/index.ts
```diff
@@ -62,6 +62,8 @@ function main(): void {
 
   readEnvironmentConfig();
 
+  initLogger();
```
tonyo, line 65: Since it's still useful to have some debug logs for some operations before we read the configs, I suggest we initialize the logger twice: as soon as possible (e.g. as it was before, at import time), and then also here, after reading the environment configuration. With this approach, you can still get debug outputs for early operations if you define `CRAFT_LOG_LEVEL=debug` in your environment manually.

BYK: >With this approach, you can still get debug outputs for early operations if you define CRAFT_LOG_LEVEL=debug in your environment manually.

Is that how `consola` works? I wouldn't expect old messages to be displayed after initializing the logger. Also what should be the default logging value? Right now since this relies on a global that is being shared, logger is still active, just at the default logging level.

tonyo: I don't expect it to reprint the old messages, but if we run `initLogger` at the import time, it will allow us to see debug/trace output for `checkForUpdates`, `printVersion`, and `readEnvironmentConfig` functions, based on `CRAFT_LOG_LEVEL` that _might_ be present in the environment. 
Then, when we run `initLogger` here for the second time, it picks up the log level value from config files and uses it (if anything is there). 

The only non-idempotent thing in `initLogger` is  `consola.reporters.push(new SentryBreadcrumbReporter());`, but after it is fixed, we can call `initLogger` pretty much after every environment change, to make sure that we're using the latest `CRAFT_LOG_LEVEL` value.

Does this make sense?

BYK: Ah, yeah it is clear now and makes total sense. Will make the updates.

src/logger.ts
```diff
@@ -53,18 +57,50 @@ class SentryBreadcrumbReporter {
...
+  consola.level = logLevel;
 }
 
-consola.level = getLogLevel();
-consola.reporters.push(new SentryBreadcrumbReporter());
+/**
+ * Initialize and return the logger
+ * @param [logLevel] The desired logging level
+ */
+export function init(logLevel?: keyof typeof LOG_LEVEL): Logger {
```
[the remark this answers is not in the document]
BYK: Lol, yeah. Over-engineering. Will fix.

PR #98, opened 2020-06-09
fix(crates): Detect stale upload caches and retry

### LINKED ISSUE #67: Crates.io Releases seem to suffer from race conditions
**Disclaimer**: I'm not 100% sure what's the issue here. When releasing [`symbolic`](https://github.com/getsentry/symbolic), I received the following error for the last couple of times:

```
ℹ info [crates] › Cloning getsentry/symbolic into /var/folders/7p/zvqpr1r53zv_b6tzlz7tw2vw0000gn/T/craft-crates-mDhCL3
ℹ info [crates] › Checking out submodules
ℹ info [crates] › Loading workspace information from /var/folders/7p/zvqpr1r53zv_b6tzlz7tw2vw0000gn/T/craft-crates-mDhCL3/Cargo.toml
✖ error Error: Process "cargo" errored with code 101
  STDOUT: cargo:
  STDERR:cargo:     Updating crates.io index
  cargo:    Packaging symbolic v7.2.0 (/var/folders/7p/zvqpr1r53zv_b6tzlz7tw2vw0000gn/T/craft-crates-mDhCL3)
  cargo: error: failed to prepare local package for uploading
  cargo:
  cargo: Caused by:
  cargo:   failed to select a version for the requirement `symbolic-demangle = "^7.2.0"`
  cargo:   candidate versions found which didn't match: 7.1.1, 6.1.4, 6.1.3, ...
  cargo:   location searched: crates.io index
  cargo: required by package `symbolic v7.2.0 (/var/folders/7p/zvqpr1r53zv_b6tzlz7tw2vw0000gn/T/craft-crates-mDhCL3)`
  cargo:
```

The error message indicates that a workspace dependency, `symbolic-demangle`, hadn't been published at the time of publishing `symbolic`. When checking crates.io, however, that crate is there. I think this might be some lag in caches.

jan-auer:
The publishing errors also happen locally. When publishing a crate with `--no-verify`, it seems that the crate caches can go out of date and `cargo` will fail to publish. We can detect this and retry publishing with exponential backoff. 

I even ran into this when publishing manually with sufficient time between the workspace crates. The error always looks like this:

```
cargo publish --no-verify
    Updating crates.io index
   Packaging symbolic v7.3.4 (/private/tmp/symbolic)
error: failed to prepare local package for uploading

Caused by:
  failed to select a version for the requirement `symbolic-symcache = "^7.3.4"`
  candidate versions found which didn't match: 7.3.3, 7.3.2, 7.3.1, ...
  location searched: crates.io index
required by package `symbolic v7.3.4 (/private/tmp/symbolic)`
```

It is unclear what the root cause of this is. When cargo updates its index, it runs a "fast path" request against the GitHub API which seems to get cached. Waiting for 20-30 seconds seems sufficient to invalidate those caches.

Fixes #67


jan-auer:
I'm open to suggestions regarding tests :)

src/targets/crates.ts
```diff
@@ -202,9 +209,23 @@ export class CratesTarget extends BaseTarget {
...
+      ...process.env,
+      CARGO_REGISTRY_TOKEN: this.cratesConfig.apiToken,
+    };
+
+    for (let i = 0; i <= 1; i++) {
+      try {
+        return await spawnProcess(CARGO_BIN, args, { env });
+      } catch (e) {
+        if (i == 0 && e.message.includes(VERSION_ERROR)) {
+          logger.debug('Potential stale cache detected, trying again...');
```
tonyo, line 222:
```suggestion
          logger.warn('Potential stale cache detected, trying again...');
```

tonyo: To make it more visible.

src/targets/crates.ts
```diff
@@ -202,9 +209,23 @@ export class CratesTarget extends BaseTarget {
...
+    const env = {
+      ...process.env,
+      CARGO_REGISTRY_TOKEN: this.cratesConfig.apiToken,
+    };
+
+    for (let i = 0; i <= 1; i++) {
+      try {
+        return await spawnProcess(CARGO_BIN, args, { env });
+      } catch (e) {
+        if (i == 0 && e.message.includes(VERSION_ERROR)) {
```
tonyo, line 221:
```suggestion
        if (i === 0 && e.message.includes(VERSION_ERROR)) {
```
Should fix the linting error.

src/targets/crates.ts
jan-auer: Just double-checking - are you certain? I'd rather swallow this and only if it occurs again emit an error.

src/targets/crates.ts
jan-auer: Good point, thanks!

jan-auer:
I'd like to verify this once with an actual release. I'll check back in once it worked and then we can merge this.

src/targets/crates.ts
tonyo: It's just to detect these cases in case the user forgot to set log level to debug.

> Usually the entire crate publishing stage is silent

We can log something else to make it less lonely?

jan-auer: OK, let me add some info logs and also the crate name in the warning.

jan-auer: Updated, how do you like the new messages?

tonyo: Bbbjutiful 👍

src/targets/crates.ts
```diff
@@ -202,9 +209,26 @@ export class CratesTarget extends BaseTarget {
...
-    return spawnProcess(CARGO_BIN, args, {
-      env: { ...process.env, CARGO_REGISTRY_TOKEN: this.cratesConfig.apiToken },
-    });
+    const env = {
+      ...process.env,
+      CARGO_REGISTRY_TOKEN: this.cratesConfig.apiToken,
+    };
+
+    logger.info(`Publishing ${crate.name}`);
+    for (let i = 0; i <= 1; i++) {
```
BYK, line 218: I think this for loop construct makes it harder to read the code than repeating the `return await spawnProcess(CARGO_BIN, args, { env });` part. If you don't want to repeat that, we can abstract that out  into its own function.

Or at least we should define `MAX_RETRIES=1` and do `i <= MAX_RETRIES`.

jan-auer: Good point, I’ll repeat for now.

jan-auer:
Unfortunately, that still did not help. Craft detects the condition successfully, but I'm still getting the error. I am not sure what the proper resolution is, because as soon as I try it manually, it just magically works. 

You can see that it even pulls a new version of the "crates.io index". Potentially increase the timeout?

```
...
ℹ info [crates] › Publishing symbolic
⚠ warn [crates] › Potential stale cache detected, trying again...
✖ error Error: Process "cargo" errored with code 101

  STDOUT: cargo:


  STDERR:cargo:     Updating crates.io index
  cargo:    Packaging symbolic v7.3.5 (/var/folders/7p/zvqpr1r53zv_b6tzlz7tw2vw0000gn/T/craft-crates-hJ3Y7K)
  cargo: error: failed to prepare local package for uploading
  cargo:
  cargo: Caused by:
  cargo:   failed to select a version for the requirement `symbolic-common = "^7.3.5"`
  cargo:   candidate versions found which didn't match: 7.3.4, 7.3.3, 7.3.2, ...
  cargo:   location searched: crates.io index
  cargo: required by package `symbolic v7.3.5 (/var/folders/7p/zvqpr1r53zv_b6tzlz7tw2vw0000gn/T/craft-crates-hJ3Y7K)`
  cargo:
```

src/targets/crates.ts
```diff
@@ -215,18 +222,15 @@ export class CratesTarget extends BaseTarget {
...
-          );
-          continue;
-        } else {
-          throw e;
-        }
+    try {
+      return await spawnProcess(CARGO_BIN, args, { env });
+    } catch (e) {
+      if (e.message.includes(VERSION_ERROR)) {
+        logger.warn(`Potential stale cache detected, trying again...`);
```
BYK, line 229:
```suggestion
        logger.warn(`Potential stale cache detected, trying again in ${RETRY_DELAY_MS}ms...`);
```

src/targets/crates.ts
```diff
@@ -215,18 +222,15 @@ export class CratesTarget extends BaseTarget {
...
+        await sleepAsync(RETRY_DELAY_MS);
+        return spawnProcess(CARGO_BIN, args, { env });
```
BYK, line 231: hah, skipping `await`, very clever optimization :)

jan-auer: It's closer to the original. However, I think I'll have to revert this and really spin for multiple times + increase the timeout. No idea what the root cause of this is.

BYK: Maybe one of these help:
 - https://github.com/rust-lang/cargo/issues/8273
 - https://github.com/rust-lang/cargo/issues/3377#issuecomment-361819305

jan-auer: Thanks, I tried that as well. Unfortunately, you can see in the above output that the crates index was in fact refreshed, but still didn’t contain the new version.

This leads me to believe that it’s server side caches that interfer with this, and the only way around would be larger delays and more retries.

BYK: What if we kept polling the API directly until we get a response (ie `https://crates.io/api/v1/crates/symbolic/7.3.5`)

jan-auer:
Ok, it does what it should do now. Can I haz a new approve before marge, please? 

Here is fully trustworthy proof: 

```
ℹ info [crates] › Publishing symbolic-common
ℹ info [crates] › Publishing symbolic-sourcemap
ℹ info [crates] › Publishing symbolic-unreal
ℹ info [crates] › Publishing symbolic-debuginfo
ℹ info [crates] › Publishing symbolic-demangle
ℹ info [crates] › Publishing symbolic-proguard
ℹ info [crates] › Publishing symbolic-minidump
ℹ info [crates] › Publishing symbolic-symcache
ℹ info [crates] › Publishing symbolic
⚠ warn [crates] › Publish failed, trying again in 2s...
⚠ warn [crates] › Publish failed, trying again in 4s...
⚠ warn [crates] › Publish failed, trying again in 8s...
⚠ warn [crates] › Publish failed, trying again in 16s...
ℹ info [crates] › Crates release complete
```

src/targets/crates.ts
```diff
@@ -227,9 +255,26 @@ export class CratesTarget extends BaseTarget {
...
+      CARGO_REGISTRY_TOKEN: this.cratesConfig.apiToken,
+    };
+
+    let delay = RETRY_DELAY_SECS;
+    logger.info(`Publishing ${crate.name}`);
+    for (let i = 0; i <= MAX_ATTEMPTS; i++) {
+      try {
+        return await spawnProcess(CARGO_BIN, args, { env });
+      } catch (e) {
+        if (i < MAX_ATTEMPTS && e.message.includes(VERSION_ERROR)) {
```
BYK, line 269: This differs from the for loop's condition, intentional? I find having two checks a bit confusing btw. Feels like this should be a while loop.

```js
const MAX_WAIT_SECS = 60;
let totalWait = 0;
let delay = RETRY_DELAY_SECS;

while (totalWait <= MAX_WAIT_SECS) {
  try {
    return await spawnProcess(CARGO_BIN, args, { env });
  catch (err) {
    if (!e.message.includes(VERSION_ERROR)) {
      throw err;
    }

    logger.warn(`Publish failed due to potentially stale cache. Trying again in ${delay}s...`);
    await sleepAsync(delay * 1000);
    totalWait += delay;
    delay *= RETRY_EXP_FACTOR;
  }
}
```

jan-auer: Yes this is intentional. The loop needs to to take another turn so that we get into the `else` branch. In your implementation, you do not throw the error in the last iteration.

I thought about a total wait time (which could be written as a `for` loop, too), but rather wanted a fixed number of retries. And then doing `RETRY_DELAY_SECS * Math.pow(RETRY_EXP_FACTOR, MAX_ATTEMPTS)` seemed to cumbersome.

BYK: >In your implementation, you do not throw the error in the last iteration.

Yes I do as I do not check the count or total wait in the if separately.

PR #103, opened 2020-06-15
feat(gha): Add GitHub Action for Craft

BYK:
Moves the definition from [getsentry/craft-action](https://github.com/getsentry/craft-action) to here.

See https://github.com/getsentry/onpremise/blob/master/.github/workflows/release.yml for an example use.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -8,6 +8,14 @@ inputs:
   version:
     description: 'The version number for the new release'
     required: true
+  no_merge:
+    description: 'Do not merge the release branch after publishing'
+    required: false
+    default: false
+  keep_branch:
+    description: 'Do not remove release branch after merging it'
+    required: false
+    default: false
   dry_run:
     description: 'Whether to not perform any mutations or not'
     required: false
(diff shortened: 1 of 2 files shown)


action.yml
```diff
@@ -0,0 +1,81 @@
...
+    TWINE_USERNAME: ${{ inputs.twine_username }}
+    TWINE_PASSWORD: ${{ inputs.twine_password }}
+    DOCKER_USERNAME: ${{ inputs.docker_username }}
+    DOCKER_PASSWORD: ${{ inputs.docker_password }}
+  args:
+    - '--no-input'
+    - ${{ inputs.action }}
+    - ${{ inputs.version }}
+    - ${{ inputs.no_merge && '--no-merge' || '' }}
+    - ${{ inputs.keep_branch && '--keep-branch' || '' }}
```
tonyo, line 81: To clarify, do we still need https://github.com/getsentry/craft-action after this PR is merged?

BYK: Nope, it will be baked and there'll be cake.

PR #110, opened 2020-06-15
build(ci): Have better defaults for CI environments

BYK:
See https://github.com/getsentry/relay/runs/773764935?check_suite_focus=true for context.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- build(ci): Have better defaults for CI environments (#110)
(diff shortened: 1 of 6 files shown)


BYK:
>Regardless though, have you thought about giving craft a writable home directory somehow? The same problem has occurred with setting up cargo, and I'm pretty sure we'll have more targets that will try to write to $HOME.

Yup, we may use `/tmp` for that. I think assuming `$HOME` to be writable may not be the best idea though. I had similar issues with a proprietary CI system in the past (assuming `$HOME` to be writable where it was not).

PR #112, opened 2020-06-18
feat(crates): Add noDevDeps option

jan-auer:
Cargo does not allow to publish crates with circular dependencies. In dev dependencies, they would theoretically be fine since they are not required to run or build the crate. We need them in the Rust SDK to provide proper doc comments in sub-crates. 

This PR adds a new `noDevDeps` option, which uses `cargo hack` to strip dev dependencies before publishing.

### DIFF
--- a/src/schemas/projectConfig.schema.ts
+++ b/src/schemas/projectConfig.schema.ts
@@ -151,6 +151,17 @@ const projectConfigJsonSchema = {
       },
       additionalProperties: false,
     },
+    cratesConfig: {
+      title: 'CratesTargetConfig',
+      description: 'Configuration options for the Crates target',
+      extends: { $ref: '#/definitions/targetConfig' },
+      properties: {
+        noDevDeps: {
+          type: 'boolean',
+        },
+      },
+      additionalProperties: false,
+    },
   },
 };
 
--- a/src/targets/__tests__/crates.test.ts
+++ b/src/targets/__tests__/crates.test.ts
@@ -22,11 +22,21 @@ function cratePackageToDependency(cratePackage: CratePackage): CrateDependency {
   };
 }
 
+function makeDev(dependency: CrateDependency): CrateDependency {
+  return {
+    ...dependency,
+    kind: 'dev',
+  };
+}
+
 describe('getPublishOrder', () => {
   process.env.CRATES_IO_TOKEN = 'xxx';
-  const target = new CratesTarget({}, new NoneArtifactProvider());
+  const target = new CratesTarget(
+    { noDevDeps: true },
+    new NoneArtifactProvider()
+  );
 
-  test('sorts crate packages properly', async () => {
+  test('sorts crate packages properly', () => {
     const packages = ['p1', 'p2', 'p3', 'p4'].map(cratePackageFactory);
     const [p1, p2, p3, p4] = packages;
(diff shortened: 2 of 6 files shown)


src/schemas/projectConfig.schema.ts
```diff
@@ -151,6 +151,17 @@ const projectConfigJsonSchema = {
...
+      title: 'CratesTargetConfig',
+      description: 'Configuration options for the Crates target',
+      extends: { $ref: '#/definitions/targetConfig' },
+      properties: {
+        noDevDeps: {
+          type: 'boolean',
+        },
+      },
+      additionalProperties: false,
+    },
```
tonyo, line 164: Hm, does this actually work?
Based on the comment above, I really doubt it.
But we can probably revisit configuration validation some time soon.

src/targets/crates.ts
```diff
@@ -196,12 +214,16 @@ export class CratesTarget extends BaseTarget {
    * @returns A promise that resolves when the upload has completed
    */
   public async publishPackage(crate: CratePackage): Promise<any> {
-    const args = [
-      'publish',
+    const args = this.config.noDevDeps
+      ? ['hack', 'publish', '--allow-dirty', '--no-dev-deps']
```
tonyo, line 218: Can we also install `cargo-hack` in Dockerfile, please?

jan-auer:
This still needs tests.

src/schemas/projectConfig.schema.ts
jan-auer: Not sure :) I only added it for completeness since I saw the others. Certainly doesn't hurt.

src/targets/crates.ts
jan-auer: Excellent suggestion, added!

jan-auer:
@tonyo should be ready now (if the Docker image builds). Added tests, too.

src/targets/__tests__/crates.test.ts
```diff
@@ -18,14 +18,25 @@ function cratePackageToDependency(cratePackage: CratePackage): CrateDependency {
...
+    ...dependency,
+    kind: 'dev',
   };
 }
 
 describe('getPublishOrder', () => {
   process.env.CRATES_IO_TOKEN = 'xxx';
-  const target = new CratesTarget({}, new NoneArtifactProvider());
+  const target = new CratesTarget(
+    { noDevDeps: true },
```
tonyo, line 35: nit: would be nice to add an interface for CratesTargetConfig (since it's not automatically generated now).

jan-auer: Do you have an example for where to put and enforce this? I looked into the other targets, but they all receive an `any` config.

tonyo: They do, but then we perform manual validation, e.g. https://github.com/getsentry/craft/blob/master/src/targets/npm.ts#L131-L158

jan-auer: Excellent, thanks. Added.

jan-auer:
This is the passing PR build: https://travis-ci.org/github/getsentry/craft/builds/699788879
@tonyo mind if we merge?

tonyo:
yow low

PR #114, opened 2020-06-23
fix(eslint): Address eslint warnings

AbhiPrasad:
This PR continues from https://github.com/getsentry/craft/pull/113 and aims to reduce all the eslint related warnings to 0.

269aef6 has the changes related to fixing `@typescript-eslint/no-unused-vars`
See: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unused-vars.md
- I added `{ argsIgnorePattern: '^_' }` to make sure that variables that start with `_` were fine to be unused, as this seems to be a pattern in the code base
- Caught some unnecessary imports of lodash though!

8539795 has the changes related to fixing `@typescript-eslint/explicit-module-boundary-types`
See: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/explicit-module-boundary-types.md
- This was mostly just typing exported functions, and using `Record<string, any>` to type config objects.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 - build: Migrate from tslint to eslint (#113)
+- fix: Add stronger types for module exports (#114)
 
 ## 0.10.1
 
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -15,5 +15,7 @@ module.exports = {
   rules: {
     '@typescript-eslint/no-explicit-any': 'off',
     'no-constant-condition': ['error', { checkLoops: false }],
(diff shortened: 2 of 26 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 - build: Migrate from tslint to eslint (#113)
+- fix: Add stronger types for module exports (#114)
```
AbhiPrasad, line 4: This is different than the PR title, but figured it is the only change that could matter for a CHANGELOG.

Open to change it though.

tonyo: That's fine 👍

PR #117, opened 2020-07-17
fix(github): Don't fail when there are queued check suites

### LINKED ISSUE #111: bug: GitHub status check does not wait for builds to start
Sometimes, when Craft is faster than GitHub and status check integrations, Craft cannot find any relevant checks, even in pending state and immediately fails: https://github.com/getsentry/onpremise/runs/777959403?check_suite_focus=true

We should figure out a way to address this issue either by trying a few times when there are no checks listed or trying to find actual integrations and match that number with the checks reported.

BYK:
When the check suites are queued but the check runs haven't started yet, Craft immediately fails. With this patch, it also checks the Check Suites API to see if there are any pending check suites.

Fixes #111.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-# Unreleased
+## Unreleased
 
 - build: Migrate from tslint to eslint (#113)
 - fix: Add stronger types for module exports (#114)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,10 @@
 # Changelog
 
+# Unreleased
+
 - build: Migrate from tslint to eslint (#113)
 - fix: Add stronger types for module exports (#114)
+- fix(github): Don't fail when there are queued check suits (#117)
 
 ## 0.10.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - build: Migrate from tslint to eslint (#113)
 - fix: Add stronger types for module exports (#114)
-- fix(github): Don't fail when there are queued check suits (#117)
+- fix(github): Don't fail when there are queued check suites (#117)
 
 ## 0.10.1
 
--- a/src/status_providers/github.ts
+++ b/src/status_providers/github.ts
@@ -40,10 +40,16 @@ export class GithubStatusProvider extends BaseStatusProvider {
     }
 
     // There are two commit status flavours we have to consider:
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,7 +1,10 @@
 # Changelog
 
+## Unreleased
+
 - build: Migrate from tslint to eslint (#113)
 - fix: Add stronger types for module exports (#114)
+- fix(github): Don't fail when there are queued check suits (#117)
```
jan-auer, line 7:
```suggestion
- fix(github): Don't fail when there are queued check suites (#117)
```

src/status_providers/github.ts
```diff
@@ -81,9 +88,18 @@ export class GithubStatusProvider extends BaseStatusProvider {
       ) {
         // Edge case, this is what GitHub returns when there are no registered legacy checks.
         logger.debug('No legacy checks detected, checking for runs...');
+        const hasPendingActiveSuites = revisionCheckSuites.check_suites.some(
+          suite =>
+            // Need the any cast as octokit lacks this prop in its types
+            (suite as any).latest_check_runs_count > 0 &&
+            suite.status === 'queued'
```
jan-auer, line 95: _nit_: This is done in two places and might qualify for a small helper function given the comment.

src/status_providers/github.ts
```diff
@@ -254,6 +283,34 @@ export class GithubStatusProvider extends BaseStatusProvider {
     return revisionStatus;
   }
 
+  /**
+   * Gets revision checks from GitHub Check runs API
+   *
+   * API docs:
+   * https://developer.github.com/v3/checks/suites/#list-check-suites-for-a-git-reference
```
jan-auer, line 290: Note that this requires the `checks:read` permission. I believe, we already require the generic `read` permission, so this shouldn't pose as an issue.

BYK:
>@BYK based on this patch, could we get rid of the Check Runs API and use the status and conclusion of the suites instead?

Possibly, I need to dig and debug more.

Btw. I finally found the docs for `latest_check_runs_count`: https://developer.github.com/changes/7/#changes-to-the-checks-rest-api (emphasis mine)

>Previously, *when querying a check suite, the REST API would return a key named unique_check_runs_count to indicate the quantity of check runs that had run as part of the latest push*. Now, this key is called latest_check_runs_count to better indicate which types of check runs are being counted.

It indeed sounds like what we need/want here.

PR #121, opened 2020-08-21
feat: Github Artifact Provider

HazAT:
aka. we can kill Zeus

giphy-4

You need to use Github actions to upload your artifacts, here is one example from the JS repo:

```yaml
  job_artifacts:
    name: Artifacts Upload
    needs: job_build
    runs-on: ubuntu-latest
    if: "contains(github.ref, 'release/')"
    steps:
      - uses: actions/checkout@v2
      - uses: actions/setup-node@v1
      - uses: actions/cache@v2
        with:
          path: |
            ${{ github.workspace }}/node_modules
            ${{ github.workspace }}/packages/**/node_modules
            ${{ github.workspace }}/packages/**/build
            ${{ github.workspace }}/packages/**/dist
            ${{ github.workspace }}/packages/**/esm
          key: ${{ runner.os }}-${{ github.sha }}
      - name: Pack
        run: yarn pack:changed
      - name: Archive Artifacts
        uses: actions/upload-artifact@v2
        with:
          name: ${{ github.sha }}
          path: |
            ${{ github.workspace }}/packages/browser/build/**
            ${{ github.workspace }}/packages/integrations/build/**
            ${{ github.workspace }}/packages/apm/build/**
            ${{ github.workspace }}/packages/tracing/build/**
            ${{ github.workspace }}/packages/**/*.tgz
```

the important part is this:

```yaml
- name: Archive Artifacts
        uses: actions/upload-artifact@v2
        with:
          name: ${{ github.sha }}
          path: |
            ${{ github.workspace }}/packages/browser/build/**
            ${{ github.workspace }}/packages/integrations/build/**
            ${{ github.workspace }}/packages/apm/build/**
            ${{ github.workspace }}/packages/tracing/build/**
            ${{ github.workspace }}/packages/**/*.tgz
```

using `actions/upload-artifact@v2` gha + setting the name to the `${{ github.sha }}` is the key since there is no other way to uniquely identify uploaded artifacts.

config in `craft.yaml`
```yaml
statusProvider:
  name: github
artifactProvider:
  name: github
```

### DIFF
--- a/src/artifact_providers/github.ts
+++ b/src/artifact_providers/github.ts
@@ -1,5 +1,8 @@
 import * as Github from '@octokit/rest';
 import * as _ from 'lodash';
+import * as fs from 'fs';
+import * as request from 'request';
+import * as path from 'path';
 
 import {
   ArtifactProviderConfig,
@@ -10,6 +13,13 @@ import { getGlobalGithubConfig } from '../config';
 import { logger as loggerRaw } from '../logger';
 import { GithubGlobalConfig } from '../schemas/project_config';
 import { getGithubClient } from '../utils/githubApi';
+import {
+  detectContentType,
+  scan,
+  withTempFile,
+  withTempDir,
+} from '../utils/files';
+import { extractZipArchive } from '../utils/system';
 
 const logger = loggerRaw.withScope(`[artifact-provider/github]`);
 
@@ -31,7 +41,7 @@ interface ArtifactList {
 /**
  * Github artifact provider
  */
-export class HithubArtifactProvider extends BaseArtifactProvider {
+export class GithubArtifactProvider extends BaseArtifactProvider {
   /** Github client */
   public readonly github: Github;
 
@@ -48,19 +58,12 @@ export class HithubArtifactProvider extends BaseArtifactProvider {
    * @inheritDoc
    */
   public async doDownloadArtifact(
-    artifact: RemoteArtifact,
-    downloadDirectory: string
+    _artifact: RemoteArtifact,
+    _downloadDirectory: string
   ): Promise<string> {
     // TODO: Return list of paths
-    await this.github.request(
-      '/repos/{owner}/{repo}/actions/artifacts/{artifact_id}/{archive_format}',
-      {
-        owner: 'octocat',
-        repo: 'hello-world',
-        artifact_id: 42,
-        archive_format: 'archive_format',
-      }
-    );
+
+    return '';
   }
 
   /**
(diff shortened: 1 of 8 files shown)


src/artifact_providers/github.ts
```diff
@@ -0,0 +1,154 @@
...
+   */
+  protected async doListArtifactsForRevision(
+    revision: string
+  ): Promise<RemoteArtifact[]> {
+    const { repoName, repoOwner } = this.config;
+    logger.debug(
+      `Fetching Github artifacts for ${repoOwner}/${repoName}, revision ${revision}`
+    );
+    const artifactResponse = ((
+      await this.github.request('GET /repos/{owner}/{repo}/actions/artifacts', {
```
tonyo, line 80: This returns a paginated response, right?
If yes, this might not work if the desired artifact is not on the first page.

src/artifact_providers/github.ts
```diff
@@ -0,0 +1,154 @@
...
+      throw new Error(`Failed to discover any artifacts`);
+    }
+
+    // We need to find the archive where name maches the revision
+    const foundArtifacts = _.filter(
+      artifactResponse.artifacts,
+      artifact => artifact.name === revision
+    );
+
+    if (foundArtifacts.length !== 1) {
```
tonyo, line 96: Curious if there's a chance we can have two or more artifacts with the same name?

src/artifact_providers/github.ts
HazAT: Correct, they are sorted by upload date but still, it could happen, so I will add paging.

src/artifact_providers/github.ts
HazAT: Yes if you have multiple jobs I guess it can happen, I will also make this more robust.

src/artifact_providers/github.ts
```diff
@@ -93,35 +102,37 @@ export class GithubArtifactProvider extends BaseArtifactProvider {
       artifact => artifact.name === revision
     );
 
-    if (foundArtifacts.length !== 1) {
+    if (foundArtifacts.length === 0) {
+      if (artifactResponse.total_count > per_page) {
```
tonyo, line 106: Looks like this have a potential of entering the infinite loop if the condition holds initially, because both `total_count` and `per_page` don't change.

src/artifact_providers/github.ts
```diff
@@ -0,0 +1,215 @@
...
+  ): Promise<ArtifactItem> {
+    const { repoName, repoOwner } = this.config;
+    const per_page = 100;
+
+    // https://docs.github.com/en/free-pro-team@latest/rest/reference/actions#artifacts
+    const artifactResponse = ((
+      await this.github.request('GET /repos/{owner}/{repo}/actions/artifacts', {
+        owner: repoOwner,
+        repo: repoName,
+        per_page,
```
tonyo, line 91: Looking at the API, I think it also expects the page number

tonyo (changes requested):

tonyo (approved):
LET'S DO THIS

HazAT:
giphy-3

Will test before merging, only will merge it once confirmed the whole flow works.

PR #125, opened 2020-10-07
feat(docker): Add sourceFormat & targetFormat options

BYK:
These options became necessary for us to be able to publish Py2 and Py3 versions of Sentry images at the same time, in the same release.

### DIFF
--- a/src/targets/docker.ts
+++ b/src/targets/docker.ts
@@ -2,6 +2,7 @@ import { logger as loggerRaw } from '../logger';
 import { TargetConfig } from '../schemas/project_config';
 import { BaseArtifactProvider } from '../artifact_providers/base';
 import { ConfigurationError } from '../utils/errors';
+import { renderTemplateSafe } from '../utils/strings';
 import { checkExecutableIsPresent, spawnProcess } from '../utils/system';
 import { BaseTarget } from './base';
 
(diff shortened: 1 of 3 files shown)


HazAT (approved):
LGTM :)

src/targets/docker.ts
```diff
@@ -55,10 +60,13 @@ export class DockerTarget extends BaseTarget {
         )
       );
     }
+
     return {
       password: process.env.DOCKER_PASSWORD,
       source: this.config.source,
       target: this.config.target,
+      sourceTemplate: this.config.sourceFormat || '{{source}}:{{revision}}',
+      targetTemplate: this.config.targetFormat || '{{target}}:{{release}}',
```
jan-auer, line 69: _nit_: Should we make this `{{version}}`? I believe that's the terminology we mostly use across craft and repos. The "release" is what we create as a result.

jan-auer: For example, we have this sort of interpolation in the `brew` target, where we support the variables `version`, `revision`, and `checksum`: 

https://github.com/getsentry/craft/blob/2c6c0bb523941abc30002f02f7a4d3fd70ced2fc/src/targets/brew.ts#L176-L178

BYK: Good point, fix coming up.

BYK: https://github.com/getsentry/craft/pull/126

PR #127, opened 2020-10-08
fix(docker): Disable HTML escaping in image names by default

BYK:
We used Mustache templates that were already available for image name formatting. Musache escapes all HTML-unsafe characters by default which rightly confused the Docker CLI client. This patch uses the `{{{varName}}}` pattern for the default format config to disable escaping.

### DIFF
--- a/README.md
+++ b/README.md
@@ -803,11 +803,11 @@ non-idempotent targets, not for the Docker target.
 
 **Configuration**
 
-| Option         | Description                                                          |
-| -------------- | -------------------------------------------------------------------- |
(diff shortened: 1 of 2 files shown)


BYK:
brace yourself

_(pun intended)_

HazAT (approved):

jan-auer (approved):
Sorry for the sloppy reviews :)

BYK:
>Sorry for the sloppy reviews :)

This is what happens when you don't have tests and nobody calls you out for them ;)

PR #128, opened 2020-10-08
feat(targets): Add optional `id` field to target config

BYK:
Craft allows using multiple instances of the same target with different (or same!) configurations but it assumes `name` field is a unique identifier. This PR adds an optional `id` field to distinguish multiple targets with the same name such as multiple Docker images to publish.


HazAT (approved):
Makes sense, good addition

PR #135, opened 2020-11-17
ref: Delete obsolete docs

BYK:
The docs directory here was a stub and was not used. This was leading to confusion and also triggering Dependapot alerts due to stale dependency definitions inside. This PR removes the folder in the hopes of proper docs in the future.


tonyo:
The docs directory is used to dogfood `gh-pages` target. 
What confusion are we talking about here?

BYK:
The actual docs live inside the README file and this directory seems unmaintained and empty of useful content.

I'm okay abandoning this PR but then we'd better make this intent more visible and update those vulnerable dependencies to stop getting dependabot alerts.

tonyo:
The docs are responsible for rendering this page: https://getsentry.github.io/craft/
I'm ok with adding the docs directory to ignore list of dependabot.

Issue #136, opened 2020-11-23
bug: `-t` flag does not accept the new `targetId[targetName]` format


BYK:
Initially reported by @jan-auer on Slack:

```shell
craft publish 0.3.0 -t gcs -t 'release[docker]' -t 'latest[docker]'
Invalid values:
  Argument: target, Given: "release[docker]", "latest[docker]", Choices: "brew", "cocoapods", "crates", "docker", "gcs", "gem", "gh-pages", "github", "npm", "nuget", "pypi", "registry", "all", "none"
```

This is a fallout from #128.

BYK:
Fixed by #137.

PR #137, opened 2020-11-23
fix(publish): Only allow valid target ids for -t

BYK:
With this patch, only targets defined in `.craft.yml` are accepted
in the `-t` flag, also fixing #136.

<pre>
craft publish --help
i info craft 0.13.2
dist publish NEW-VERSION

🛫 Publish artifacts

Positionals:
  NEW-VERSION  Version to publish                            [string] [required]

Options:
  --no-input         Suppresses all user prompts      [boolean] [default: false]
  --dry-run          Dry run mode: do not perform any real actions
                                                      [boolean] [default: false]
  --target, -t       Publish to this target
              [string] <b>[choices: "npm", "gh-pages", "gcs", "registry", "github",
            "docker[release]", "docker[latest]", "all", "none"] [default: "all"]</b>
  --rev, -r          Source revision (git SHA or tag) to publish (if not release
                     branch head)                                       [string]
  --no-merge         Do not merge the release branch after publishing
                                                      [boolean] [default: false]
  --keep-branch      Do not remove release branch after merging it
                                                      [boolean] [default: false]
  --keep-downloads   Keep all downloaded files        [boolean] [default: false]
  --no-status-check  Do not check for build status    [boolean] [default: false]
  -v, --version      Show version number                               [boolean]
  -h, --help         Show help                                         [boolean]
</pre>

### DIFF
--- a/src/targets/index.ts
+++ b/src/targets/index.ts
@@ -1,3 +1,4 @@
+import { TargetConfig } from 'src/schemas/project_config';
 import { BaseTarget } from './base';
 import { BrewTarget } from './brew';
 import { CocoapodsTarget } from './cocoapods';
@@ -54,4 +55,10 @@ export function getTargetByName(
   targetName: string
 ): typeof BaseTarget | undefined {
   return TARGET_MAP[targetName];
+}
+
+export function getTargetId(target: TargetConfig): string {
+  return target.id
+    ? `${target.id}[${target.name}]`
+    : target.name || '__undefined__';
 }
--- a/src/targets/index.ts
+++ b/src/targets/index.ts
@@ -59,6 +59,6 @@ export function getTargetByName(
 
 export function getTargetId(target: TargetConfig): string {
   return target.id
-    ? `${target.id}[${target.name}]`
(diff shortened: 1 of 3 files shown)


BYK:
//@chadwhitacre

src/targets/index.ts
```diff
@@ -55,3 +56,9 @@ export function getTargetByName(
 ): typeof BaseTarget | undefined {
   return TARGET_MAP[targetName];
 }
+
+export function getTargetId(target: TargetConfig): string {
+  return target.id
+    ? `${target.id}[${target.name}]`
```
tonyo, line 62: For me personally it would be more intuitive to swap id and name, so we e.g. have `docker` as the generic target, and `docker[release]` or `docker#release` as more specific one.

tonyo (approved):

src/targets/index.ts
BYK: I agree and first I thought this would be a breaking change but it won't. I'll make that change and then merge.

Thanks for the feedback!

PR #142, opened 2020-12-01
fix(publish): Fix error when special target 'all' is used

### LINKED ISSUE #141: Bug: 0.13.3 rejects all targets and no longer publishes
Version 0.13.3 no longer publishes any package with the following message.

```
CRAFT_LOG=debug craft publish 8.0.0
ℹ info craft 0.13.3
ℹ info "craft" version is compatible with the minimal version from the configuration file.
ℹ info Publishing version: "8.0.0"
ℹ info Using "GithubStatusProvider" for status checks
ℹ info Using "GithubArtifactProvider" for artifacts
ℹ info No config provided for Github status provider, calculating the combined status...
ℹ info Revision 4b8e22baaa35414ad48ddc872bbd2a565a952c86 has been built successfully.
⚠ warn No valid targets detected! Exiting.
```

I believe the culprit to be in #137

BYK:
Fixes #141, follow up to #137 where there was a forgotten negation causing this issue.

Tests will follow in a separate PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 0.13.4
 
-- fix(publish): Fix error when special target 'all' is used
+- fix(publish): Fix error when special target 'all' is used (#142)
 
 ## 0.13.3
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.13.4
+
+- fix(publish): Fix error when special target 'all' is used
+
 ## 0.13.3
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.13.4
+
+- fix(publish): Fix error when special target 'all' is used
```
jan-auer, line 5: _nit_: Add PR number here.

BYK:
```suggestion
- fix(publish): Fix error when special target 'all' is used (#142)
```

PR #144, opened 2020-12-04
feat(publish): Add optional post-release script

BYK:
This PR adds an optional post-release script which is needed for
projects like [getsentry/sentry](https://git.io/JIs4n) and
[getsentry/onpremise](https://git.io/JIs4l). This would help keeping
the common publish/release repo and its downstream project independent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,7 +255,7 @@ preReleaseCommand: bash scripts/bump-version.sh
 This command will run after a successful `publish`. By default, it is set to
 `bash scripts/post-release.sh`. It will *not* error if the default script is
 missing though, as this may not be needed by all projects. Please refer to the
-[Post-release scipt conventions section](#post-release-script-conventions)
+[Post-release script conventions section](#post-release-script-conventions)
 for more details.
 
 ```yaml
@@ -907,7 +907,7 @@ Here is how you can integrate your GitHub project with `craft`:
 Among other actions, `craft prepare` runs an external, project-specific command
 or script that is responsible for version bumping. By default, this script
 should be located at: `./scripts/bump-version.sh`. The command can be configured
-by specifying `preReleaseCommand` configuration option in `craft.yml`.
+by specifying the `preReleaseCommand` configuration option in `craft.yml`.
 
 The following requirements are on the script interface and functionality:
 
--- a/README.md
+++ b/README.md
@@ -242,11 +242,24 @@ github:
 ### Pre-release Command
 
 This command will run on your newly created release branch as part of `prepare`
-command. By default, it is set to "bash scripts/bump-version.sh". Please refer
-to [this section](#pre-release-version-bumping-script-conventions) for more details.
+command. By default, it is set to `bash scripts/bump-version.sh`. Please refer
+to the [Pre-release version bumping script conventions section](#pre-release-version-bumping-script-conventions)
+for more details.
 
 ```yaml
 preReleaseCommand: bash scripts/bump-version.sh
+```
+
+### Post-release Command
+
+This command will run after a successful `publish`. By default, it is set to
+`bash scripts/post-release.sh`. It will *not* error if the default script is
+missing though, as this may not be needed by all projects. Please refer to the
+[Post-release scipt conventions section](#post-release-script-conventions)
+for more details.
+
+```yaml
+preReleaseCommand: bash scripts/post-release.sh
 ```
 
 ### Release Branch Name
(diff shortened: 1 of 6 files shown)


src/commands/publish.ts
```diff
@@ -396,6 +402,49 @@ async function handleReleaseBranch(
...
+ */
+export async function runPostReleaseCommand(
+  newVersion: string,
+  postReleaseCommand?: string
+): Promise<boolean> {
+  let sysCommand: string;
+  let args: string[];
+  if (postReleaseCommand !== undefined && postReleaseCommand.length === 0) {
+    // Not running post-release command
+    logger.warn('Not running the pre-release command: no command specified');
```
tonyo, line 422:
```suggestion
    logger.warn('Not running the post-release command: no command specified');
```

src/commands/publish.ts
```diff
@@ -396,6 +402,49 @@ async function handleReleaseBranch(
...
+    return false;
+  } else if (postReleaseCommand) {
+    [sysCommand, ...args] = shellQuote.parse(postReleaseCommand);
+  } else if (existsSync(DEFAULT_POST_RELEASE_SCRIPT_PATH)) {
+    sysCommand = '/bin/bash';
+    args = [DEFAULT_POST_RELEASE_SCRIPT_PATH];
+  } else {
+    // Not running post-release command
+    logger.warn(
```
tonyo, line 431: Would suggest to make these this message and the one above `info` (or lower) instead of `warn`, most of the people will not want to see a warning about a thing they don't really care about

tonyo (approved):

chadwhitacre (approved):
I like the feature. It makes sense to support the specific use case of bumping the version post-release, and in general as a parallel to the pre-release hook. That said, this PR has the speed/quality dial set pretty far on "speed." I'm approving in case "speed" is where we _want_ the dial set. If we want to turn the dial more towards quality and take on less tech debt, here are some suggestions:

1. Refactor to deduplicate code between `runPreReleaseCommand` and `runPostReleaseCommand`.
1. Test the post-release command hook in the test suite. There are tests for the pre-release hook. Either duplicate or refactor, based on decision under (1).
1. Update the documentation in the README. There's a section for the pre-release script; either duplicate or refactor.

BYK:
@chadwhitacre will address points 2 and 3 in your feedback before merging, thanks a lot! (I may take a stab at 1 but no promises).

jan-auer (approved):

BYK:
I need another review, especially for the README changes.

@chadwhitacre - I've opted not to refactor/DRY the common parts in script runners as the optional nature of the `post-release` script makes this effort mostly fruitless (adds more complexity).

README.md
```diff
@@ -242,13 +242,26 @@ github:
...
 ```yaml
 preReleaseCommand: bash scripts/bump-version.sh
 ```
 
+### Post-release Command
+
+This command will run after a successful `publish`. By default, it is set to
+`bash scripts/post-release.sh`. It will *not* error if the default script is
+missing though, as this may not be needed by all projects. Please refer to the
+[Post-release scipt conventions section](#post-release-script-conventions)
```
chadwhitacre, line 258:
```suggestion
[Post-release script conventions section](#post-release-script-conventions)
```

README.md
```diff
@@ -891,17 +904,15 @@ Here is how you can integrate your GitHub project with `craft`:
...
 ## Pre-release (Version-bumping) Script: Conventions
 
-Among other actions, `craft prepare` runs an external project-specific command
+Among other actions, `craft prepare` runs an external, project-specific command
 or script that is responsible for version bumping. By default, this script
-should be located at the following path: `scripts/bump-version.sh` (relative
-to the project root). The command can be configured by specifying
-`preReleaseCommand` configuration option in `craft.yml`.
+should be located at: `./scripts/bump-version.sh`. The command can be configured
+by specifying `preReleaseCommand` configuration option in `craft.yml`.
```
chadwhitacre, line 910:
```suggestion
by specifying the `preReleaseCommand` configuration option in `craft.yml`.
```

chadwhitacre (approved):
Couple doc nits, otherwise lgtm.

README.md
```diff
@@ -242,13 +242,26 @@ github:
...
+### Post-release Command
+
+This command will run after a successful `publish`. By default, it is set to
+`bash scripts/post-release.sh`. It will *not* error if the default script is
+missing though, as this may not be needed by all projects. Please refer to the
+[Post-release script conventions section](#post-release-script-conventions)
+for more details.
+
+```yaml
+preReleaseCommand: bash scripts/post-release.sh
```
tonyo, line 262: -> `postReleaseCommand`

README.md
```diff
@@ -922,6 +933,41 @@ export npm_config_git_tag_version=false
 npm version "${NEW_VERSION}"
 ```
 
+## Post-release Script: Conventions
+
+Among other actions, `craft publish` runs an external, project-specific command
+or script that can do things like bumping the development version. By default,
+this script should be located at: `./scripts/bump-version.sh`. Unlike the
```
tonyo, line 940: -> `./scripts/post-release.sh`

tonyo (commented):
A couple small things

README.md
BYK: #146 - sorry about this 😞

PR #129, opened 2020-10-13
ci(release): Add GHA-based release action

BYK:
#skip-changelog

PR #147, opened 2020-12-09
ci(release): We don't need a special things for checkout

BYK:
Follow up/fix to #129 as we separated the publish step now.


BYK:
Will merge w/o review as I need to use this now and it really doesn't need a review as it works: https://github.com/getsentry/craft/runs/1527338029?check_suite_focus=true

PR #149, opened 2020-12-11
ci(release): Add GH_RELEASE_PAT as we need a PAT to trigger builds

BYK:
This essentially reverts commit 8aab991e96c929285d9de5048d21efa74d211e0f.

Reason is GitHub preventing triggering more actions form inside another
action using the default `GITHUB_TOKEN`:

- https://stackoverflow.com/a/64078507/90297
- https://github.community/t/push-from-action-does-not-trigger-subsequent-action/16854

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,7 +11,7 @@ on:
 jobs:
   release:
     runs-on: ubuntu-latest
-    name: 'Cut a new release'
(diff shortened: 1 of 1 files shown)


BYK:
Test run: https://github.com/getsentry/craft/runs/1539359570?check_suite_focus=true

chadwhitacre (approved):

PR #156, opened 2021-01-04
Add tests for 'version'

djKooks:
Hello. I've add some test cases for 'versions'.
- Empty string
- Ignore values following after 'patch'

Please check 🙏


src/utils/__tests__/version.test.ts
```diff
@@ -22,6 +22,10 @@ describe('getVersion', () => {
   test('extracts a SemVer version from text', () => {
     expect(getVersion('1.0.0 (foobar)')).toBe('1.0.0');
   });
+
+  test('extracts a SemVer version after \'patch\'', () => {
```
tonyo, line 26:
```suggestion
  test('extracts a SemVer, but ignores subpatch level', () => {
```

tonyo (approved):
Thanks!

PR #157, opened 2021-01-05
upgrade(@google-cloud/storage): ^5.7.0

BYK:
This is to upgrade the vulnerable `axios` dependency. The latest version has the following breaking changes which are all okay:

- automatically detect contentType if not provided (googleapis/nodejs-storage#1190)
- drop keepAcl parameter in file copy (googleapis/nodejs-storage#1166)
- drop support for node.js 8.x
- allow leading slashes in file name (googleapis/nodejs-storage#820)
- upgrade engines field to >=8.10.0 (googleapis/nodejs-storage#688)

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "**/yargs-parser": ">=18.1.2"
   },
   "dependencies": {
-    "@google-cloud/storage": "2.3.4",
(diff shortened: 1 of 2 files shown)


jan-auer (approved):
Assuming that this is covered in tests 👍

PR #160, opened 2021-01-07
feat(aws-lambda): AWS Lambda layer target

iker-barriocanal:
Add a target to publish a layer and sets its permissions in each region for the JS SDK.


src/targets/awsLambda.ts
```diff
@@ -0,0 +1,139 @@
...
+  'ap-northeast-2',
+  'ca-central-1',
+  'eu-central-1',
+  'eu-west-1',
+  'eu-west-2',
+  'eu-west-3',
+  'sa-east-1',
+];
+
+const layerName = 'SentryNodeServerlessSdk';
```
scefali, line 38: @iker-barriocanal It would be great if we could make this some sort of environment variable or something so that when we test this locally, we can upload it to a different layer without modifying the code. But we can default to `SentryNodeServerlessSdk` if no env variable is set.

src/targets/awsLambda.ts
```diff
@@ -0,0 +1,139 @@
...
+          Content: {
+            ZipFile: artifactBuffer,
+          },
+          LayerName: layerName,
+          CompatibleRuntimes: compatibleRuntimes,
+          LicenseInfo: 'MIT',
+        })
+        .promise();
+
+      if (publishedLayer.Version === undefined) {
```
scefali, line 120: @iker-barriocanal when would the `Version` be undefined?

src/targets/awsLambda.ts
```diff
@@ -0,0 +1,139 @@
...
+  'ca-central-1',
+  'eu-central-1',
+  'eu-west-1',
+  'eu-west-2',
+  'eu-west-3',
+  'sa-east-1',
+];
+
+const layerName = 'SentryNodeServerlessSdk';
+const compatibleRuntimes = ['nodejs10.x', 'nodejs12.x'];
```
scefali, line 39: @iker-barriocanal When we do the layer for the Python runtimes, can we use the same script?

src/targets/awsLambda.ts
```diff
@@ -0,0 +1,139 @@
...
+      return undefined;
+    }
+
+    const artifactBuffer = fs.readFileSync(
+      await this.artifactProvider.downloadArtifact(packageFiles[0])
+    );
+
+    for (let i = 0; i < awsAllRegions.length; i++) {
+      const currentRegion = awsAllRegions[i];
+      const lambda = new Lambda({ region: currentRegion });
```
scefali, line 108: @iker-barriocanal I noticed this loop is pretty slow because we have to do each region sequentially. You can actually do them all in parallel using `Promise.all`:
```
const uploadOneRegion = (currentRegion: string) => {
   const lambda = new Lambda({ region: currentRegion });
   ...
}

await Promise.all(awsAllRegions.map(uploadOneRegion));
```

src/targets/awsLambda.ts
scefali: As discussed, we should throw an error if `Version` is `undefined`.

scefali (approved):

src/targets/__tests__/awsLambda.test.ts
```diff
@@ -0,0 +1,216 @@
...
+function setAwsEnvironmentVariables() {
+  process.env.AWS_ACCESS_KEY_ID = 'test aws access key';
+  process.env.AWS_SECRET_ACCESS_KEY = 'test aws secret access key';
+}
+
+describe('get layer name environment variable', () => {
+  test('default environment variable', () => {
+    if ('AWS_LAYER_NAME' in process.env) {
+      delete process.env.AWS_LAYER_NAME;
+    }
```
tonyo, line 52: Perhaps would be a bit safer to clear the variable in a `beforeEach` block, for the future.

src/targets/__tests__/awsLambda.test.ts
```diff
@@ -0,0 +1,216 @@
...
+      AwsLambdaTarget
+    );
+    awsTarget.artifactProvider.downloadArtifact = downloadArtifactMock.bind(
+      awsTarget
+    );
+    awsTarget.publishAwsLayer = publishAwsLayerMock.bind(AwsLambdaTarget);
+    awsTarget.addAwsLayerPermissions = addLayerPermissionsMock.bind(
+      AwsLambdaTarget
+    );
+    awsTarget.publish('', '');
```
tonyo, line 192: After running `publish` we should check if it actually called those mocked functions with right arguments. Also, `publish` is async, so you should await on it at some point.

src/targets/__tests__/awsLambda.test.ts
```diff
@@ -0,0 +1,216 @@
...
+    // thrown; when it's on dry run, the error is logged and `undefined` is
+    // returned. Thus, both alternatives have been considered.
+    try {
+      const tooManyPackagesFound = await awsTarget.publish(
+        'version',
+        'revision'
+      );
+      expect(tooManyPackagesFound).toBe(undefined);
+    } catch (error) {
+      expect(error instanceof Error).toBe(true);
```
tonyo, line 145: This test looks suspiciously similar to the previous one. Wdyt about actually checking the error message to confirm that it fails for the right reason?

tonyo (changes requested):
Mostly minor comments.

Separate 👍 👍 for adding tests for the target!

Please also add the documentation for this new target to README.

src/targets/awsLambda.ts
```diff
@@ -0,0 +1,187 @@
...
+    }
+    return {
+      awsAccessKeyId: process.env.AWS_ACCESS_KEY_ID,
+      awsSecretAccessKey: process.env.AWS_SECRET_ACCESS_KEY,
+    };
+  }
+
+  /**
+   * Publishes current lambda layer zip bundle to AWS Lambda.
+   * @param _version New version to be released.
```
rhcarvalho, line 104: Why the underscore?

src/targets/awsLambda.ts
```diff
@@ -0,0 +1,187 @@
...
+   * @param revision Git commit SHA to be published.
+   */
+  public async publish(_version: string, revision: string): Promise<any> {
+    logger.debug('Fetching artifact list...');
+    const packageFiles = await this.getArtifactsForRevision(revision, {
+      includeNames: DEFAULT_AWS_LAMBDA_DIST_REGEX,
+    });
+
+    if (packageFiles.length == 0) {
+      reportError('Cannot release to AWS Lambda: no packages found');
```
rhcarvalho, line 114:
```suggestion
      reportError('Cannot publish AWS Lambda Layer: no packages found');
```

src/targets/awsLambda.ts
```diff
@@ -0,0 +1,187 @@
...
+      return undefined;
+    } else if (packageFiles.length > 1) {
+      reportError(`Cannot release to AWS Lambda:
```
rhcarvalho, line 117:
```suggestion
      reportError(`Cannot publish AWS Lambda Layer:
```

src/targets/awsLambda.ts
```diff
@@ -0,0 +1,187 @@
...
+        Content: {
+          ZipFile: artifactBuffer,
+        },
+        LayerName: AwsLambdaTarget.layerName,
+        CompatibleRuntimes: compatibleRuntimes,
+        LicenseInfo: 'MIT',
+      });
+
+      if (publishedLayer.Version === undefined) {
+        reportError(`Error while publishing AWS Layer to ${currentRegion}`);
```
rhcarvalho, line 139: s/AWS Layer/AWS Lambda Layer/

This error message is missing information to be useful, right now even reading the code leaves one puzzled regarding what happened.
Why do we check the `publishedLayer.Version`? Is it a proxy for the Lambda Client request worked fine? What should craft users (e.g. someone releasing sentry-javascript) do in case this error pops up? Re-run? Config error? Unknown?

src/targets/awsLambda.ts
```diff
@@ -0,0 +1,187 @@
...
+    const publishRegionToLambda = async (currentRegion: string) => {
+      const lambda = new Lambda({ region: currentRegion });
+
+      const publishedLayer = await this.publishAwsLayer(lambda, {
+        Content: {
+          ZipFile: artifactBuffer,
+        },
+        LayerName: AwsLambdaTarget.layerName,
+        CompatibleRuntimes: compatibleRuntimes,
+        LicenseInfo: 'MIT',
```
rhcarvalho, line 135: This should be project-specific config, not hard-coded in Craft.

src/targets/awsLambda.ts
```diff
@@ -0,0 +1,187 @@
...
+      reportError(`Cannot release to AWS Lambda:
+      multiple packages with matching patterns were found.`);
+      return undefined;
+    }
+
+    const artifactBuffer = fs.readFileSync(
+      await this.artifactProvider.downloadArtifact(packageFiles[0])
+    );
+
+    const publishRegionToLambda = async (currentRegion: string) => {
```
rhcarvalho, line 126: nit:

```suggestion
    const publishRegion = async (currentRegion: string) => {
```

For consistency: `publish` is the general method, `publishRegion` for a specific region.

src/targets/awsLambda.ts
```diff
@@ -0,0 +1,187 @@
...
+    lambda: Lambda,
+    layerPermissionData: Lambda.AddLayerVersionPermissionRequest
+  ): Promise<
+    PromiseResult<Lambda.AddLayerVersionPermissionResponse, AWSError>
+  > {
+    return await lambda
+      .addLayerVersionPermission(layerPermissionData)
+      .promise();
+  }
+}
```
rhcarvalho, line 187: Why wrap these one-liners into methods? Wouldn't it be the same to mock the Lambda client in tests and write less code?

src/targets/awsLambda.ts
```diff
@@ -0,0 +1,187 @@
...
+  awsAccessKeyId: string;
+  /** AWS secret access key, set as `AWS_SECRET_ACCESS_KEY`. */
+  awsSecretAccessKey: string;
+}
+
+/**
+ * The default layer name is used when no `AWS_LAYER_NAME`
+ * environment variable is found.
+ */
+export const defaultLayerName = 'SentryNodeServerlessSDK';
```
rhcarvalho, line 52: This should be a required project-specific config, and not read from the environment.
Different projects, different layer names, no default (fail if not specified).

src/targets/__tests__/awsLambda.test.ts
rhcarvalho: Good catch, the layer name should come in as project-specific config, not environment variable. These lines should go :)

rhcarvalho (commented):
Good stuff @iker-barriocanal 

Dropped some comments to help move this to completion. Let's chat if I can help.

High level notes:

- Need to split clearly what is part of project config and what is part of the Craft implementation.
- Missing some documentation on how to use this new target in a craft-powered project (like sentry-javascript and sentry-python), essentially documenting what is the required config.

src/targets/awsLambda.ts
iker-barriocanal: It's an unused variable. AWS Lambda doesn't care about that version since it has it's own version counter for each layer ([available in the ARN](https://docs.aws.amazon.com/lambda/latest/dg/configuration-versions.html#versioning-versions-using)); where new versions are calculated on an auto-increment basis, starting from 1.

src/targets/awsLambda.ts
```diff
@@ -0,0 +1,187 @@
...
+  'ap-southeast-2',
+  'ap-northeast-1',
+  'ap-northeast-2',
+  'ca-central-1',
+  'eu-central-1',
+  'eu-west-1',
+  'eu-west-2',
+  'eu-west-3',
+  'sa-east-1',
+];
```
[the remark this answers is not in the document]
scefali: @rhcarvalho We can use the `describe-regions` command https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-regions.html

src/targets/awsLambda.ts
rhcarvalho: Aha, I didn't notice it was unused... then why is it there? The examples in tests call either `.publish('version', 'revision')` or `.publish('', '')`, can't this be simply `publish(revision: string)`?

iker-barriocanal: The [method](https://github.com/getsentry/craft/blob/7bda0be6556047f7c39afa2107b5b0e2c368b2e0/src/targets/base.ts#L49) is overridden from the base class `BaseTarget`, which contains both parameters.

README.md
```diff
@@ -889,6 +889,38 @@ targets:
   - name: gem
 ```
 
+### AWS Lambda Layer
```
tonyo, line 892:
```suggestion
### AWS Lambda Layer (`aws-lambda-layer`)
```

README.md
```diff
@@ -889,6 +889,38 @@ targets:
...
+
+The target will create a new public lambda layer in each available region with
+the extracted artifact from GitHub actions.
```
tonyo, line 895: Let's not mention GitHub actions specifically, it's just one of potential artifact providers.

README.md
```diff
@@ -889,6 +889,38 @@ targets:
...
+**Environment**
+| Name | Description |
+| --------------------- | -------------------------------------------------------------------------- |
+| AWS_ACCESS_KEY | The access key of the AWS account to create and publish the layers. |
+| AWS_SECRET_ACCESS_KEY | The secret access key of the AWS account to create and publish the layers. |
+
+**Configuration**
+| Option | Description |
+| -------------- | ------------------------------------------------------------------------ |
+| includeNames | **optional** Pattern (between `/`) of the ZIP artifact to be used as the content of the layer. If after filtering the artifacts with this option there's no exactly one artifact, it'll fail. |
```
tonyo, line 906: `includeNames` exists for all targets, so perhaps we can link to https://github.com/getsentry/craft#per-target-options and mention the default value here?

README.md
```diff
@@ -889,6 +889,38 @@ targets:
...
+### AWS Lambda Layer
+
+The target will create a new public lambda layer in each available region with
+the extracted artifact from GitHub actions.
+
+**Environment**
+| Name | Description |
+| --------------------- | -------------------------------------------------------------------------- |
+| AWS_ACCESS_KEY | The access key of the AWS account to create and publish the layers. |
+| AWS_SECRET_ACCESS_KEY | The secret access key of the AWS account to create and publish the layers. |
```
rhcarvalho, line 901: nit: You cleaned up the tables above, maybe you can do the same with the new ones?

src/artifact_providers/base.ts
```diff
@@ -199,7 +199,7 @@ export abstract class BaseArtifactProvider {
    * @param downloadDirectory Directory where downloaded artifact is stored
    * @returns Absolute path to the saved file
    */
-  protected abstract async doDownloadArtifact(
+  protected abstract doDownloadArtifact(
```
tonyo, line 202: Could you clarify why `async` was removed?

README.md
```diff
@@ -889,6 +889,38 @@ targets:
...
+
+```yaml
+targets:
+  - name: aws-lambda-layer
+    includeNames: /^sentry-node-serverless-\d+(\.\d+)*\.zip$/
+    layerName: SentryNodeServerlessSDK
+    compatibleRuntimes:
+      - nodejs10.x
+      - nodejs12.x
+    license: MIT
```
rhcarvalho, line 921: 👍

src/targets/awsLambdaLayer.ts
```diff
@@ -0,0 +1,202 @@
...
+/**
+ * Target responsible for uploading files to AWS Lambda.
+ */
+export class AwsLambdaLayerTarget extends BaseTarget {
+  /** Target name */
+  public readonly name: string = 'aws-lambda-layer';
+  /** Target options */
+  public readonly awsLambdaConfig: AwsLambdaTargetOptions;
+  /** Name of the layer to be published */
+  static layerName: string;
```
rhcarvalho, line 34: Why is this `static`? Hmm, looks like it is never used?

package.json
```diff
@@ -20,6 +20,9 @@
     "**/yargs-parser": ">=18.1.2"
   },
   "dependencies": {
+    "@aws-sdk/client-ec2": "^3.2.0",
+    "@aws-sdk/client-lambda": "^3.2.0",
+    "@aws-sdk/types": "^3.1.0",
```
tonyo, line 25: This should probably go to `devDependencies`

src/targets/__tests__/awsLambda.test.ts
```diff
@@ -0,0 +1,261 @@
...
+      awsTarget.publishAwsLayer = publishAwsLayerMockUndefinedVersion.bind(
+        AwsLambdaLayerTarget
+      );
+      const publishedLayerVersion = await awsTarget.publish('', '');
+      // `publish` should report an error. When it's not dry run, the error is
+      // thrown; when it's on dry run, the error is logged and `undefined` is
+      // returned. Thus, both alternatives have been considered.
+      expect(publishedLayerVersion).toBe(undefined);
+    } catch (error) {
+      expect(error instanceof Error).toBe(true);
```
tonyo, line 253: To confirm: wouldn't this always be true?

If you want to confirm that a function throws an error, please consider using https://jestjs.io/docs/en/expect.html#tothrowerror

src/targets/awsLambdaLayer.ts
```diff
@@ -0,0 +1,202 @@
...
+    };
+  }
+
+  /**
+   * Requests all regions that are enabled for the current account (or all
+   * regions) to AWS. For more information, see
+   * https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/EC2.html#describeRegions-property
+   */
+  public async getAwsRegions(): Promise<DescribeRegionsCommandOutput> {
+    const ec2 = new EC2({ region: 'us-east-2' });
```
rhcarvalho, line 68: Could use `AWS_DEFAULT_REGION` like aws-cli; can be left for a future improvement.

src/targets/awsLambdaLayer.ts
```diff
@@ -0,0 +1,202 @@
...
+ */
+export class AwsLambdaLayerTarget extends BaseTarget {
+  /** Target name */
+  public readonly name: string = 'aws-lambda-layer';
+  /** Target options */
+  public readonly awsLambdaConfig: AwsLambdaTargetOptions;
+  /** Name of the layer to be published */
+  static layerName: string;
+  /** All AWS regions available for the current user */
+  public awsRegions: any;
```
rhcarvalho, line 36:
```suggestion
  public awsRegions: string[];
```

?

rhcarvalho: It looks like this could be simply a local variable in `publish` though, seems not to be used anywhere else.

rhcarvalho (approved):
Looks good to me, I think we can go ahead and update sentry-python and sentry-javascript to use this and make any adjustments later if need be. Good job @iker-barriocanal!

rhcarvalho:
Re-requesting Anton if he still has any concerns from the points raised earlier.

tonyo:
@rhcarvalho Let's give Iker a break 😅  I've already left one a few minutes ago.

rhcarvalho:
Oh sorry, I did not see it before I posted.

src/artifact_providers/base.ts
iker-barriocanal: It's an abstract method, so it shouldn't force the methods overriding it to be `async` since there are other ways to to return a promise. Also, to make Typescript happy (error TS1243).

src/targets/__tests__/awsLambda.test.ts
iker-barriocanal: The code inside `catch`, yes. `publish` will only throw if it's not running in dry-run mode; if it is, it doesn't throw anything and returns `undefined`. So, the approach is to run `publish` and expect to get `undefined`; if it throws an error before that, catch it. The method `toThrow`, afaik, requires to make the function call inside the `expect`, passing the test when the dry-run mode is off but failing when it's on.

package.json
tonyo: (I mean `@aws-sdk/types`, assuming it's just TypeScript types).

tonyo (approved):

Issue #167, opened 2021-01-19
Registry target, create files if not exist yet


HazAT:
When introducing a new package, `craft` should try to create the required folder/file structure so this doesn't have to be done manually.

```
[12:23:18 PM] ======================================
[12:23:18 PM] === Publishing to target: registry ===
[12:23:18 PM] ======================================
[12:23:18 PM] [[REGISTRY]] Cloning "https://github.com/getsentry/sentry-release-registry/" to "/tmp/craft-release-registry-DFIZaD"...
[12:23:19 PM] [[REGISTRY]] Adding the version file to the registry for canonical name "npm:@sentry/wasm"...
[12:23:19 PM] Error: ENOENT: no such file or directory, open '/tmp/craft-release-registry-DFIZaD/packages/npm/@sentry/wasm/latest.json'
at Object.openSync (fs.js:462:3)
at Object.readFileSync (fs.js:364:35)
at RegistryTarget.addVersionToRegistry (/craft/dist/targets/registry.js:335:47)
at RegistryTarget.pushVersionToRegistry (/craft/dist/targets/registry.js:361:20)
at runMicrotasks (<anonymous>)
at processTicksAndRejections (internal/process/task_queues.js:97:5)
at async Object.withTempDir (/craft/dist/utils/files.js:76:16)
at async RegistryTarget.publish (/craft/dist/targets/registry.js:398:9)
at async /craft/dist/commands/publish.js:135:13
at async Object.withTempDir (/craft/dist/utils/files.js:76:16)
at async publishToTargets (/craft/dist/commands/publish.js:101:5)
at async publishMain (/craft/dist/commands/publish.js:422:9)
at async Object.handler (/craft/dist/commands/publish.js:449:16)
```

tonyo:
Please note, that the current behavior is there intentionally.
It helps to protect against invalid canonical names; also, according to the current logic, we take `latest.json` (that already has links, SDK name, etc.) and just put there the new version and file infos.

If everyone is fine with letting the registry target to create new folders/packages, that behavior ^ has to change as well, and we'll have to probably keep all that meta information in the registry in `.craft.yaml` now.

BYK:
If we had an early warning system, I think having the current behavior in place might have been useful. Currently, it just made things worse as far as I can tell.

kamilogorek:
We agreed internally that it should stay as is for now.

PR #168, opened 2021-01-20
ref: Decouple functionality from the registry

iker-barriocanal:
This PR is not adding any functionality, but refactoring the `registry` target to decouple some functionality, so that there are more pieces with smaller responsibility each.

Currently the target `aws-lambda-layer`, after creating new layers, only logs to console the layer's ARNs; but we want to have the data of the created new layers in the registry. However, only the `registry` target is able to write there. In order to have the targets independent between them, the goal is to make the `aws-lambda-layer` target able to write to the registry, so that it doesn’t need to interact with the `registry` target. This PR is focused on the refactoring part of that small feature. Besides `aws-lambda-layer`, additional serverless targets might be created in the future.

Aiming to leave the codebase a bit better and make the code more readable and usable, additional refactoring has been made not to have the `registry` target do that many things. Thus, some functionality the `registry` target does has been broken down into other modules (again, some of the modules won't be used in the new feature of the `aws-lambda-layer` target, and will only be used by the `registry` target for now).

This refactoring of decoupling functionality from the registry target (`src/targets/registry`) includes:
- Checksums. Casting and getting the artifact checksums are available in the `src/utils/checksums` module.
- Symlinks. Creating symlinks is now achievable in the `src/utils/symlink` module.
- Getting package paths. Getting the path of a package given its type and the canonical name is now implemented in the `src/utils/packagePath` module.
- Canonical parsing. Parsing the canonical name of a package to a list of directories is now also implemented in the `src/utils/packagePath` module.
- Getting the package manifest and updating a manifest's symlinks are now two functions acting as `utils` for the `registry` target, in the `src/utils/registry` directory.


HazAT:
I didn't look at the code yet but just FYI it would be nice if we can maybe also fix that.
https://github.com/getsentry/craft/issues/167

tonyo:
@HazAT Please check the comment here https://github.com/getsentry/craft/issues/167#issuecomment-763537152
I believe that issue deserves a separate discussion/fix.

src/utils/gitTasks.ts
```diff
@@ -0,0 +1,58 @@
...
+ * @param repoPath The repo to be cloned.
+ * @param directory The directory to be cloned the repo in.
+ */
+export async function gitClone(
+  repoPath: string,
+  directory: string
+): Promise<void> {
+  await simpleGit()
+    .silent(true)
+    .clone(repoPath, directory);
```
tonyo, line 57: We should probably be more consistent, and also return `git` here, similar to `gitCheckout`.

src/utils/gitTasks.ts
```diff
@@ -0,0 +1,58 @@
...
+ * @param directory The directory to perform the git action in.
+ * @param branch The tag or revision to be checked out.
+ *  Additional arguments can be provided by including an array of strings.
+ */
+export async function gitCheckout(
+  directory: string,
+  what: string | string[] = 'master'
+): Promise<simpleGit.SimpleGit> {
+  const git = simpleGit(directory).silent(true);
+  git.checkout(what);
```
tonyo, line 42: Do we need `await` here?

src/utils/packagePath.ts
```diff
@@ -0,0 +1,55 @@
+import * as path from 'path';
+import { RegistryPackageType } from '../targets/registry';
+import { parseCanonical } from './canonical';
```
tonyo, line 3: Let's merge this file with `canonical` module for simplicity (unless there's a reason to keep them separate).

src/utils/packagePath.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+/**
+ * Returns the path to the package from its canonical name.
+ *
+ * @param registryDir The path to the local registry.
+ * @param canonical The app's canonical name.
+ */
+export function getPackageDirPath(
+  registryConfType: string,
```
tonyo, line 42: Looking at how it's used in the function, wdyt about calling it `packageType`? It can also be typed as `RegistryPackageType`

Let's also add a doc string for it.

tonyo (changes requested):
Some of the things that were extracted seem too tied to the registry target still, which makes them less reusable or general-purpose (for example, functions `addPackageVersionToRegistry` and `createSymlinks`). I don't have good suggestions how to decouple them better, but let's at least move all these new modules to the `registry` submodule (so it'll be `utils/registry`), to illustrate that they're not really useful outside interacting with the release registry.

For functions that are more generic (e.g., git operations), please check if we already use similar snippets in other files, and whether it makes sense to start using your new util functions or not.

BYK:
@iker-barriocanal it would also be great to tie this to some existing issues or planned future work to ensure this is not refactoring for refactoring sake. This seems like a pretty large change and I'd love to have a very good reason behind it before taking the risk.

iker-barriocanal:
@BYK at the moment, updates on the registry are done through the registry target. However, when new AWS Lambda layers are created, their information also needs to be added (see https://github.com/getsentry/sentry-release-registry/pull/30). It's easier to avoid code duplication if some functionality is decoupled.

iker-barriocanal:
@tonyo some of the things can (and will) be more easily be used by other modules (e.g. `symlink`), even if they are only used by the registry atm.

Other modules (e.g. `utils/checksum`, `utils/packagePath` and `utils/registry`) are also only used by the registry now, and in fact not planning to use them shortly; but I also wanted to decouple responsibilities and not have that many things done in the same file, since I feel it's easier to read and in general to deal with. Open to bring these back, though.

tonyo (approved):

src/utils/packagePath.ts
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Returns the path to the SDK, given its canonical name.
+ *
+ * @param registryDir The path to the local registry.
+ * @param canonical The SDK's canonical name.
+ * @returns The SDK path.
+ */
+function getSdkPackagePath(registryDir: string, canonical: string): string {
+  const packageDirs = parseCanonical(canonical);
+  return [registryDir, 'packages'].concat(packageDirs).join(path.sep);
```
BYK, line 14:
```suggestion
  return path.posix.join(registryDir, 'packages', ...packageDirs);
```

src/utils/packagePath.ts
```diff
@@ -0,0 +1,79 @@
...
+ * @returns The app path.
+ */
+function getAppPackagePath(registryDir: string, canonical: string): string {
+  const packageDirs = parseCanonical(canonical);
+  if (packageDirs[0] !== 'app') {
+    throw new ConfigurationError(
+      `Invalid canonical entry for an app: ${canonical}`
+    );
+  }
+  return [registryDir, 'apps'].concat(packageDirs.slice(1)).join(path.sep);
```
BYK, line 31:
```suggestion
  return path.posix.join(registryDir, 'apps', ...packageDirs.slice(1));
```

src/utils/packagePath.ts
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Parses registry canonical name to a list of registry directories
+ *
+ * Example: "npm:@sentry/browser" -> ["npm", "@sentry", "browser"]
+ *
+ * @param canonicalName Registry canonical name
+ * @returns A list of directories
+ */
+export function parseCanonical(canonicalName: string): string[] {
```
BYK, line 65: May want to replace this with https://github.com/npm/npm-package-arg

src/utils/checksum.ts
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * TODO: this all has to be replaced with JSON schema.
+ *
+ * @param checksums Raw checksum configuration.
+ */
+export function castChecksums(checksums: any[]): ChecksumEntry[] {
+  if (!checksums) {
+    return [];
+  }
+  if (!Array.isArray(checksums)) {
```
BYK, line 27: Shouldn't this be checked via TypeScript?

src/utils/checksum.ts
```diff
@@ -0,0 +1,80 @@
...
+  return checksums.map(
+    (item: any): ChecksumEntry => {
+      if (typeof item !== 'object' || !item.algorithm || !item.format) {
+        throw new ConfigurationError(
+          `Invalid checksum type: ${JSON.stringify(item)}`
+        );
+      }
+      if (
+        !Object.values(HashAlgorithm).includes(item.algorithm) ||
+        !Object.values(HashOutputFormat).includes(item.format)
```
BYK, line 41: I would move these out of the loop:

```js
const VALID_ALGORITHMS = new Set(Object.values(HashAlgorithm));
const VALID_OUTPUT_FORMATS = new Set(Object.values(HashOutputFormat));
```

BYK (approved):
Added some nits but LGTM and can be merged as is.

PR #171, opened 2021-01-27
fix(remote): Clone with auth

jan-auer:
Fixes a regression from #168.

cc @iker-barriocanal


BYK (approved):

jan-auer:
Disclaimer: I did not look at tests, neither verify that it actually works. This is just spotted from code review and should resolve the following failure:

```
Error: fatal: could not read Username for 'github.com': No such device or address
```

PR #172, opened 2021-02-01
feat: Update the sentry release registry with AWS Lambda layer versions

iker-barriocanal:
# Summary
The latest versions of the AWS Lambda layers will be needed in our docs (manual setup) as well as in Sentry (automated setup). This PR updates the release registry with the information of the latest updated layers, by generating files that describe them during the release process.

# Registry updates
The `aws-lambda-layer` target will have a change in `.craft.yml`, where `compatibleRuntimes` will be a list of objects containing a name and a list of versions (e.g. `name: node` and `versions: ['nodejs10.x', 'nodejs12.x']`). The new files will be placed in the directory `aws-lambda-layers/{runtime name}/`, and will be named `{version to release}.json`.

Each version file contains both dynamic and static attributes. Dynamic attributes are extracted when layers are created:

- `canonical`. The canonical name of the layer, following the pattern `aws-layer:{runtime name}`.
- `sdk_version`. The version to be released.
- `account_number`. The account number of the account creating the layers. This can be found in the ARN of the created layer (`arn:aws:lambda:{region}:{account number}:layer:{layer name}:{layer version}`).
- `layer_name`. The name of the layer. This value is taken from the `.craft.yml` project config file.
- `regions`. A list of mappings of regions to versions, since different regions may have different versions.

Static attributes are common attributes to all the versions and won't usually change. They are located in the `base.json` file, placed in each runtime directory. All the attributes in this file will be added to the new version files. `name`, `repo_url` and `main_docs_url` are the only static attributes so far.

It's important that the directories already exist in the registry, since no new directories are created (already existing for `node` https://github.com/getsentry/sentry-release-registry/pull/35). If the directory structure doesn't exist, a warning message will appear; and ARNs will always be logged. Besides, symlinks targeting to the major, minor and latest version will be created. `latest.json` will always link to the latest created version.

# PR changes
The main changes in this PR are the new `src/utils/awsLambdaLayerManager`, a file containing the interaction of Craft and AWS. All the functionality of this file was previously on `src/targets/awsLambdaLayer`, but a couple of util functions. This abstraction has been made to make the code easier to deal with.

The `src/targets/awsLambdaLayer`, where the core functionality of the `aws-lambda-layer` target is, has now two extra features. On the one hand, the relevant information is collected (static and dynamic attributes), written to a file, and symlinks updated. On the other hand, these new files are created in a cloned repo, and then committed and pushed to the remote.

Lastly, the README has been updated, and tests and mocks have been added too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.16.2
+
+- feat(aws-lambda): Update the sentry release registry with AWS Lambda layer versions (#172)
+
 ## 0.16.1
 
 - fix(gcs-target): Fix incorrect upload path to GCS when it has a leading slash (#170)
--- a/src/targets/awsLambdaLayer.ts
+++ b/src/targets/awsLambdaLayer.ts
@@ -139,117 +139,8 @@ export class AwsLambdaLayerTarget extends BaseTarget {
       await this.artifactProvider.downloadArtifact(packageFiles[0])
     );
 
-    /** Creates symlinks to the new version file, and updates previous ones if needed. */
-    const createVersionSymlinks = (
-      directory: string,
-      versionFilepath: string
-    ): void => {
-      const latestVersionPath = path.posix.join(directory, 'latest.json');
-      if (fs.existsSync(latestVersionPath)) {
-        const previousVersion = fs
-          .readlinkSync(latestVersionPath)
-          .split('.json')[0];
-        createSymlinks(versionFilepath, version, previousVersion);
-      } else {
-        // When no previous versions are found, just create symlinks.
-        createSymlinks(versionFilepath, version);
-      }
-    };
-
     logger.debug('Fetching AWS regions...');
     const awsRegions = extractRegionNames(await getRegionsFromAws());
-
-    /** Publishes new AWS Lambda layers for every runtime. */
-    const publishRuntimes = async (directory: string): Promise<void> => {
-      await Promise.all(
-        this.config.compatibleRuntimes.map(
-          async (runtime: CompatibleRuntime) => {
-            const layerManager = new AwsLambdaLayerManager(
-              runtime,
-              this.config.layerName,
-              this.config.license,
-              artifactBuffer,
-              awsRegions
-            );
-
-            let publishedLayers = [];
-            try {
-              publishedLayers = await layerManager.publishToAllRegions();
-            } catch (error) {
-              logger.error(
-                `Did not publish layers for ${runtime.name}. ` +
-                  `Something went wrong with AWS: ${error.message}`
-              );
-              return;
-            }
-
-            // If no layers have been created, don't do extra work updating files.
-            if (publishedLayers.length == 0) {
-              logger.info(`${runtime.name}: no layers published.`);
-              return;
-            } else {
-              logger.info(
-                `${runtime.name}: ${publishedLayers.length} layers published.`
-              );
-            }
-
-            // Base directory for the layer files of the current runtime.
-            const runtimeBaseDir = path.posix.join(
-              directory,
-              this.AWS_REGISTRY_DIR,
-              runtime.name
-            );
-            if (!fs.existsSync(runtimeBaseDir)) {
-              logger.warn(
-                `Directory structure for ${runtime.name} is missing, skipping file creation.`
-              );
-              return;
-            }
-
-            const regionsVersions = publishedLayers.map(layer => {
-              return {
-                region: layer.region,
-                version: layer.version.toString(),
-              };
-            });
-
-            // Common data specific to all the layers in the current runtime.
-            const runtimeData = {
-              canonical: layerManager.getCanonicalName(),
-              sdk_version: version,
-              account_number: getAccountFromArn(publishedLayers[0].arn),
-              layer_name: this.config.layerName,
-              regions: regionsVersions,
-            };
-
-            const baseFilepath = path.posix.join(
-              runtimeBaseDir,
-              this.BASE_FILENAME
-            );
-            const newVersionFilepath = path.posix.join(
-              runtimeBaseDir,
-              `${version}.json`
-            );
-
-            if (!fs.existsSync(baseFilepath)) {
-              logger.warn(`The ${runtime.name} base file is missing.`);
-              fs.writeFileSync(newVersionFilepath, JSON.stringify(runtimeData));
-            } else {
-              const baseData = JSON.parse(
-                fs.readFileSync(baseFilepath).toString()
-              );
-              fs.writeFileSync(
-                newVersionFilepath,
-                JSON.stringify({ ...baseData, ...runtimeData })
-              );
-            }
-
-            createVersionSymlinks(runtimeBaseDir, newVersionFilepath);
-            logger.info(`${runtime.name}: created files and updated symlinks.`);
-          }
-        )
-      );
-    };
 
     const remote = this.awsLambdaConfig.registryRemote;
     const username = await getAuthUsername(this.github);
@@ -261,7 +152,12 @@ export class AwsLambdaLayerTarget extends BaseTarget {
         logger.info(`Cloning ${remote.getRemoteString()} to ${directory}...`);
         await git.clone(remote.getRemoteStringWithAuth(), directory);
 
-        await publishRuntimes(directory);
+        await this.publishRuntimes(
+          version,
+          directory,
+          awsRegions,
+          artifactBuffer
+        );
 
         await git.add(['.']);
         await git.checkout('master');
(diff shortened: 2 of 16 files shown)


src/utils/awsLambdaLayerManager.ts
```diff
@@ -0,0 +1,162 @@
...
+  awsRegions.Regions?.map(currentRegion => {
+    if (currentRegion.RegionName !== undefined) {
+      regionNames.push(currentRegion.RegionName);
+    }
+  });
+  return regionNames;
+}
+
+/** Substring used to separate the different ARN parts. */
+const ARN_SEPARATOR = ':';
```
tonyo, line 151: Let's move these constants to the top of the file.

src/utils/__tests__/awsLambdaLayerManager.test.ts
```diff
@@ -0,0 +1,90 @@
...
+  );
+}
+
+describe('canonical', () => {
+  test('get canonical name', () => {
+    const manager = getTestAwsLambdaLayerManager();
+    const canonicalSuffix = manager
+      .getCanonicalName()
+      .split(CANONICAL_SEPARATOR)[1];
+    expect(canonicalSuffix).toBe(COMPATIBLE_RUNTIME_DATA.name);
```
tonyo, line 31: nit/personal opinion: replacing `COMPATIBLE_RUNTIME_DATA.name` with the actual value (`test runtime`) would make the test a bit more robust because it'd avoid failing in unfortunate situations when e.g. `name` attribute is changed and also `getCanonicalName` changes the logic, in that situation the test has a chance of passing.

src/targets/awsLambdaLayer.ts
```diff
@@ -1,25 +1,47 @@
...
+  getAccountFromArn,
+  getRegionsFromAws,
+} from '../utils/awsLambdaLayerManager';
+import { createSymlinks } from '../utils/symlink';
+import { withTempDir } from '../utils/files';
+import { isDryRun } from '../utils/helpers';
 
 const logger = loggerRaw.withScope(`[aws-lambda-layer]`);
 
+const DEFAULT_REGISTRY_REMOTE: GithubRemote = new GithubRemote(
```
tonyo, line 32: This now duplicates the value in `src/targets/registry`, wdyt about extracting it as a constant/function somewhere?

src/targets/awsLambdaLayer.ts
```diff
@@ -143,56 +143,145 @@ export class AwsLambdaLayerTarget extends BaseTarget {
...
-        VersionNumber: publishedLayer.Version,
-        StatementId: 'public',
-        Action: 'lambda:GetLayerVersion',
-        Principal: '*',
-      });
-
-      logger.info(`Published layer in ${currentRegion}:
-        ${publishedLayer.LayerVersionArn}`);
+    /** Creates symlinks to the new version file, and updates previous ones if needed. */
+    const createVersionSymlinks = (
```
tonyo, line 147: Could you explain why this (and another one below) is an anon function assigned to const and not a proper method?

src/targets/awsLambdaLayer.ts
```diff
@@ -29,20 +51,30 @@ export class AwsLambdaLayerTarget extends BaseTarget {
   /** Target name */
   public readonly name: string = 'aws-lambda-layer';
   /** Target options */
-  public readonly awsLambdaConfig: AwsLambdaTargetOptions;
+  public readonly awsLambdaConfig: AwsLambdaTargetConfig;
+  /** GitHub client. */
+  public readonly github: Github;
+  /** Github repo configuration. */
+  public readonly githubRepo: GithubGlobalConfig;
```
tonyo, line 58: Do we need `githubRepo` in this class?

src/targets/awsLambdaLayer.ts
```diff
@@ -143,56 +143,145 @@ export class AwsLambdaLayerTarget extends BaseTarget {
...
+              logger.info(`${runtime.name}: no layers published.`);
+              return;
+            } else {
+              logger.info(
+                `${runtime.name}: ${publishedLayers.length} layers published.`
+              );
+            }
+
+            // Base directory for the layer files of the current runtime.
+            const RUNTIME_BASE_DIR = path.posix.join(
```
tonyo, line 201:
```suggestion
            const runtimeBaseDir = path.posix.join(
```
Let's reserve uppercase identifiers for global/class constants

tonyo (changes requested):
Left a few minor comments.

General thoughts:
* The registry update now runs unconditionally, wdyt about making it configurable on a target level? Or, do we e.g. plan to release beta layer versions that don't need to be pushed to the registry?
* I'm a bit salty that we now have to blur borders of target responsibilities (basically, running the registry target functions from the aws-lambda target). Not your fault @iker-barriocanal, but more like a reminder that we might want to reconsider the overall design of how targets work together and whether we need, for example, some form of state (local or remote) when publishing.

src/targets/awsLambdaLayer.ts
iker-barriocanal: Extracted to `getRegistryGithubRemote` in `src/utils/githubApi`.

src/utils/githubApi.ts
```diff
@@ -107,6 +107,13 @@ export class GithubRemote {
   }
 }
 
+/**
+ * Returns a GithubRemote object to the sentry release registry.
+ */
+export function getRegistryGithubRemote(): GithubRemote {
+  return new GithubRemote('getsentry', 'sentry-release-registry');
+}
```
tonyo, line 115: nit: I think I get why you put it here, but since it won't be used by any other github api consumers other than in the context of registry interactions, I'd vote for moving it to `src/utils/registry.ts`

src/targets/registry.ts
```diff
@@ -31,13 +31,11 @@ import {
 import * as registryUtils from '../utils/registry';
 import { getPackageDirPath } from '../utils/packagePath';
 import { isDryRun } from '../utils/helpers';
+import { getRegistryGithubRemote } from '../utils/registry';
```
tonyo, line 34: nit: `utils/registry` is already imported as `registryUtils`.

src/targets/awsLambdaLayer.ts
```diff
@@ -143,56 +141,194 @@ export class AwsLambdaLayerTarget extends BaseTarget {
...
+ *
+ * If the dry-run mode is enabled, the release is not pusheable.
+ * If the release is a preview release, unless otherwise stated in the
+ * configuration, the release is not pusheable.
+ * In any other case, the release is pusheable.
+ *
+ * @param version The new version to be released.
+ * @param linkPrereleases Whether the current release is a prerelease.
+ */
+function isPusheableToRegistry(
```
tonyo, line 320:
```suggestion
function isPushableToRegistry(
```

tonyo (approved):
🏄

tonyo:
Oh, and please update the changelog

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.16.2
```
tonyo, line 3: Note: if you don't plan to release a new version immediately after the PR, let's not fill the version right away, but write "Unreleased" in the header. Who knows, maybe the next release will be a major bump

src/targets/awsLambdaLayer.ts
```diff
@@ -143,56 +141,194 @@ export class AwsLambdaLayerTarget extends BaseTarget {
...
+
+        let publishedLayers = [];
+        try {
+          publishedLayers = await layerManager.publishToAllRegions();
+        } catch (error) {
+          logger.error(
+            `Did not publish layers for ${runtime.name}. ` +
+              `Something went wrong with AWS: ${error.message}`
+          );
+          return;
```
BYK, line 217: This seems redundant?

src/targets/awsLambdaLayer.ts
```diff
@@ -143,56 +141,194 @@ export class AwsLambdaLayerTarget extends BaseTarget {
...
+        }
+
+        const regionsVersions = publishedLayers.map(layer => {
+          return {
+            region: layer.region,
+            version: layer.version.toString(),
+          };
+        });
```
BYK, line 248:
```js
const regionsVersions = publishedLayers.map(({region, version}) => ({region, version: version.toString()}));
```

src/targets/awsLambdaLayer.ts
```diff
@@ -143,56 +141,194 @@ export class AwsLambdaLayerTarget extends BaseTarget {
...
+        if (!fs.existsSync(baseFilepath)) {
+          logger.warn(`The ${runtime.name} base file is missing.`);
+          fs.writeFileSync(newVersionFilepath, JSON.stringify(runtimeData));
+        } else {
+          const baseData = JSON.parse(fs.readFileSync(baseFilepath).toString());
+          fs.writeFileSync(
+            newVersionFilepath,
+            JSON.stringify({ ...baseData, ...runtimeData })
+          );
+        }
```
BYK, line 277:
```js
let baseData;
if (fs.existsSync(baseFilePath)) {
  baseData = JSON.parse(fs.readFileSync(baseFilepath, {encoding: 'utf-8'}));
} else {
  baseData = {};
  logger.warn(`The ${runtime.name} base file is missing.`);
}

fs.writeFileSync(
  newVersionFilepath,
  JSON.stringify({ ...baseData, ...runtimeData }),
);

File: src/targets/awsLambdaLayer.ts [hunk 14]
```diff
@@ -143,56 +141,194 @@ export class AwsLambdaLayerTarget extends BaseTarget {
...
+ * @param directory The directory where symlinks will be created.
+ * @param version The new version to be released.
+ * @param versionFilepath Path to the new version file.
+ */
+function createVersionSymlinks(
+  directory: string,
+  version: string,
+  versionFilepath: string
+): void {
+  const latestVersionPath = path.posix.join(directory, 'latest.json');
```
BYK, line 297: `'latest.json'` should probably be a constant:
```js
const LATEST_VERSION_FILE_NAME='latest.json';
```

src/targets/awsLambdaLayer.ts
```diff
@@ -143,56 +141,194 @@ export class AwsLambdaLayerTarget extends BaseTarget {
...
+  if (fs.existsSync(latestVersionPath)) {
+    const previousVersion = fs
+      .readlinkSync(latestVersionPath)
+      .split('.json')[0];
```
BYK, line 301:
```js
const [previousVersion] = fs.readlinkSync(latestVersionPath).split('.json', 1);
```

May consider using [`path.parse()`](https://nodejs.org/api/path.html#path_path_parse_path) or similar path methods too. I checked but didn't find anything that gives you the full path sans the extension but not sure if that's your real intention here. Looks like `path.basename(latestVersionPath, '.json')` would fit here.

src/targets/awsLambdaLayer.ts
```diff
@@ -143,56 +141,194 @@ export class AwsLambdaLayerTarget extends BaseTarget {
...
+  } else {
+    // When no previous versions are found, just create symlinks.
+    createSymlinks(versionFilepath, version);
+  }
+}
+
+/**
+ * Returns whether the current version release should be pushed to the registy.
+ *
+ * If the dry-run mode is enabled, the release is not pusheable.
```
BYK, line 312: s/pusheable/pushable/

src/utils/awsLambdaLayerManager.ts
```diff
@@ -0,0 +1,161 @@
...
+    };
+  }
+
+  /**
+   * Publishes new AWS Lambda layers to all the regions.
+   *
+   * @returns Array of the published layers.
+   */
+  public async publishToAllRegions(): Promise<PublishedLayer[]> {
+    return await Promise.all(
```
BYK, line 104: `return await` is an anti-pattern as it adds an unnecessary cycle wait. Why do you need this?

src/utils/awsLambdaLayerManager.ts
```diff
@@ -0,0 +1,161 @@
...
+    this.awsRegions = awsRegions;
+  }
+
+  /**
+   * Publishes an AWS Lambda layer to the given region.
+   * @param region The AWS region to publish the layer to.
+   * @returns Information about the published layer: region, arn and version.
+   */
+  public async publishLayerToRegion(region: string): Promise<PublishedLayer> {
+    const lambda = new Lambda({ region: region });
```
BYK, line 69:
```js
const lambda = new Lambda({ region });
```

src/utils/awsLambdaLayerManager.ts
```diff
@@ -0,0 +1,161 @@
...
+/**
+ * Requests all regions that are enabled for the current account (or all
+ * regions) to AWS. For more information, see
+ * https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/EC2.html#describeRegions-property
+ */
+export async function getRegionsFromAws(): Promise<
+  DescribeRegionsCommandOutput
+> {
+  logger.debug('Fetching AWS regions...');
+  const ec2 = new EC2({ region: 'us-east-2' });
```
BYK, line 130: Shouldn't this default region be carried out to a global constant?

src/utils/awsLambdaLayerManager.ts
```diff
@@ -0,0 +1,161 @@
...
+export function extractRegionNames(
+  awsRegions: DescribeRegionsCommandOutput
+): string[] {
+  const regionNames: string[] = [];
+  awsRegions.Regions?.map(currentRegion => {
+    if (currentRegion.RegionName !== undefined) {
+      regionNames.push(currentRegion.RegionName);
+    }
+  });
+  return regionNames;
```
BYK, line 151:
```js
return awsRegions.Regions?.filter(currentRegion => currentRegion.RegionName !== undefined) || [];
```

src/utils/awsLambdaLayerManager.ts
```diff
@@ -0,0 +1,161 @@
...
+}
+
+/**
+ * Extracts the AWS account number from the given ARN and returns it
+ * (as a string).
+ * @param arn The ARN of the account.
+ */
+export function getAccountFromArn(arn: string): string {
+  return arn.split(ARN_SEPARATOR)[ARN_ACCOUNT_INDEX];
```
BYK, line 160: This kind of screams for a more readable RegExp to me.

src/utils/registry.ts
```diff
@@ -45,3 +46,10 @@ export function updateManifestSymlinks(
   fs.writeFileSync(versionFilePath, manifestString);
   createSymlinks(versionFilePath, version, previousVersion);
 }
+
+/**
+ * Returns a GithubRemote object to the sentry release registry.
+ */
+export function getRegistryGithubRemote(): GithubRemote {
+  return new GithubRemote('getsentry', 'sentry-release-registry');
```
BYK, line 54: Shouldn't these be constants?

PR #174, opened 2021-02-03
fix: Replace actual versionFilePath

iker-barriocanal:
Fixes an issue where symlinks could not be created, since there's the `versionFilePath` string instead of the actual version file path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 0.17.1
 
-- fix(registry): Replace actual versionFilePath (#174)
+- fix(registry): Replace actual the versionFilePath (#174)
 
 ## 0.17.0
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.17.1
+
+- fix(registry): Replace actual the versionFilePath (#174)
+
```
tonyo, line 6: Since the policy is now "auto":

```suggestion
## Unreleased

- fix(registry): Replace the actual versionFilePath (#174)

```

tonyo (changes requested):

tonyo (approved):
Let's add some kind of test coverage for this later.

PR #177, opened 2021-02-05
feat(github): Retry on 404s

BYK:
Allows retries on `404` responses from GitHub as their API may lag occassionally when making artifacts or check runs available.


chadwhitacre (approved):
Untested, seems reasonabe.

tonyo (approved):

PR #178, opened 2021-02-11
ref(aws-lambda): Catch potential exceptions when publishing AWS Lambda layers to new regions

iker-barriocanal:
AWS regions are requested using the EC2 client, and Lambda layers are published to these regions using the Lambda client. However, it may be possible that an exception is raised (such as a timeout exception). With this PR these exceptions are catched and warned, without stopping Craft. Note that even if there's an exception, layers may be created.

### DIFF
--- a/src/utils/awsLambdaLayerManager.ts
+++ b/src/utils/awsLambdaLayerManager.ts
@@ -66,6 +66,7 @@ export class AwsLambdaLayerManager {
    * @returns Information about the published layer: region, arn and version.
    */
   public async publishLayerToRegion(region: string): Promise<PublishedLayer> {
+    logger.debug(`Publishing layer to ${region}...`);
     const lambda = new Lambda({ region: region });
     const publishedLayer = await lambda.publishLayerVersion({
       Content: {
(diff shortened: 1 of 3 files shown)


HazAT (approved):

src/utils/awsLambdaLayerManager.ts
```diff
@@ -101,11 +102,22 @@ export class AwsLambdaLayerManager {
    * @returns Array of the published layers.
    */
   public async publishToAllRegions(): Promise<PublishedLayer[]> {
-    return await Promise.all(
-      this.awsRegions.map(region => {
-        return this.publishLayerToRegion(region);
+    const publishedLayers = await Promise.all(
```
BYK, line 105: There is [`Promise.allSettled`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) beginning from Node 12.9.

This block would be much easier to follow (and more efficient) with the following code:
```js
const = publishedLayers = await Promise.all(
  this.awsRegions.map(
    async region => this.publishLayerToRegion(region).catch(error => logger.warn(`Something went wrong when publishing to AWS region ${region}: ${error.message}`))
  )
);
```

src/utils/awsLambdaLayerManager.ts
```diff
@@ -101,11 +102,22 @@ export class AwsLambdaLayerManager {
...
+            'Something went wrong with AWS trying to publish to region ' +
+              `${region}: ${error.message}`
+          );
+          return undefined;
+        }
       })
     );
+    return publishedLayers.filter(layer => {
+      return layer !== undefined;
+    }) as PublishedLayer[];
```
BYK, line 120: Better as `return publishedLayers.filter(Boolean);`

PR #185, opened 2021-02-25
Fix: wording for optional post-release script file

marandaneto:
The current wording led me to think that I'm missing something, but this is an optional script file, hence this would make more sense IMO.

### DIFF
--- a/src/commands/publish.ts
+++ b/src/commands/publish.ts
@@ -432,7 +432,7 @@ export async function runPostReleaseCommand(
   } else {
     // Not running post-release command
     logger.info(
-      `Not running the post-release command: '${DEFAULT_POST_RELEASE_SCRIPT_PATH}' not found`
+      `Not running the post-release command: no '${DEFAULT_POST_RELEASE_SCRIPT_PATH}' script file specified`
     );
     return false;
   }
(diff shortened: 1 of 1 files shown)


src/commands/publish.ts
```diff
@@ -432,7 +432,7 @@ export async function runPostReleaseCommand(
   } else {
     // Not running post-release command
     logger.info(
-      `Not running the post-release command: '${DEFAULT_POST_RELEASE_SCRIPT_PATH}' not found`
+      `Not running the post-release command: no '${DEFAULT_POST_RELEASE_SCRIPT_PATH}' script file specified`
```
tonyo, line 435:
```suggestion
      `Not running the post-release command: no post-release script ('${DEFAULT_POST_RELEASE_SCRIPT_PATH}') specified`
```

What about this?

tonyo: Mm, I don't like this either.

marandaneto, line 435:
```suggestion
      `Not running the optional post-release command: '${DEFAULT_POST_RELEASE_SCRIPT_PATH}' not found`
```

tonyo (approved):

PR #186, opened 2021-03-02
feat: Add fastlane target

HazAT:


chadwhitacre:
Is this still worth keeping open, @HazAT?

HazAT:
For context, this PR was meant to be used for releasing our Mobile app with the new workflow. But this is a very early entry point and a lot of things need to be figured out to get there.

kamilogorek:
I'll let myself close this for the sake of keeping this repo clean, but will remember about it if we ever decide to do fastlane.

PR #187, opened 2021-03-12
fix(registry): Ensure up-to-date remote before pushing

BYK:
Fixes getsentry/publish#174

### DIFF
--- a/src/targets/registry.ts
+++ b/src/targets/registry.ts
@@ -341,7 +341,10 @@ export class RegistryTarget extends BaseTarget {
 
     const git = simpleGit(directory).silent(true);
     logger.info(`Cloning "${remote.getRemoteString()}" to "${directory}"...`);
-    await git.clone(remote.getRemoteStringWithAuth(), directory);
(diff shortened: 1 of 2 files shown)


src/targets/registry.ts
```diff
@@ -367,16 +370,18 @@ export class RegistryTarget extends BaseTarget {
     );
 
     // Commit
-    await git.checkout('master');
     await git.add(['.']);
     await git.commit(`craft: release "${canonicalName}", version "${version}"`);
 
+    // Ensure we are still up to date with upstream
+    await git.pull('origin', 'master', { rebase: true });
```
jan-auer, line 377: There's still a chance for a race condition. What if you put this in a bounded loop?

BYK: I think, at our scale it is very unlikely that we'll hit the condition you mention and a bounded loop is still a band-aid fix. When that time comes, we should look into proper queues, not more band aids.

iker-barriocanal (approved):

PR #188, opened 2021-03-19
fix(registry): Specify correct rebase flag

jan-auer:
Passes `--rebase` as option, instead of a literal `rebase` which git interprets as a remote. Additionally, I'm opting for the array-based syntax as it's more concise and in line with other git invocations in the same file. See [how to specify options](https://www.npmjs.com/package/simple-git#how-to-specify-options).

Disclaimer: I did not test this branch. Please test before merging.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## Unreleased
 
-- fix(registry): Ensure up-to-date remote before pushing (#186)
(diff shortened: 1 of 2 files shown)


BYK (approved):

jan-auer:
Verified via manual publish of sentry-javascript 6.2.3

PR #190, opened 2021-03-22
fix(github-artifacts): Better pagination and retry on missing artifacts

### LINKED ISSUE #182: GH Artifact provider: Unable to retrieve artifact list for revision
I'm trying to release [Symbol Collector 1.2.2](https://github.com/getsentry/symbol-collector/actions/runs/585018062) and it's failing while processing the artifact.

[No change to how the artifact is build change since the previous version 1.2.1](https://github.com/getsentry/symbol-collector/compare/1.2.1...main).

It seems zlib related (log below) but I downloaded the artifact locally and could extract files without any problems.

The issue affects version `0.16.1`. I upgraded to the latest `0.17.2` and received the same error.

I've received two different errors that alternate (somewhat randomly) with each other if I retry the command:

```
ℹ info [artifact-provider/github] › Fetching Github artifacts for getsentry/symbol-collector, revision 0dd644d0d67336217dada5da81c9e1b5fa0aabed
ℹ info [artifact-provider/github] › Requesting archive URL from Github...
ℹ info [artifact-provider/github] › Downloading ZIP from Github artifacts...
ℹ info [artifact-provider/github] › Finished downloading.
ℹ info [artifact-provider/github] › Extracting "/var/folders/3r/9m5hqv695t97rd0qllgcbyjc0000gn/T/craft-601315Z0kfxUZaeb3" to "/var/folders/3r/9m5hqv695t97rd0qllgcbyjc0000gn/T/craft-7b1FS2"...
✖ error [artifact-provider] › Unable to retrieve artifact list for revision 0dd644d0d67336217dada5da81c9e1b5fa0aabed!
✖ error Error: unexpected end of file
  at Zlib.zlibOnError [as onerror] (zlib.js:182:17)
```

And
```
ℹ info [artifact-provider/github] › Fetching Github artifacts for getsentry/symbol-collector, revision 0dd644d0d67336217dada5da81c9e1b5fa0aabed
ℹ info [artifact-provider/github] › Requesting archive URL from Github...
ℹ info [artifact-provider/github] › Downloading ZIP from Github artifacts...
ℹ info [artifact-provider/github] › Finished downloading.
ℹ info [artifact-provider/github] › Extracting "/var/folders/3r/9m5hqv695t97rd0qllgcbyjc0000gn/T/craft-60601P38ruEZ8pQlz" to "/var/folders/3r/9m5hqv695t97rd0qllgcbyjc0000gn/T/craft-9Xn1HL"...
✖ error [artifact-provider] › Unable to retrieve artifact list for revision 0dd644d0d67336217dada5da81c9e1b5fa0aabed!
✖ error Error: invalid signature: 0x6f63227b
  at /Users/bruno/.volta/tools/image/packages/@sentry/craft/0.16.1/node_modules/unzipper/lib/parse.js:59:26
  at tryCatcher (/Users/bruno/.volta/tools/image/packages/@sentry/craft/0.16.1/node_modules/bluebird/js/release/util.js:16:23)
  at Promise._settlePromiseFromHandler (/Users/bruno/.volta/tools/image/packages/@sentry/craft/0.16.1/node_modules/bluebird/js/release/promise.js:510:31)
  at Promise._settlePromise (/Users/bruno/.volta/tools/image/packages/@sentry/craft/0.16.1/node_modules/bluebird/js/release/promise.js:567:18)
  at Promise._settlePromise0 (/Users/bruno/.volta/tools/image/packages/@sentry/craft/0.16.1/node_modules/bluebird/js/release/promise.js:612:10)
  at Promise._settlePromises (/Users/bruno/.volta/tools/image/packages/@sentry/craft/0.16.1/node_modules/bluebird/js/release/promise.js:691:18)
  at Async._drainQueue (/Users/bruno/.volta/tools/image/packages/@sentry/craft/0.16.1/node_modules/bluebird/js/release/async.js:133:16)
  at Async._drainQueues (/Users/bruno/.volta/tools/image/packages/@sentry/craft/0.16.1/node_modules/bluebird/js/release/async.js:143:10)
  at Immediate.Async.drainQueues (/Users/bruno/.volta/tools/image/packages/@sentry/craft/0.16.1/node_modules/bluebird/js/release/async.js:17:14)
  at processImmediate (internal/timers.js:461:21)
```

BYK:
GitHub sometimes takes a few minutes until it lists freshly uploaded artifacts on its API. This PR adds a retry mechanism when we cannot find any artifacts for the given revision. It also improves the pagination mechanism to no scan artifacts beyond the revision's commit date to shorten lookup times.

Fixes #182.

### DIFF
--- a/src/artifact_providers/github.ts
+++ b/src/artifact_providers/github.ts
@@ -1,5 +1,4 @@
 import * as Github from '@octokit/rest';
-import * as _ from 'lodash';
 import * as fs from 'fs';
 import * as request from 'request';
 import * as path from 'path';
@@ -20,6 +19,8 @@ import {
   withTempDir,
 } from '../utils/files';
 import { extractZipArchive } from '../utils/system';
+
+const MAX_TRIES = 3;
 
 const logger = loggerRaw.withScope(`[artifact-provider/github]`);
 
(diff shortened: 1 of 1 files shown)


BYK:
This is a draft because I don't know how to test this without screwing something up.

src/artifact_providers/github.ts
```diff
@@ -81,46 +82,74 @@ export class GithubArtifactProvider extends BaseArtifactProvider {
...
+
+      if (revisionDate === undefined) {
+        revisionDate = (await this.github.git.getCommit({
+          owner,
+          repo,
+          commit_sha: revision
+        })).data.committer.date;
+      }
+      const lastArtifact = allArtifacts[allArtifacts.length-1];
+      checkNextPage = lastArtifact.created_at >= revisionDate;
```
tonyo, line 131: Is the "sorted by creation date" guarantee actually provided by the GitHub API we use?

BYK: They don't document any order but it looks like it is inverse creation order from this: https://api.github.com/repos/getsentry/craft/actions/artifacts

And honestly, if it was random, that would be extremely inefficient for GitHub themselves too so I think this logic is safe.

I also checked to see if they support this on their v4 GraphQL API so we can enforce the ordering ourselves, but not there yet ☹️

tonyo: If enough data supports it, I guess we can assume that, but then please add a comment to the code somewhere that it's just our assumption.

Otherwise looks ok, can approve when it's tested.

BYK:
Tested this locally and it works (at least the happy path). It is hard to test the flaky GitHub scenario so we'll have to wait for that one.

src/artifact_providers/github.ts
```diff
@@ -81,46 +82,80 @@ export class GithubArtifactProvider extends BaseArtifactProvider {
...
+    const { repoName: repo, repoOwner: owner } = this.config;
     const per_page = 100;
 
     // https://docs.github.com/en/free-pro-team@latest/rest/reference/actions#artifacts
     const artifactResponse = ((
       await this.github.request('GET /repos/{owner}/{repo}/actions/artifacts', {
-        owner: repoOwner,
-        repo: repoName,
+        owner: owner,
+        repo: repo,
```
jan-auer, line 96: I'm wondering how prettier doesn't reformat this.

```suggestion
        owner,
        repo,
```

jan-auer (approved):
An overall different approach could be to query workflow runs for the release branch and then using the run id query the artifacts for this single run.

src/artifact_providers/github.ts
BYK: :(

Will do in a follow up.

BYK:
>An overall different approach could be to query workflow runs for the release branch and then using the run id query the artifacts for this single run.

Ah, I actually checked that API but didn't look at the parameters to see it allowed filtering based on branch name and trigger. That said I think it would still be a challenge to scan all the runs and get relevant artifacts so our "happy path" would end up making more requests and more complex.

However, _if_ we consider tying status checks to artifacts, we can tell Craft to get an artifact from a specific workflow run. This may cause some brittle config as a downside but might be worth exploring.

PR #191, opened 2021-03-25
docs: Document the `id` field in target config

### LINKED ISSUE #189: Document `id` field for targets
We have added an optional (but very useful) `id` field for target configurations and use this with the `-t` option in #128 and a follow up PR. That said this is not documented properly.

Thanks @jan-auer for reporting.

BYK:
Fixes #189.


README.md
```diff
@@ -452,6 +468,7 @@ The following options can be applied to every target individually:
 | -------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `includeNames` | **optional**. Regular expression: only matched files will be processed by the target. There is one special case that `includeNames` supports, if your build doesn't any artifacts you can write `includeNames: /none/`, this will skip the check for artifacts towards Zeus entirely. |
 | `excludeNames` | **optional**. Regular expression: the matched files will be skipped by the target. Matching is performed after testing for inclusion (via `includeNames`).                                                                                                                            |
+| `id`           | **optional**. A unique id for the target type so one can refert that individually with the `-t` option with the `publish` command. (see the example config above)                                                                                                                     |
```
tonyo, line 471: `refert that` -> `refer to that target`?

Would be also great to have an example of how to do this (is it `-t registry#browser`?) if we don't have it anywhere.

tonyo (changes requested):

tonyo (approved):

PR #192, opened 2021-03-29
feat(publish): Store and restore publish state

BYK:
A very common occurance is publishing being inturrupted in the
middle by some error, causing the whole publish operation to stop
and the operator to resume by explicitly determining the unpublished
targets and then passing these via the `-t` argument.

This patch adds a simple JSON state file, keyed by the release version
that gets deleted after a successful publish. When the file exists, it is
automatically used to skip over already published targets.

### DIFF
--- a/src/commands/publish.ts
+++ b/src/commands/publish.ts
@@ -2,6 +2,12 @@ import * as Github from '@octokit/rest';
 import * as inquirer from 'inquirer';
 import { Arguments, Argv, CommandBuilder } from 'yargs';
 import chalk from 'chalk';
+import {
+  existsSync,
+  readFileSync,
+  writeFileSync,
+  promises as fsPromises,
+} from 'fs';
 import { join } from 'path';
 import * as shellQuote from 'shell-quote';
 import * as stringLength from 'string-length';
@@ -115,6 +121,12 @@ export interface PublishOptions {
   keepBranch: boolean;
 }
 
+export interface PublishState {
+  published: {
+    [targetId: string]: boolean;
+  };
+}
+
 /**
  * Checks that the passed version is a valid version string
  *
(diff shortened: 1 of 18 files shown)


BYK:
I think it is time to rethink how we handle `SpecialTarget.All` and `SpecialTarget.None` but this PR is already large so deferring that work.

src/commands/publish.ts
```diff
@@ -520,6 +489,25 @@ export async function publishMain(argv: PublishOptions): Promise<any> {
...
   }
 
+  logger.info(`Looking for publish state file for ${newVersion}...`);
+  const publishStateFile = `.craft-publish-${newVersion}.json`;
+  const earlierStateExists = existsSync(publishStateFile);
+  let publishState: PublishState;
+  if (earlierStateExists) {
+    logger.info(`Found publish state file, resuming from there...`);
+    publishState = JSON.parse(readFileSync(publishStateFile).toString());
+    targetList = new Set(getAllTargetNames());
```
tonyo, line 499: We should probably rename `targetList` to `targetSet` or similar, to avoid any confusion later e.g. regarding ordering.

src/commands/publish.ts
```diff
@@ -575,6 +589,10 @@ export async function publishMain(argv: PublishOptions): Promise<any> {
       argv.noMerge,
       argv.keepBranch
     );
+    if (!isDryRun()) {
+      // intentionally DO NOT await unlinking
```
tonyo, line 593: Can you explain why we don't want to await here?

src/targets/cocoapods.ts
```diff
@@ -24,7 +24,7 @@ const DEFAULT_COCOAPODS_BIN = 'pod';
 const COCOAPODS_BIN = process.env.COCOAPODS_BIN || DEFAULT_COCOAPODS_BIN;
 
 /** Options for "cocoapods" target */
-export interface CocoapodsTargetOptions extends TargetConfig {
+export interface CocoapodsTargetOptions {
```
tonyo, line 27: What's the main motivation behind this?

src/utils/files.ts
```diff
@@ -83,7 +83,8 @@ export async function withTempDir<T>(
     return await callback(directory);
   } finally {
     if (cleanup) {
-      await rimraf(directory);
+      // Intentionally DO NOT await removal
```
tonyo, line 86: Also here: why the change?

src/commands/publish.ts
BYK: We do not want to delay any further steps to wait for the IO of a file removal that doesn't matter anymore. Also, if the removal fails for some reason `await` raises an exception, blocking further steps, stalling the publish (whereas it actually succeeded).

src/targets/cocoapods.ts
BYK: No motivation, just pleasing TypeScript. Once I made the `name` field required, all these configs started to give type errors, complaining about a missing `name` attribute. Upon inspection, I realized none of these were actually extending `TargetConfig`, there was nothing shared. These were internal configs. What needed to be the `TargetConfig` is the one we pass in the constructor, which I refactored already.

src/utils/files.ts
BYK: Similar to the one above: when I did a dry run, finalizing publish failed on the removal of a temp directory. This is to avoid that.

src/utils/__tests__/files.test.ts
```diff
@@ -28,6 +28,9 @@ describe('withTempDir', () => {
       }, cleanupEnabled);
     } finally {
       if (cleanupEnabled) {
+        // We intentionally do not block on the clean up operation
+        // so wait ~100ms before checking
+        await new Promise(resolve => setTimeout(resolve, 100));
```
chadwhitacre, line 33: Sleep is never a good idea. I mean, sleep is a great idea ... but not here. No better way? Is this related to the await removals that @tonyo is asking about?

chadwhitacre (commented):
Does this actually help us, though, in the context of a GitHub Actions deploy? How will the state file persist from one run to the next?

src/utils/__tests__/files.test.ts
BYK: >Sleep is never a good idea. I mean, sleep is a great idea ... but not here.

Totally agree.

>No better way?

We can try polling with a... timeout?

>Is this related to the await removals that @tonyo is asking about?

Yup :(

BYK:
>Does this actually help us, though, in the context of a GitHub Actions deploy? How will the state file persist from one run to the next?

Not yet. We'll need to hook this up into our fail handler and then store/serialize this data to be unstored/unserialized at the start of the job.

tonyo (approved):

PR #193, opened 2021-03-31
feat(cli): Add a new `targets` command

BYK:
Adds a `targets` command to get a JSON list of available targets for a project. This is to be able to programmatically consume this informatino from places like getsentry/publish.

### DIFF
--- a/src/commands/targets.ts
+++ b/src/commands/targets.ts
@@ -1,0 +1,17 @@
+import { Argv, CommandBuilder } from 'yargs';
+import { getConfiguration } from '../config';
+import { getAllTargetNames, getTargetId } from '../targets';
+
+export const builder: CommandBuilder = (yargs: Argv) => {
+  return yargs;
+};
+
+export function handler(): any {
+  const definedTargets = getConfiguration().targets || [];
+  const possibleTargetNames = new Set(getAllTargetNames());
+  const allowedTargetNames = definedTargets
+    .filter(target => target.name && possibleTargetNames.has(target.name))
+    .map(getTargetId);
+
+  console.log(JSON.stringify(allowedTargetNames));
+}
--- a/src/index.ts
+++ b/src/index.ts
@@ -43,12 +43,9 @@ function processNoInput<T>(arg: T): T {
  * Prints the current version
  */
 function printVersion(): void {
-  if (
-    process.argv.indexOf('-v') === -1 &&
-    process.argv.indexOf('--version') === -1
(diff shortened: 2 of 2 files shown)


src/commands/targets.ts
```diff
@@ -0,0 +1,17 @@
...
+};
+
+export function handler(): any {
+  const definedTargets = getConfiguration().targets || [];
+  const possibleTargetNames = new Set(getAllTargetNames());
+  const allowedTargetNames = definedTargets
+    .filter(target => target.name && possibleTargetNames.has(target.name))
+    .map(getTargetId);
+
+  console.log(JSON.stringify(allowedTargetNames));
```
tonyo, line 16: wdyt about adding indent or using `formatJson` (craft's utility function to print an indented JSON) here?

src/index.ts
```diff
@@ -43,12 +43,9 @@ function processNoInput<T>(arg: T): T {
  * Prints the current version
  */
 function printVersion(): void {
-  if (
-    process.argv.indexOf('-v') === -1 &&
-    process.argv.indexOf('--version') === -1
-  ) {
+  if (process.argv.includes('-v') && process.argv.includes('--version')) {
```
tonyo, line 46: You probably meant "||".

What I don't really like about this is that the function is called `printVersion`, but now we're only printing it in the debug mode. I think I understand the reasoning behind this (so that the output of `targets` is machine-readable), but then at least let's print in debug unconditionally?

src/commands/targets.ts
BYK: Oh, yeah that sounds lovely.

src/index.ts
BYK: >You probably meant "||".

Yup, I did. :)

>I think I understand the reasoning behind this (so that the output of targets is machine-readable), but then at least let's print in debug unconditionally?

Yup but I am not sure what you mean by "unconditionally". That should already be the case?

tonyo: Hmm, wait a sec, what's the expected behavior for you here?
Until now, this function always printed the version unless the "--version" (or the short version) was provided, because in those cases we don't want to print the version twice.
And I'd preserve that logic (just changing the level to debug).

BYK: >And I'd preserve that logic (just changing the level to debug).

Isn't that what I'm doing here? Unless `-v` or `--version` is passed, always log but at `debug` level.

tonyo (approved):

PR #198, opened 2021-04-06
ref(logger): Move some extra info to debug level

BYK:
Our default info level logging is a bit too noisy and outputs redundant information that would only be useful when debugging. This also caused issues withe the new `targets` command on CI as we were always outputting `[no-input]` when we detected CI. This PR moves most of these messages to debug level.

### DIFF
--- a/src/commands/prepare.ts
+++ b/src/commands/prepare.ts
@@ -534,7 +534,7 @@ export async function releaseMain(argv: ReleaseOptions): Promise<any> {
     // Commit the pending changes
     await commitNewVersion(git, newVersion);
   } else {
-    logger.debug('Not commiting anything since preReleaseCommand is empty.');
+    logger.debug('Not committing anything since preReleaseCommand is empty.');
   }
 
   // Push the release branch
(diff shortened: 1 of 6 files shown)


src/index.ts
```diff
@@ -20,7 +20,7 @@ function processDryRun<T>(arg: T): T {
   }
 
   if (isDryRun()) {
-    logger.info('[dry-run] Dry-run mode is on!');
+    logger.debug('[dry-run] Dry-run mode is on!');
```
jan-auer, line 23: This is very useful to see as a confirmation that the dry run flag was accepted, which makes quite a difference when running craft manually.

jan-auer (approved):

src/commands/prepare.ts
```diff
@@ -534,7 +534,7 @@ export async function releaseMain(argv: ReleaseOptions): Promise<any> {
     // Commit the pending changes
     await commitNewVersion(git, newVersion);
   } else {
-    logger.info('Not commiting anything since preReleaseCommand is empty.');
+    logger.debug('Not commiting anything since preReleaseCommand is empty.');
```
tonyo, line 537:
```suggestion
    logger.debug('Not committing anything since preReleaseCommand is empty.');
```

src/index.ts
tonyo: Agreed, let's keep it at `info`.

tonyo (approved):

src/index.ts
BYK:
```suggestion
    logger.info('[dry-run] Dry-run mode is on!');
```

PR #199, opened 2021-04-09
fix(github): Revert retry on 404s

BYK:
This PR reverts #177 as not only it did not address the issue it
was intending to fix, it introduced delays in where we actually
expect a 404 response. It also caused a weird failure where we
threw an error due to a 404 response which should have been ignored:
https://github.com/getsentry/publish/runs/2281793805?check_suite_focus=true#step:8:59

The retry issue was resolved via #190.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- fix(github): Revert retry on 404s (#199)
+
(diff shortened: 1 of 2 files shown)


chadwhitacre (changes requested):
Changelog?

chadwhitacre (approved):

tonyo (approved):

PR #200, opened 2021-04-19
fix(gcs): Fix GCS artifact provider on Windows

BYK:
Use `path.posix` instead of `path` when generating the `prefix` for GCS file listing so Craft works as expected on Windows too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - fix(github): Revert retry on 404s (#199)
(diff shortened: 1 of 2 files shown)


jan-auer (commented):
There are more cases where we call `path.join` in this file that would need POSIX joins instead: lines 217, 229, 245.

I made a quick scan through the repo and it looks like GCS is the only broken target.

BYK:
@jan-auer fixed those others too

jan-auer (approved):
LGTM, and great catch!

PR #202, opened 2021-04-19
feat(config): Use GitHub as default provider

BYK:
This PR switches default artifact and status providers to GitHub if the minimum required Craft version is set to 0.21.0 or above. This is *not* a breaking change for older projects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,7 +180,7 @@ Options:
 The command will find a release branch for the provided version. The normal flow
 is for this release branch to be created automatically by `craft prepare`, but
 that's not strictly necessary. Then, it subscribes to the latest status checks on
-that branch. Once the checks pass, it downloads the release artifacts from the 
+that branch. Once the checks pass, it downloads the release artifacts from the
 artifact provider configured in `.craft.yml` and uploads them to the targets named
 on the command line (and pre-configured in `.craft.yml`).
 
@@ -375,14 +375,13 @@ requireNames:
 ## Status Provider
 
 You can configure which status providers `craft` will use to check for your build status.
-By default, it will take Zeus but you can also use GitHub directly.
-This is helpful if you don't want to rely on Zeus for asking if you build is green or not.
+By default, it will use GitHub but you can add more providers if needed.
 
 **Configuration**
 
 | Option   | Description                                                                                        |
 | -------- | -------------------------------------------------------------------------------------------------- |
-| `name`   | Name of the status provider: either `zeus` (default) or `github`                                   |
+| `name`   | Name of the status provider: either `github` (default) or `zeus` (deprecated)                                   |
 | `config` | In case of `github`: may include `contexts` key that contains a list of required contexts (checks) |
 
 **Example:**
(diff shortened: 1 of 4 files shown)


src/config.ts
```diff
@@ -299,3 +334,12 @@ export function getStatusProviderFromConfig(): BaseStatusProvider {
     }
   }
 }
+
+export function getChangelogPolicyFromConfig(): ChangelogPolicy {
+  const config = getConfiguration() || {};
```
chadwhitacre, line 339: Why do we need the `|| {}` here but not above?

chadwhitacre (approved):
Seems reasonable, untested, one nit.

src/config.ts
```diff
@@ -195,6 +201,28 @@ export function checkMinimalConfigVersion(): void {
...
+  if (!minVersionRaw) {
+    return false;
+  }
+
+  const minVersion = parseVersion(minVersionRaw);
+  if (!minVersion) {
+    throw new Error(`Cannot parse the minimal version: "${minVersionRaw}"`);
+  }
+
+  const epochVersion = parseVersion(DEFAULTS_EPOCH_VERSION);
```
jan-auer, line 216: The `epochVersion` should probably be passed in, once you have multiple epochs. Let's say we change the default for the preReleaseCommand in one of the next versions.

src/config.ts
```diff
@@ -299,3 +334,12 @@ export function getStatusProviderFromConfig(): BaseStatusProvider {
     }
   }
 }
+
+export function getChangelogPolicyFromConfig(): ChangelogPolicy {
+  const config = getConfiguration() || {};
+
+  return (
+    config.changelogPolicy ??
+    (isAfterEpoch() ? ChangelogPolicy.Auto : ChangelogPolicy.None)
```
jan-auer, line 343: Let's check with SDK devs that we actually want this change. I believe we still haven't found ideal tooling to ensure changelogs are always kept up-to-date in our development flow, so a lot of repositories would rely on `simple` to error when the changelog is missing.

jan-auer (commented):
Code looks good, a few comments regarding the approach below.

jan-auer (commented):
Code looks good, a few comments regarding the approach below.

src/config.ts
BYK: Just copy pasted. I can do a go over for all usages.

src/config.ts
BYK: Let's revisit that when the need arises, wdyt?

src/config.ts
BYK: Fair point, I can revert this one

BYK:
@jan-auer shall we emit a warning when we detect the old or the new default pre-epoch to urge people to upgrade?

tonyo (approved):

src/config.ts
jan-auer: Sure, that works for me 👍

jan-auer (approved):
Before merging, please update the defaults in the readme with a note to the version where this changes:

- https://github.com/getsentry/craft#status-provider
- https://github.com/getsentry/craft#artifact-provider

If I may suggest, it would be great to have a verbose message this time in the changelog notifying of the config values that are changing and what config to add in case of upgrading. Something like:

> **Breaking Change**: This version changes the defaults configuration values for `statusProvider` and `artifactProvider` to `"github"` if `minVersion` is greater or equal to `0.21.0`. If your craft configuration file does not set these providers explicitly, you can keep the old defaults by adding this to your config:
> 
> ```yaml
> minVersion: 0.21.0
> artifactProvider: zeus
> statusProvider: zeus
> ```
> 
> Support for Zeus will be dropped in a future release and we highly recommend updating your CI workflows.

README.md
```diff
@@ -958,17 +934,33 @@ targets:
...
-    default, configurable through `releaseBranchPrefix`)
-  - Add ZEUS_HOOK_BASE as protected to CI environment
-- Add `.craft.yml` configuration file to your project
+1. Set up a workflow that builds your assets and runs your tests. Allow building
+   release branches (their names follow `release/{VERSION}` by default,
+   configurable through `releaseBranchPrefix`).
+2. Use the official `actions/upload-artifact@v2` action to upload your assets.
+   Here is an example config (step) of an archive job:
+
+    ```yaml
```
jan-auer, line 943: Indentation is one off, does it copy/paste fine from the rendering?

jan-auer (approved):

README.md
BYK: Possible. Will fix.

PR #203, opened 2021-04-20
build: Move to single-file builds for craft

BYK:
This PR reduces the build times dramatically (10x) by generating
a minimized single-file build of Craft binary via `esbuild`. Using
a single-file build makes is more portable so one can simply do
`curl <CRAFT-URL>` and use craft directly instead of doing an
`npm install`. This is especially useful for our centralized publish
repo flows and actions.

**IMPORTANT:** This PR sacrifices the update checker as it uses lazy imports for certain things. We can try to make it work with static builds but I'm not sure if it's a feature that's worth the effort.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,13 +2,13 @@
   "name": "@sentry/craft",
   "version": "0.20.0",
   "description": "The universal sentry workflow CLI",
-  "main": "dist/index.js",
+  "main": "dist/craft",
   "repository": "https://github.com/getsentry/craft",
   "author": "Sentry",
   "license": "MIT",
   "bin": {
-    "craft": "dist/index.js",
-    "sentry-craft": "dist/index.js"
+    "craft": "dist/craft",
+    "sentry-craft": "dist/craft"
   },
   "resolutions": {
     "**/set-value": ">=2.0.1",
@@ -77,6 +77,7 @@
     "@types/yargs": "^15.0.3",
     "@typescript-eslint/eslint-plugin": "^3.3.0",
     "@typescript-eslint/parser": "^3.3.0",
+    "esbuild": "^0.11.6",
     "eslint": "^7.2.0",
     "eslint-config-prettier": "^6.11.0",
     "jest": "^26.5.3",
@@ -87,17 +88,18 @@
     "typescript": "^4.1.3"
   },
   "scripts": {
-    "build": "yarn run compile-config-schema && tsc -p tsconfig.build.json",
+    "build:fat": "yarn run compile-config-schema && tsc -p tsconfig.build.json",
     "build:watch": "yarn run compile-config-schema && tsc -p tsconfig.build.json --watch",
+    "build": "yarn compile-config-schema && esbuild src/index.ts --bundle --platform=node --target=node10.8 --outfile=dist/craft --minify",
     "precli": "run-s build",
-    "cli": "node dist",
+    "cli": "node dist/craft",
     "clean": "rimraf dist coverage",
     "lint": "run-s lint:prettier lint:eslint",
     "lint:prettier": "prettier-check 'src/**/*.ts'",
     "lint:eslint": "eslint src  --ext .ts --format stylish",
     "fix": "run-s fix:tslint fix:prettier",
     "fix:prettier": "prettier --write 'src/**/*.ts'",
-    "fix:eslint": "eslint src  --ext .ts --fix --format stylish",
+    "fix:eslint": "eslint src --ext .ts --fix --format stylish",
     "test": "jest",
     "test:watch": "jest --watch --notify",
     "compile-config-schema": "node ./scripts/config-json-schema-to-ts.js"
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
   "resolutions": {
     "**/set-value": ">=2.0.1",
     "**/https-proxy-agent": ">=2.2.3",
+    "**/is-ci": "^3.0.0",
     "**/node-forge": ">=0.10.0",
     "**/dot-prop": ">=4.2.1",
     "**/kind-of": ">=6.0.3",
@@ -33,6 +34,7 @@
     "cli-table": "0.3.1",
     "consola": "1.4.5",
     "inquirer": "6.2.1",
+    "is-ci": "^3.0.0",
     "js-yaml": "3.12.0",
     "json-schema-to-typescript": "5.7.0",
     "lodash": "4.17.19",
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "tar": "4.4.8",
     "tmp": "0.1.0",
     "unzipper": "0.9.7",
+    "update-notifier": "^5.1.0",
     "yargs": "15.1.0"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "tar": "4.4.8",
     "tmp": "0.1.0",
     "unzipper": "0.9.7",
-    "update-notifier": "2.5.0",
+    "update-notifier": "^5.1.0",
     "yargs": "15.1.0"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -50,7 +50,6 @@
     "tar": "4.4.8",
     "tmp": "0.1.0",
     "unzipper": "0.9.7",
-    "update-notifier": "^5.1.0",
     "yargs": "15.1.0"
   },
   "devDependencies": {
@@ -90,7 +89,7 @@
   "scripts": {
     "build:fat": "yarn run compile-config-schema && tsc -p tsconfig.build.json",
     "build:watch": "yarn run compile-config-schema && tsc -p tsconfig.build.json --watch",
-    "build": "yarn compile-config-schema && esbuild src/index.ts --bundle --platform=node --target=node10.8 --outfile=dist/craft --minify",
+    "build": "yarn compile-config-schema && esbuild src/index.ts --define:importLazy=require --bundle --platform=node --target=node10.8 --outfile=dist/craft",
     "precli": "run-s build",
     "cli": "node dist/craft",
     "clean": "rimraf dist coverage",
--- a/package.json
+++ b/package.json
@@ -50,7 +50,6 @@
     "tar": "4.4.8",
     "tmp": "0.1.0",
     "unzipper": "0.9.7",
-    "update-notifier": "^5.1.0",
     "yargs": "15.1.0"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -89,7 +89,7 @@
   "scripts": {
     "build:fat": "yarn run compile-config-schema && tsc -p tsconfig.build.json",
     "build:watch": "yarn run compile-config-schema && tsc -p tsconfig.build.json --watch",
-    "build": "yarn compile-config-schema && esbuild src/index.ts --define:importLazy=require --bundle --platform=node --target=node10.8 --outfile=dist/craft",
+    "build": "yarn compile-config-schema && esbuild src/index.ts --bundle --platform=node --target=node10.8 --outfile=dist/craft --minify",
     "precli": "run-s build",
     "cli": "node dist/craft",
     "clean": "rimraf dist coverage",
--- a/package.json
+++ b/package.json
@@ -13,7 +13,6 @@
   "resolutions": {
     "**/set-value": ">=2.0.1",
     "**/https-proxy-agent": ">=2.2.3",
-    "**/is-ci": "^3.0.0",
     "**/node-forge": ">=0.10.0",
     "**/dot-prop": ">=4.2.1",
     "**/kind-of": ">=6.0.3",
@@ -34,7 +33,6 @@
     "cli-table": "0.3.1",
     "consola": "1.4.5",
     "inquirer": "6.2.1",
-    "is-ci": "^3.0.0",
     "js-yaml": "3.12.0",
     "json-schema-to-typescript": "5.7.0",
     "lodash": "4.17.19",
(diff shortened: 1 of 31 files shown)


tonyo (approved):
🏄

chadwhitacre:
> This PR sacrifices the update checker as it uses lazy imports for certain things. We can try to make it work with static builds but I'm not sure if it's a feature that's worth the effort.

@BYK You and I talked about this elsewhere, doesn't feel great to sacrifice functionality for technical reasons. If we must, we must.

BYK:
@tonyo I took the path for adding `update-notifier` back and actually us using `is-ci` to populate the default value of the `--no-input` argument. Thoughts?

src/index.ts
```diff
@@ -76,7 +86,7 @@ function main(): void {
     .option('no-input', {
       boolean: true,
       coerce: once(processNoInput),
-      default: false,
+      default: isCI,
```
chadwhitacre, line 89: Do I understand correctly (from [this comment](https://github.com/getsentry/craft/pull/203#issuecomment-824254467)) that this is intended to work around the `esbuild` warning with `update-notifier` lazy-loading? If so, what happens if I run `esbuild`-built `craft` outside of CI? 🤔

At the very least we should update the README/usage docs to reflect what we're doing here.

src/utils/env.ts
```diff
@@ -109,7 +109,7 @@ export function readEnvironmentConfig(overwriteExisting = false): void {
   let newEnv = {} as any;
 
   // Read from home dir
-  const homedirEnvFile = join(homedir(), ENV_FILE_NAME);
```
chadwhitacre, line 112: These are all automated linting changes, right? Would've been better to do a separate PR for these so that the material changes are obvious.

tsconfig.build.json
```diff
@@ -1,12 +1,14 @@
...
   "compilerOptions": {
     "target": "es2017",
     "baseUrl": ".",
     "outDir": "dist",
     "rootDir": "src",
     "types": ["node"],
-    "noImplicitThis": false
+    "noImplicitThis": false,
+    "esModuleInterop": true,
+    "importHelpers": true
```
chadwhitacre, line 11: Are these somehow related to the import changes I'm seeing throughout the PR? Are these part-and-parcel of the task at hand or incidental?

src/utils/version.ts
```diff
@@ -154,13 +151,9 @@ export function checkForUpdates(): void {
...
-  const message = `Update available ${chalk.dim(
-    notifier.update.current
-  )}${chalk.reset(' → ')}${chalk.green(
-    notifier.update.latest
-  )}\nRun ${chalk.cyan('yarn global add ')}${chalk.cyan(pkg.name)} to update`;
-
-  notifier.notify({ defer: false, message });
+  notifier.notify({
+    defer: false,
+  });
```
chadwhitacre, line 156: What is the default message, now that we're no longer passing one?

package.json
```diff
@@ -87,17 +90,18 @@
     "typescript": "^4.1.3"
   },
   "scripts": {
-    "build": "yarn run compile-config-schema && tsc -p tsconfig.build.json",
+    "build:fat": "yarn run compile-config-schema && tsc -p tsconfig.build.json",
     "build:watch": "yarn run compile-config-schema && tsc -p tsconfig.build.json --watch",
+    "build": "yarn compile-config-schema && esbuild src/index.ts --bundle --platform=node --target=node10.8 --outfile=dist/craft --minify",
```
chadwhitacre, line 95: This is the important part.

chadwhitacre: 👍 

```
$ yarn && yarn build
[...]
$ ./dist/craft 
craft <command>

Commands:
  craft prepare NEW-VERSION  🚢 Prepare a new release branch
                          [aliases: p, prerelease, prepublish, prepare, release]
  craft publish NEW-VERSION  🛫 Publish artifacts         [aliases: pp, publish]
  craft targets              List defined targets as JSON array
  craft artifacts <command>  📦 Manage artifacts          [aliases: a, artifact]

Options:
  --no-input     Suppresses all user prompts          [boolean] [default: false]
  --dry-run      Dry run mode: do not perform any real actions
                                                      [boolean] [default: false]
  -v, --version  Show version number                                   [boolean]
  -h, --help     Show help                                             [boolean]

Not enough non-option arguments: got 0, need at least 1
$
```

chadwhitacre: Are all six of these warnings okay?

```
$ node ./scripts/config-json-schema-to-ts.js
 > node_modules/yargs/index.js:12:39: warning: Indirect calls to "require" will not be bundled (surround with a try/catch to silence this warning)
    12 │   const argv = yargs(processArgs, cwd, require)
       ╵                                        ~~~~~~~

 > node_modules/yargs/yargs.js:378:33: warning: Indirect calls to "require" will not be bundled (surround with a try/catch to silence this warning)
    378 │     const req = parentRequire || require
        ╵                                  ~~~~~~~

 > node_modules/update-notifier/index.js:5:42: warning: Indirect calls to "require" will not be bundled (surround with a try/catch to silence this warning)
    5 │ const importLazy = require('import-lazy')(require);
      ╵                                           ~~~~~~~

 > node_modules/require-main-filename/index.js:2:25: warning: Indirect calls to "require" will not be bundled (surround with a try/catch to silence this warning)
    2 │   _require = _require || require
      ╵                          ~~~~~~~

 > node_modules/yargs/lib/apply-extends.js:57:82: warning: This call to "require" will not be bundled because the argument is not a string literal (surround with a try/catch to silence this warning)
    57 │ ... isPath ? JSON.parse(fs.readFileSync(pathToDefault, 'utf8')) : require(config.extends)
       ╵                                                                   ~~~~~~~

 > node_modules/yargs-parser/build/index.cjs:1005:19: warning: This call to "require" will not be bundled because the argument is not a string literal (surround with a try/catch to silence this warning)
    1005 │             return require(path);
         ╵                    ~~~~~~~

6 warnings
```

src/commands/artifacts.ts
```diff
@@ -20,4 +23,10 @@ export const builder: CommandBuilder = (yargs: Argv) =>
     })
     .demandCommand()
     .demandOption('rev', 'Please specify the revision')
-    .commandDir('artifacts_cmds');
+    .command(list)
+    .command(download);
```
chadwhitacre, line 27: What's the point of this change? Looks like more than just the minor import changes I'm seeing in most other files.

chadwhitacre (changes requested):

package.json
chadwhitacre: Also, I expected to be able to move the `craft` executable around and have it still work but it doesn't.

```
$ mv craft ~/Desktop
$ cd ~/Desktop
$ ./craft 
internal/modules/cjs/loader.js:883
  throw err;
  ^

Error: Cannot find module 'is-ci'
Require stack:
- /Users/chadwhitacre/Desktop/craft
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:880:15)
    at Function.Module._load (internal/modules/cjs/loader.js:725:27)
    at Module.require (internal/modules/cjs/loader.js:952:19)
    at require (internal/modules/cjs/helpers.js:88:18)
    at a5 (/Users/chadwhitacre/Desktop/craft:148:1046)
    at /Users/chadwhitacre/Desktop/craft:148:1129
    at new cce (/Users/chadwhitacre/Desktop/craft:148:2429)
    at ace.exports (/Users/chadwhitacre/Desktop/craft:152:494)
    at sze (/Users/chadwhitacre/Desktop/craft:610:2291)
    at Yus (/Users/chadwhitacre/Desktop/craft:649:699)
    at Object.<anonymous> (/Users/chadwhitacre/Desktop/craft:649:1229)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:72:12) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [ '/Users/chadwhitacre/Desktop/craft' ]
}
$
```

What is the expected behavior here?

src/index.ts
BYK: > If so, what happens if I run esbuild-built craft outside of CI? 🤔

Don't see how this is relevant. I now made `is-ci` an explicit required dependency of the project. We use it to auto detect/set the value of `--no-input` argument. Not sure if it needs a comment as if somebody removes this require/usage, builds will start failing.

tsconfig.build.json
BYK: These are part-and-parcel of the task at hand as without these changes, `esbuild` cannot build the package. The reason being TypeScripts's legacy `import` behavior before ES6 module standard was established.

src/utils/version.ts
BYK: This is what the default message looks like:


We added the customization to suggest `yarn` when needed but this was done upstream now: yeoman/update-notifier#132

package.json
BYK: >Are all six of these warnings okay?

Yes

>Also, I expected to be able to move the craft executable around and have it still work but it doesn't.

No, this means you are looking at an older version of this PR or something. (otherwise the docker image check would fail too with the same error)

src/commands/artifacts.ts
BYK: It is to overcome `yargs`' dynamic nested command generation. To be able to build with esbuild, all imports should be static whereas `yargs` was doing a directory scan with dynamic imports to build this on-the-fly.

package.json
BYK: >No, this means you are looking at an older version of this PR or something. (otherwise the docker image check would fail too with the same error)

Oh, it _is_ failing. I'll fix this.

src/index.ts
chadwhitacre: Right, but why are we doing that on this PR? Is it related somehow to `update-notifier`?

BYK: Yes, without us requiring this library explicitly, `update-notifier` fails as it lazily imports this library and since nothing else refers to it, it is not bundeled.

chadwhitacre: Ah, gotcha gotcha, so `update-notifier` requires this, and since we now need to explicitly require it we figured why not make use of it elsewhere, something like that?

BYK:
@chadwhitacre I walked back from using `update-notifier` as it adds unnecessary bulk and even if there is a way for the author to make it compatible with static linker, they were dismissed without much thought (https://github.com/yeoman/update-notifier/issues/123#issuecomment-334099457, https://github.com/yeoman/update-notifier/issues/159#issuecomment-494054084). There is [a fork](https://www.npmjs.com/browse/depended/update-notifier-webpack) but I don't think this is worth such an effort. If we find this feature, I'm quite sure we can implement a simpler version without all this pain.

src/index.ts
BYK: >Ah, gotcha gotcha, so update-notifier requires this, and since we now need to explicitly require it we figured why not make use of it elsewhere, something like that?

Yup, but turns out even that doesn't work so 🤷🏻‍♂️

chadwhitacre (approved):
Approving with the following caveats:

1. It's too bad we have to choose between `update-notifier` and `esbuild`. :-/
2. I don't see that we actually use the Docker image ourselves, not sure why we bother building that.
3. I'm not sure I trust that we won't hit a bug at some point because of [the other warnings](https://github.com/getsentry/craft/pull/203/files#r618421507) we're ignoring.
4. I don't really understand why we're adding `dist/craft` to the GitHub artifacts we're uploading, since we will be `curl`ing from elsewhere in `action-prepare-release` (and continuing to craft as an action via GitHub ref in `prepare`).

BYK:
>It's too bad we have to choose between update-notifier and esbuild. :-/

We can get similar functionality by ourselves or resort back to using `update-notifier-webpack` if we find it necessary. I'm not willing to depend on a package where the author is dismissive of simple requests to make the package easier to use.

>I don't see that we actually use the Docker image ourselves, not sure why we bother building that.

It is being used by the Craft GitHub Action: https://github.com/getsentry/craft/blob/d34cf6531c3ca5e46dfdcae962fdb8d7916fd6ce/action.yml#L17-L24

This action is being used in the publish repo: https://github.com/getsentry/publish/blob/514489dfce586772287ce3bcc7a01c86f49b1ce1/.github/workflows/publish.yml#L60-L65

>I'm not sure I trust that we won't hit a bug at some point because of the other warnings we're ignoring.

I'll revisit our `yargs` usage where most of the warnings are coming from. I made sure all dynamic require paths are covered with the changes in this patch.

>I don't really understand why we're adding dist/craft to the GitHub artifacts we're uploading, since we will be curling from elsewhere in action-prepare-release (and continuing to craft as an action via GitHub ref in prepare).

These will be used when cutting a release to GitHub where we attach all the assets from the build to the release tag.

chadwhitacre:
> These will be used when cutting a release to GitHub where we attach all the assets from the build to the release tag.

... and presumably this copies the assets over so the 90-day retention period doesn't apply.

BYK:
>... and presumably this copies the assets over so the 90-day retention period doesn't apply.

Yup. As far as I'm aware release assets are stored indefinitely.

chadwhitacre:
> cutting a release to GitHub where we attach all the assets from the build to the release tag

https://github.com/getsentry/craft/blob/7bf3f70009f32f6a9c20bb9127488a5fb8c66c68/src/targets/github.ts#L267

https://github.com/getsentry/craft/blob/7bf3f70009f32f6a9c20bb9127488a5fb8c66c68/src/targets/github.ts#L306

👍

PR #207, opened 2021-04-26
fix: Upgrade simple-git to latest version

BYK:
We have locally reproduced reports of `git.tags()` not working, which is used in `prepare` step with the new builds. I assumed this to be a bug with `simple-git` and upgraded it. Now the issue seems resolved so let's go with it.


chadwhitacre (approved):

chadwhitacre:
Approved. Would be great to see a regression test, though. I'm not sure what `git.tags` calls we're fixing here, nor how they broke.

chadwhitacre:
Presumably this one, it's the only one I find:

https://github.com/getsentry/craft/blob/61c971d94c6fa29e28a66e6105cf0c618a435af2/src/commands/prepare.ts#L354

chadwhitacre:
And ftr per @jennmueng in private Slack here's how the bug manifests:

```
npm WARN saveError ENOENT: no such file or directory, open '/tmp/.craft_tTX/package.json'
npm notice created a lockfile as package-lock.json. You should commit this file.
npm WARN enoent ENOENT: no such file or directory, open '/tmp/.craft_tTX/package.json'
npm WARN .craft_tTX No description
npm WARN .craft_tTX No repository field.
npm WARN .craft_tTX No README data
npm WARN .craft_tTX No license field.

+ @sentry/craft@0.21.0
added 1 package from 1 contributor and audited 1 package in 3.155s
found 0 vulnerabilities

[12:23:35 PM] TypeError: Cannot read property 'indexOf' of undefined
at Jls (/tmp/.craft_tTX/node_modules/@sentry/craft/dist/craft:639:2508)
at processTicksAndRejections (internal/process/task_queues.js:93:5)
at async $Jt (/tmp/.craft_tTX/node_modules/@sentry/craft/dist/craft:643:514)
at async Object.Zls [as handler] (/tmp/.craft_tTX/node_modules/@sentry/craft/dist/craft:643:1210)
```

https://github.com/getsentry/sentry-react-native/runs/2438165609?check_suite_focus=true

PR #208, opened 2021-04-28
feat(config): Automatically detect GitHub config when missing

BYK:
We don't need people to explicitly state `github` global configuration
as it can be detected from `git` itself. Since we rely on the presence
of a `git` repo in `publish` and `prepare` we can safely rely on this
too and remove another piece of redundant configuration.

Most of the changes in this patch are related to the conversion of `getGlobalGithubConfig` from a sync function to an async function. There were also a few places which relied on `config.github` directly which is also removed now.

Ideally we'd abstract the raw config object in the future and provide a global, normalized configuration to all downstream dependent code.

~~This patch also adds tests around this new functionality.~~


BYK:
I will add the tests for this new functionality separately as they cause some very hard to debug import errors. I'm guessing it has something to do with Jest and mocking but I've already spent multiple full days on debugging without any luck and now this PR blocks #209 (due to its `github` config refactor).

package.json
```diff
@@ -57,6 +57,8 @@
     "esbuild": "^0.11.6",
     "eslint": "^7.2.0",
     "eslint-config-prettier": "^6.11.0",
+    "git-url-parse": "^11.4.4",
+    "inquirer": "6.2.1",
```
chadwhitacre, line 62: lol d'oh!

src/config.ts
```diff
@@ -225,26 +227,44 @@ export function isAfterEpoch(): boolean {
...
+    }
   }
 
-  if (!repoGithubConfig.repo) {
-    throw new ConfigurationError('GitHub target: repo not found');
-  }
+  _globalGithubConfigCache = Object.freeze(repoGithubConfig);
 
-  return repoGithubConfig;
+  return getGlobalGithubConfig();
```
chadwhitacre, line 267: No `await` here because it is recursive. 👍

src/targets/base.ts
```diff
@@ -20,13 +20,17 @@ export class BaseTarget {
...
   /** Artifact filtering options for the target */
   public readonly filterOptions: FilterOptions;
+  /** Github repo configuration */
+  public readonly githubRepo: GithubGlobalConfig;
 
   public constructor(
     config: TargetConfig,
-    artifactProvider: BaseArtifactProvider
+    artifactProvider: BaseArtifactProvider,
+    githubRepo: GithubGlobalConfig
```
chadwhitacre, line 29: Can we default this to null/undefined to avoid having to modify every target, since most targets don't need this?

package.json
BYK: Ugh, will fix good catch!

src/__tests__/config.test.ts
```diff
@@ -39,20 +39,15 @@ describe('validateConfiguration', () => {
...
 
-  test('fails with empty configuration', () => {
-    // this ensures that we do actually run the expect line in the catch block
-    expect.assertions(1);
-
-    try {
-      validateConfiguration({});
-    } catch (e) {
-      expect(e.message).toMatch(/should have required property/i);
-    }
```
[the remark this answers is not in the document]
BYK: Because the only required field was `github` in the config and with that also becoming optional it does **not** fail with empty config: the test case is now obsolete :)

src/commands/prepare.ts
```diff
@@ -481,12 +482,12 @@ export async function releaseMain(argv: ReleaseOptions): Promise<any> {
...
   const config = getConfiguration();
-  const githubConfig = config.github;
+  const githubConfig = await getGlobalGithubConfig();
 
   // Move to the directory where the config file is located
-  const configFileDir = dirname(getConfigFilePath());
+  const configFileDir = getConfigFileDir() || '.';
   process.chdir(configFileDir);
-  logger.debug(`Working directory:`, configFileDir);
+  logger.debug(`Working directory:`, process.cwd());
```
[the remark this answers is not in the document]
BYK: I feel like I made it more accurate 😁

src/config.ts
```diff
@@ -225,26 +227,44 @@ export function isAfterEpoch(): boolean {
 /**
  * Return the parsed global Github configuration
  */
-export function getGlobalGithubConfig(): GithubGlobalConfig {
+let _globalGithubConfigCache: GithubGlobalConfig | null;
```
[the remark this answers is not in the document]
BYK: We are making at least one call to the `git` executable which is not cheap.

src/config.ts
BYK: smart, right?

src/targets/base.ts
BYK: I'm not a fan of that as then we need to do non-null checks down the line. This will especially be pronounced in `publish` where we get all the targets without knowing the exact target type (as it is determined at runtime).

src/config.ts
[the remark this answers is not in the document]
chadwhitacre: I don't buy it. Git is not interpreted (is it?) so it should not be expensive. This feels like over-complication to me.

src/targets/base.ts
chadwhitacre: Do we? Why would we need to check for null if the only targets that use `githubRepo` are the ones that will actually have it in the constructor?

chadwhitacre (approved):

src/config.ts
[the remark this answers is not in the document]
BYK: It is the cost of spawning a new process, not related to `git`. That said, to get the information we need, `git` touches the filesystem multiple times which is also costly.

src/targets/base.ts
BYK: Even if they have it in the constructor, the type system will force us to make sure it is not null. We _may_ be able to overcome this so I'll look into that later.

PR #212, opened 2021-04-30
upgrade(ts-jest): Upgrade ts-jest to latest version

BYK:


jan-auer (commented):
Meta discussion: Do PRs like 210, 211 and this one really need a changelog entry? Maybe we could collect them in a `**Internal**` section, since they are unlikely to affect users.

BYK:
@jan-auer 
>Meta discussion: Do PRs like 210, 211 and this one really need a changelog entry? Maybe we could collect them in a **Internal** section, since they are unlikely to affect users.

I think it is fine to omit them. Just trying to form the habit of adding changelog entries :)

jan-auer (approved):

PR #213, opened 2021-04-30
ref: Eradicate lodash from project

BYK:


src/artifact_providers/zeus.ts
```diff
@@ -160,25 +159,26 @@ export class ZeusArtifactProvider extends BaseArtifactProvider {
...
+        }
       }
-    );
+    });
 
-    return dedupedZeusArtifacts.map(zeusArtifact =>
-      this.convertToRemoteArtifact(zeusArtifact)
-    );
+    return zeusArtifacts
+      .filter((artifact, idx, arr) => artifact.name !== arr[idx + 1]?.name)
```
jan-auer, line 181: This sort + filter is correct, but it is quite hard to read and understand now. Could you make this two for plain old `for` loops for readability, instead? Or, once we drop support for Node 10, we can instead switch to `Object.fromEntries`. 

1. step: get maximum artifact by name
2. step: map and convert

Same in the store impl.

src/commands/artifacts_cmds/download.ts
```diff
@@ -94,7 +94,13 @@ async function handlerMain(argv: ArtifactsDownloadOptions): Promise<any> {
   const filesToDownload = argv.all
     ? artifacts.map(ar => ar.filename)
     : argv.names;
-  const nameToArtifact = _.fromPairs(artifacts.map(ar => [ar.filename, ar]));
+  const nameToArtifact = artifacts.reduce(
+    (dict, artifact) => {
+      dict[artifact.filename] = artifact;
+      return dict;
```
jan-auer, line 100: Starting with Node 8.3, you can use object spread here for better readability, which might even remove the need for the type annotation below. Same in `crates/getPublishOrder`.

tonyo (commented):
Before reviewers dig into this, could you elaborate on "why"?
It'll have to stay as subdependency anyway, and up to this point it just worked.

src/targets/registry.ts
```diff
@@ -270,12 +269,12 @@ export class RegistryTarget extends BaseTarget {
 
     await mapLimit(artifacts, MAX_DOWNLOAD_CONCURRENCY, async artifact => {
       const fileData = await this.getArtifactData(artifact, version, revision);
-      if (!_.isEmpty(fileData)) {
+      if (Object.keys(fileData).length > 0) {
```
jan-auer, line 272: _nit_: This could be put behind a utility function to show clear intent. 

If someone wants to optimize, they can then easily switch this out for an implementation that doesn't require collecting keys (FWIW, misses `hasOwnProperty` check):


```js
function isEmpty(obj) {
  for (key in obj) return true;
  return false;
}
```

jan-auer (approved):

jan-auer:
Sorry, I didn't intend this to be merged right away. Did not notice that auto-merge was enabled from the diff screen.
@BYK Please consider my comments and also give @tonyo the chance to review this.

src/artifact_providers/zeus.ts
BYK: Yup, agreed. This is the kind of feedback I was looking for, thanks a lot!

src/commands/artifacts_cmds/download.ts
BYK: Can you elaborate on how object spread would help here?

src/targets/registry.ts
BYK: Can do.

BYK:
@tonyo 
>Before reviewers dig into this, could you elaborate on "why"?
>It'll have to stay as subdependency anyway, and up to this point it just worked.

It does work. I was just going over our dependencies, upgrading them etc. and with the new ES features, almost all lodash functionality can be achieved with equally readable and concise JS code. I noticed it being a subdependency after the fact but I think us having less dependencies is a good step forward. I also think some portions are now more understandable as the semantics of `filter`, `empty` etc when it comes to objects are not very clear unless you check the docs.

I'm okay reverting the patch if you disagree though.

tonyo:
It's okay, no need to revert.
I'm further and further away from being (and wanting to be) an authority on this project, so if someone agrees that a thing should be done - so be it.

src/commands/artifacts_cmds/download.ts
jan-auer: Was thinking like this, although looking at it now it's also not much more readable :)

```js
const nameToArtifact = artifacts.reduce((dict, a) => ({...d, [a.filename]: a}), {});
```

PR #214, opened 2021-05-03
ci: Reinstate linting and make it automatic

BYK:


README.md
```diff
@@ -955,9 +955,12 @@ Here is how you can integrate your GitHub project with `craft`:
 
    - The name of the artifacts is very important and needs to be `name: ${{ github.sha }}`. Craft uses this as a unique id to fetch the artifacts.
    - Keep in mind that this action maintains the folder structure and zips everything together. Craft will download the zip and recursively walk it to find all assets.
+
 3. Add `.craft.yml` configuration file to your project
-  - List there all the targets you want to publish to
-  - Configure additional options (changelog management policy, tag prefix, etc.)
+
+- List there all the targets you want to publish to
+- Configure additional options (changelog management policy, tag prefix, etc.)
```
jan-auer, line 962: This seems unintended.

package.json
```diff
@@ -88,15 +86,10 @@
...
+    "precli": "build",
     "cli": "node dist/craft",
     "clean": "rimraf dist coverage",
-    "lint": "run-s lint:prettier lint:eslint",
-    "lint:prettier": "prettier-check 'src/**/*.ts'",
-    "lint:eslint": "eslint src  --ext .ts --format stylish",
-    "fix": "run-s fix:tslint fix:prettier",
-    "fix:prettier": "prettier --write 'src/**/*.ts'",
-    "fix:eslint": "eslint src --ext .ts --fix --format stylish",
+    "lint": "prettier --write .",
```
jan-auer, line 92: There is common terminology with `sentry-javascript`, and even with Makefiles in other repos where:
- `lint` means read-only liniting
- `fix` means writing back changes

I'd prefer to keep this consistent. I know this isn't documented yet, will add this to develop docs when I get the chance.

jan-auer (commented):
Personally speaking, it's unfortunate that it adds these parentheses, but I also understand the reasoning. 

Please merge `master` and rerun the formatter to resolve conflicts.

README.md
BYK: Hmm, let me see how I can fix this.

package.json
BYK: I don't find read-only linting useful as if something can be automatically fixed, then it _should be_ automatically fixed. What am I missing here?

(I get the unspoken convention, just challenging it before I adopt)

jan-auer: There are cases where automatic fixes can't or shouldn't be applied. In most languages, we also lint for conventions, code structure, etc.

BYK: Unless you present me specific cases, I'll go ahead and keep auto-fix in-place until we find a need to switch to read-only mode.

jan-auer: The specific case is 2/3 of all Rust lints that are not auto-fixable. 

Let me turn this around: About two years ago we set the convention of `lint` and `fix` in every repo. Unless there is a hard reason why we **must** diverge from that convention in craft, we should default to keeping it consistent, please.

BYK: >The specific case is 2/3 of all Rust lints that are not auto-fixable. (Edit: I'm not to familiar with latest advancements in JS, but there's gotta be some more advanced linting than just reformatting)

This is JS though and we are just formatting code here with linting :)

>Let me turn this around: About two years ago we set the convention of lint and fix in every repo. Unless there is a hard reason why we must diverge from that convention in craft, we should default to keeping it consistent, please.

This is just a way to block progress for the sake of status quo. If you have cases for JS, I'll see how they affect this, otherwise please do not block simplifications with unclear or unmaterialized concerns.

chadwhitacre: > I'll go ahead and keep auto-fix

Can you spell it `fix` instead of `lint`? I think that would address @jan-auer's concern. For comparison, [`sentry-cli` has `fix` scripts and no `lint` scripts](https://github.com/getsentry/sentry-cli/blob/d309c5a607a310eebd23c1a2e49624e6ea192e21/package.json#L26-L36).

BYK: Done.

chadwhitacre: Does that help, @jan-auer?

jan-auer: That works :)

chadwhitacre (approved):

PR #215, opened 2021-05-03
ref: More readable code for Zeus artifact idiosyncrasies

BYK:


src/artifact_providers/zeus.ts
```diff
@@ -159,6 +159,17 @@ export class ZeusArtifactProvider extends BaseArtifactProvider {
...
+
+    // Sort by name first, for grouping, and then updated at for recency.
+    // This costs us O(N*logN) operations.
+    // After this, we'll do a filter sweep to only pick the _latest_ entry for
+    // the same name. We can do this without an explicit grouping pass thanks to
+    // our initial sort: if the next item doesn't have the same name, we are at
+    // the end of the group and that item is the last, most recent item.
+    // Alternative would be:
+    //   1. Group by name: O(N)
+    //   2. Sort for each name: O(M*N*log(N)) -- where M is the # of groups
```
jan-auer, line 171: That is incorrect, 1 and 2 can be performed with `O(N)` in a single step.

BYK: Updated the comment for clarity. I do not think you can easily group for one thing and then sort for another thing in one `O(N)` pass. It would also require writing custom code again instead of using more-readable helpers.

jan-auer: You don't really need to group and then sort, you can just group and always keep the "maximum" item if there is already one. Once you're done with the list, you're left with your result set. Basically this in **pseudo code**:

```js
const lastArtifacts = {};
for (const artifact of zeusArtifacts) {
  const last = lastArtifacts[artifact.name];
  if (!last || last.updated_at < artifact.updated_at) {
    lastArtifacts[artifact.name] = artifact;
  }
}

return Object.values(lastArtifacts)
  .map(a => this.convertToRemoteArtifact(a));
```

BYK: Ah, gotcha. This kind of looks better than my crazy sort comparator. Let me see how this turns out 🤔

jan-auer (approved):

PR #217, opened 2021-05-04
ref(inquirer): Replace inquirer with prompts

BYK:
Inspired by facebook/create-react-app#10083. `prompts` is a modern and
lighter alternative to `inquirer` and it is almost a drop-in
replacement. This change shaves off 0.7MB from our minified builds.


chadwhitacre (approved):
Did not test but seems reasonable.

PR #220, opened 2021-05-10
feat: Add projectPath config option to GitHub

BYK:
In monorepos like getsentry/sentry-ruby, Craft cannot determine and fetch the remote changelog because it assumes `.craft.yml` is always at git root. This patch adds this optional config and automatically fills it when a `.git` root is found so we can have changelogs in monorepos.

This is to fix https://github.com/getsentry/sentry-ruby/pull/1435#issuecomment-834465195

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,7 +254,7 @@ github:
 ```
 
 If you are using Craft in a monorepo where each project has their own `.craft.yml`
-configuration, you need o set the `github` information with the inclusion of
+configuration, you need to set the `github` information with the inclusion of
 `projectPath`. Although Craft can infer the relative path of your project, it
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -242,14 +242,32 @@ located in the project root.
...
+configuration itself:
 
 ```yaml
 github:
   owner: getsentry
   repo: sentry-javascript
 ```
 
+If you are using Craft in a monorepo where each project has their own `.craft.yml`
+configuration, you need o set the `github` information with the inclusion of
```
chadwhitacre, line 257: o > to

chadwhitacre (changes requested):
Typo in doc.

Not thrilled by lack of tests.

chadwhitacre (approved):

PR #221, opened 2021-05-10
feat: Add config CLI command

BYK:
This command allows us to print the interpreted & validated config in pretty JSON format both for validation and consuming the Craft config.

### DIFF
--- a/src/commands/config.ts
+++ b/src/commands/config.ts
@@ -1,0 +1,14 @@
+import { getConfiguration, getGlobalGithubConfig } from '../config';
+import { formatJson } from '../utils/strings';
+
+export const command = ['config'];
+export const description = 'List defined targets as JSON array';
+
+export async function handler(): Promise<void> {
+  const github = await getGlobalGithubConfig();
+  const config = {
+    ...getConfiguration(),
+    github,
+  };
+  console.log(formatJson(config));
+}
--- a/.craft.yml
+++ b/.craft.yml
@@ -1,7 +1,4 @@
 minVersion: '0.21.0'
-github:
-  owner: getsentry
-  repo: craft
 changelogPolicy: auto
 requireNames:
   - /^sentry-craft.*\.tgz$/
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -20,7 +20,7 @@ steps:
   # Smoke tests
   - name: 'us.gcr.io/$PROJECT_ID/craft:$COMMIT_SHA'
     args:
-      - '--help'
+      - 'config'
     timeout: 60s
   - name: 'gcr.io/cloud-builders/docker'
     secretEnv: ['DOCKER_PASSWORD']
--- a/src/index.ts
+++ b/src/index.ts
@@ -13,6 +13,7 @@ import { getPackageVersion } from './utils/version';
 import * as prepare from './commands/prepare';
 import * as publish from './commands/publish';
 import * as targets from './commands/targets';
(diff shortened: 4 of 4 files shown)


src/commands/prepare.ts
```diff
@@ -5,7 +5,6 @@ import simpleGit, { SimpleGit } from 'simple-git';
 import { Arguments, Argv, CommandBuilder } from 'yargs';
 
 import {
-  checkMinimalConfigVersion,
```
chadwhitacre, line 8: You sure you want to do this? Performing this check only under the new `config` command significantly weakens the protection against version mismatch.

src/commands/config.ts
```diff
@@ -0,0 +1,14 @@
+import { getConfiguration, getGlobalGithubConfig } from '../config';
+import { formatJson } from '../utils/strings';
+
+export const command = ['config'];
+export const description = 'List defined targets as JSON array';
```
chadwhitacre, line 5: This is a lie. copy/pasta!

chadwhitacre (changes requested):

chadwhitacre:
> both for validation and consuming the Craft config.

Say more? What are the use cases for this? Are we validating our own validation? Who wants to consume Craft's validated output besides Craft?

src/commands/prepare.ts
BYK: I'm not moving this under the new command, I'm moving it to the core `getConfiguration()` method which needs o be called by _every_ command that does anything useful. So we are increasing he protection :)

src/commands/config.ts
BYK: Lol, good catch. Fixing!

BYK:
>Say more? What are the use cases for this? Are we validating our own validation? Who wants to consume Craft's validated output besides Craft?

I'm hoping to use this for testing Craft config changes in other repos to validate we are not breaking things as we did earlier.

src/commands/config.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const command = ['config'];
+export const description =
+  'Print the parsed, processed, and validated Craft config for the current project in pretty-JSON.';
+
+export async function handler(): Promise<void> {
+  const github = await getGlobalGithubConfig();
+  const config = {
+    ...getConfiguration(),
+    github,
```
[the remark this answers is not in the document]
BYK: >As written, this takes the inferred info over an explicit config. Is that what we want? Can we abstract out this logic if it's used in multiple places so that we are consistent?

The "inferred" one is still looking at the config first and then tries to infer. This is already abstracted away in #218. I'd love to bake this into `getConfiguration()` itself but one is sync and one is async so it may require a larger refactor. 	I'll look into this in a follow-up patch.

chadwhitacre (approved):

src/commands/config.ts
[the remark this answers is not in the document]
chadwhitacre: Confirmed that config overrides inferred:

```
$ ./dist/craft config | jq .github.repo
"cruft"
$ gd
```
```diff
diff --git a/.craft.yml b/.craft.yml
index b2be798..01cad7e 100644
--- a/.craft.yml
+++ b/.craft.yml
@@ -1,7 +1,7 @@
 minVersion: '0.21.0'
 github:
   owner: getsentry
-  repo: craft
+  repo: cruft
 changelogPolicy: auto
 requireNames:
   - /^sentry-craft.*\.tgz$/
```

Issue #180, opened 2021-02-11
Publish a version of Craft GitHub action


BYK:
We have the craft github action but it's always master, also the action is not on GH marketplace. Acceptance:

Craft Action Versioned
Craft Action on GH Marketplace

/cc @chadwhitacre

PR #224, opened 2021-05-11
ref: Unify global flags

BYK:
This PR unifies how we process the global flags for the app, namely: `dry-run`, `no-input`, and `log-level`. These are now processed the same way and can be set via `CRAFT_<CAPITAL_CASE_NAME>` env variables, thanks to some `yargs` magic.

This same `yargs` magic allows us to fill in _any_ CLI argument via the same method. This is great for the Craft action (#180) as we can now easily set all flags/arguments via ENV variables, saving us monstrosities such as https://github.com/getsentry/craft/blob/897266f8b4f76cf4447dd603928ffc80ffb11160/action.yml#L23-L24


BYK:
This probably deserves a README update.

chadwhitacre (approved):
This is beautiful. Good work! 👏

PR #225, opened 2021-05-11
upgrade(unzipper): Upgrade unzipper to latest version

BYK:
The version we are using is quite old and there are numerous bug fixes including security patches. This version also comes with TypeScript types.


src/utils/system.ts
```diff
@@ -358,11 +358,9 @@ export async function extractZipArchive(
   return new Promise<void>((resolve, reject) => {
     try {
       fs.createReadStream(filePath)
-        .pipe(unzipper.Extract({ strip: 1, path: dir }))
+        .pipe(unzipper.Extract({ path: dir, concurrency: 2 }))
         .on('error', reject)
-        .on('finish', () => {
-          setTimeout(resolve, 100);
-        });
+        .on('close', resolve);
```
chadwhitacre, line 363: Do we have test coverage for this change? If not can you paste some console output to prove that it works? 😁

BYK: Just did a test release to https://github.com/bitsandfoxes/testrepo

This is also on the docs for `unzipper`: https://github.com/ZJONSSON/node-unzipper/#extract-to-a-directory

chadwhitacre: wfm thx

chadwhitacre (approved):

PR #226, opened 2021-05-12
fix(github-artifacts): Allow multiple artifacts on the SHA, use latest

BYK:
For some reason we did not allow more than one artifact tied to a commit SHA (which may happen with branches etc). Now we allow this and pick the most recent one.

### DIFF
--- a/src/artifact_providers/github.ts
+++ b/src/artifact_providers/github.ts
@@ -97,7 +97,7 @@ export class GithubArtifactProvider extends BaseArtifactProvider {
 
     // We need to find the most recent archive where name matches the revision.
     const foundArtifact = allArtifacts.reduce((result, artifact) =>
-      artifact.name === revision && result?.created_at < artifact.created_at
+      artifact.name === revision && result.created_at < artifact.created_at
         ? artifact
         : result
     );
--- a/src/artifact_providers/github.ts
+++ b/src/artifact_providers/github.ts
@@ -95,7 +95,7 @@ export class GithubArtifactProvider extends BaseArtifactProvider {
 
     const allArtifacts = artifactResponse.artifacts;
 
-    // We need to find the most recent archive where name maches the revision
+    // We need to find the most recent archive where name matches the revision.
     const foundArtifact = allArtifacts.reduce((result, artifact) =>
       artifact.name === revision && result?.created_at < artifact.created_at
         ? artifact
--- a/src/artifact_providers/github.ts
+++ b/src/artifact_providers/github.ts
@@ -95,20 +95,15 @@ export class GithubArtifactProvider extends BaseArtifactProvider {
 
     const allArtifacts = artifactResponse.artifacts;
 
-    // We need to find the archive where name maches the revision
-    const foundArtifacts = allArtifacts.filter(
-      artifact => artifact.name === revision
+    // We need to find the most recent archive where name maches the revision
+    const foundArtifact = allArtifacts.reduce((result, artifact) =>
+      artifact.name === revision && result?.created_at < artifact.created_at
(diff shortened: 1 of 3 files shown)


BYK:
Test run: https://github.com/getsentry/release-tester/releases/0.11.0

src/artifact_providers/zeus.ts
```diff
@@ -134,14 +134,12 @@ export class ZeusArtifactProvider extends BaseArtifactProvider {
...
-      // with that commit. (For all other artifact providers, not knowing about
-      // the commit and having no files associated with the commit are one and
-      // the same.) In the former case (known commit, no files), zeus will
-      // return an empty list, whereas in the latter case (unknown commit), it
-      // will error. This error message check and the length check below are
-      // here to disambiguate those two situations.
+      // Zeus may know about a commit (from its role as a status provider) but
+      // not have any files associated with that commit. In the former case
+      // (known commit, no files), Zeus will return an empty list, whereas in
+      // the latter case(unknown commit), it will error.
```
chadwhitacre, line 140:
```suggestion
      // the latter case (unknown commit), it will error.
```

src/artifact_providers/zeus.ts
```diff
@@ -154,12 +152,8 @@ export class ZeusArtifactProvider extends BaseArtifactProvider {
...
     }
 
-    // similarly, zeus is the only artifact provider which will store multiple
-    // copies of the same file for a given revision, so to mimic the behavior of
-    // the other providers (which overwrite pre-existing, identically-named
-    // files within the same commit), for each filename, take the one with the
-    // most recent update time
-
+    // Zeus stores multiple copies of the same file for a given revision,
+    // take the one with the most recent update time
```
chadwhitacre, line 156:
```suggestion
    // take the one with the most recent update time.
```

src/artifact_providers/github.ts
```diff
@@ -95,20 +95,15 @@ export class GithubArtifactProvider extends BaseArtifactProvider {
 
     const allArtifacts = artifactResponse.artifacts;
 
-    // We need to find the archive where name maches the revision
-    const foundArtifacts = allArtifacts.filter(
-      artifact => artifact.name === revision
+    // We need to find the most recent archive where name maches the revision
```
chadwhitacre, line 98:
```suggestion
    // We need to find the most recent archive where name matches the revision.
```

src/artifact_providers/github.ts
```diff
@@ -95,20 +95,15 @@ export class GithubArtifactProvider extends BaseArtifactProvider {
...
     const allArtifacts = artifactResponse.artifacts;
 
-    // We need to find the archive where name maches the revision
-    const foundArtifacts = allArtifacts.filter(
-      artifact => artifact.name === revision
+    // We need to find the most recent archive where name matches the revision.
+    const foundArtifact = allArtifacts.reduce((result, artifact) =>
+      artifact.name === revision && result?.created_at < artifact.created_at
+        ? artifact
+        : result
```
chadwhitacre, line 102: I think this selects `result` if `result` has no `created_at`. Is that what we want? 🤔

BYK: Good catch, that said `result` will always have a `created_at`. This code says "if result itself is missing" which essentially boils down to the same conclusion you came to. And no, that's not something we want. I was saved here by how `reduce` works: if you don't provide an explicit starting value, it uses the first item in the array so `result` is never undefined.

Will fix this.

BYK: Fixed.

chadwhitacre (approved):

PR #228, opened 2021-05-12
fix(debug): Fix mangled provider names on debug/info logging

BYK:
Since the single-file builds, we started to see log lines such as `Got status "success" from status provider: Vae`. This is because we minify the output, which changes function/class names for optimization and these lines rely on this `constructor.name` property. This patch adds explicit names to avoid using the `--keep-names` option in `esbuild` as that adds ~300kiB to the final file size just for this minor issue. This patch however is weightless*.


chadwhitacre (approved):
Love the fit and finish!

PR #229, opened 2021-05-12
fix(cli): Allow extra CRAFT_* env variables

BYK:
Since we used 'strict mode' in `yargs` and then enabled mapping of _any_ `CRAFT_*` env variables to CLI args, Craft now complains when it sees env variables like `CRAFT_GCS_STORE_CREDS_JSON` which are only directly used from specific targets and not available through the CLI. In an ideal world, either these wouldn't start with the `CRAFT_` prefix, or would be available as CLI args. Since we are not in that world, we are doing a compromise: converting one simple thing we acn into a CLI arg, and then relaxing the strictness by only being strict about the commands but not options. Since this feature (`strictCommands()` was added in a later version, we also upgrade `yargs` to `15.4.1` which is the canonical version being used by other tools in the repo.

Fix up to #224.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "ts-jest": "^26.5.5",
     "typescript": "^4.1.3",
     "unzipper": "0.10.11",
-    "yargs": "15.1.0"
+    "yargs": "15.4.1"
   },
   "scripts": {
     "build:fat": "yarn run compile-config-schema && tsc -p tsconfig.build.json",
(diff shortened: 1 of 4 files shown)


BYK:
I did manually verify that this build fixes the issue by using `CRAFT_YOLO=yolo` which fails without the patch and works without issues after the patch.

chadwhitacre (approved):

PR #235, opened 2021-05-16
fix(cli): Fix global flag parsing interference

BYK:
This fixes the release failures such as https://github.com/getsentry/sentry/runs/2591288014?check_suite_focus=true.
The culprit is #224 where we switched the global `--no-input` and
`--dry-run` flags from boolean to string args.
This now causes  `craft prepare --no-input "21.5.0"` to be interpreted
as `craft prepare --no-input=21.5.0`.

This patch fixes the issue.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -22,6 +22,46 @@ function printVersion(): void {
   }
 }
 
+const GLOBAL_BOOLEAN_FLAGS = {
+  'no-input': {
+    coerce: envToBool,
+    default: isCI,
+    describe: 'Suppresses all user prompts',
+    global: true,
+  },
+  'dry-run': {
+    coerce: envToBool,
+    // TODO(byk): Deprecate this in favor of CRAFT_DRY_RUN
+    default: process.env.DRY_RUN,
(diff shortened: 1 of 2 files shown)


HazAT (approved):

src/index.ts
```diff
@@ -22,6 +22,46 @@ function printVersion(): void {
...
+};
+
+/**
+ * This function is to pre-process and fix one of yargs' shortcomings:
+ * We want to use some flags as booleans: just their existence on the CLI should
+ * set them to true. That said since we also allow setting them through
+ * environment variables, we need to parse many string values that would set
+ * them to false. Moreover, we want to be able to override already-set env
+ * variables with the `--flag=no` kind of notation (using the `=` symbol) but
+ * not via the positional argument notation (`--flag no`). The only way to do
```
chadwhitacre, line 48: Feels really special-casey. :-/

BYK: Yeah, it is. Right now it is not easy to replace `yargs` with anything else but I'll be on the look.

PR #236, opened 2021-05-17
fix(logging): Fix scoped loggers not respecting log level

BYK:
After the `consola` upgrade in #224, we stopped respecting the log level that is set further in execution in scoped logger which were created eagerly at module initialization time and not inheriting the log level that was changed afterwards. This patch moves these module-level loggers into base class constructors and standardizes them. This both fixes the issue and gives a common logging style for all artifact providers and targets.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.22.2
+
+- fix(logging): Fix scoped loggers not respecting log level (#236)
+
 ## 0.22.1
 
 - fix(cli): Fix global flag parsing interference (#235)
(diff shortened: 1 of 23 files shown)


chadwhitacre (commented):
> a common logging style for all artifact providers and targets.

Why only artifact providers and targets? Why not also status providers, stores, utils, commands, config, and index?

BYK:
>Why only artifact providers and targets? Why not also status providers, stores, utils, commands, config, and index?

I just refactored where this already existed. Not interested in adding more ATM as I mostly want to fix the issue we have.

chadwhitacre:
> where this already existed.

Where what already existed? The bug with not respecting log level? Why do we only have that problem in targets and artifact providers and not elsewhere?

chadwhitacre (approved):

PR #237, opened 2021-05-17
fix(github-artifacts): Fix incorrect artifact resolution

### LINKED ISSUE #234: Release with GHA artifact missing/wrong file
Trying to deploy `sentry-unity` but the list of available artifacts is missing the package.

The build in question is: https://github.com/getsentry/sentry-unity/actions/runs/843357161 which contains an artifact named after the commit sha `b7e538648f85e65a44a7f6a4c79432ca12ce2576`.

Craft finds the build but shows only `results.xml`.
I downloaded the artifact from GHA and found it contains: `package-release.zip`, the expected file.

### Environment

```
craft --version
0.22.0
```

Configuration: https://github.com/getsentry/sentry-unity/blob/b7e538648f85e65a44a7f6a4c79432ca12ce2576/.craft.yml

### Steps to Reproduce

```
➜  sentry-unity git:(main) ✗ craft publish 0.0.14
ℹ Publishing version: "0.0.14"                                                                                                                                                                                              09:42:39
ℹ Revision b7e538648f85e65a44a7f6a4c79432ca12ce2576 has been built successfully.                                                                                                                                            09:42:40
ℹ Fetching Github artifacts for getsentry/sentry-unity, revision b7e538648f85e65a44a7f6a4c79432ca12ce2576                                                                                        [artifact-provider/github] 09:42:40
ℹ Finished downloading.                                                                                                                                                                          [artifact-provider/github] 09:42:41
ℹ                                                                                                                                                                                                                           09:42:41
ℹ Available artifacts:                                                                                                                                                                                                      09:42:41
┌─────────────┬─────────┬─────────┬─────────────┐
│ File Name   │ Size    │ Updated │ ContentType │
├─────────────┼─────────┼─────────┼─────────────┤
│ results.xml │ 18.3 kB │         │             │
└─────────────┴─────────┴─────────┴─────────────┘

ℹ Looking for publish state file for 0.0.14...                                                                                                                                                                              09:42:41
ℹ Publishing to targets:                                                                                                                                                                                                    09:42:41
ℹ   - upm                                                                                                                                                                                                                   09:42:41
  - github
  - registry
```

BYK:
Fixes #234. This is a follow up to #226. Although @chadwhitacre pointed out a logic issue in an earlier revision, we have failed to see that we were _always_ returning an artifact even if we don't have a proper name match as `reduce` without an initial value, starts with the first element of the array. This patch fixes the issue.

### DIFF
--- a/src/artifact_providers/__tests__/github.test.ts
+++ b/src/artifact_providers/__tests__/github.test.ts
@@ -1,0 +1,211 @@
+jest.mock('../../utils/githubApi.ts');
+import { getGithubClient } from '../../utils/githubApi';
+import { GithubArtifactProvider, ArtifactItem } from '../github';
+
+class TestGithubArtifactProvider extends GithubArtifactProvider {
+  public testGetRevisionArtifact(revision: string): Promise<ArtifactItem> {
+    return this.getRevisionArtifact(revision);
+  }
+}
+
+describe('GitHub Artifact Provider', () => {
+  let githubArtifactProvider: TestGithubArtifactProvider;
+  let mockClient: { request: jest.Mock };
+
+  beforeEach(() => {
+    jest.resetAllMocks();
+    mockClient = {
+      request: jest.fn(),
+    };
+    (getGithubClient as jest.MockedFunction<
+      typeof getGithubClient
+      // @ts-ignore we only need to mock a subset
+    >).mockReturnValueOnce(mockClient);
+
+    githubArtifactProvider = new TestGithubArtifactProvider({
+      repoOwner: 'getsentry',
+      repoName: 'craft',
+    });
+  });
+
+  describe('listArtifactsForRevision', () => {
+    test('it should get the artifact with the revision name', async () => {
+      mockClient.request.mockResolvedValueOnce({
+        status: 200,
+        data: {
+          total_count: 2,
+          artifacts: [
+            {
+              id: 60233710,
+              node_id: 'MDg6QXJ0aWZhY3Q2MDIzMzcxMA==',
+              name: '1b843f2cbb20fdda99ef749e29e75e43e6e43b38',
+              size_in_bytes: 6511029,
+              url:
+                'https://api.github.com/repos/getsentry/craft/actions/artifacts/60233710',
+              archive_download_url:
+                'https://api.github.com/repos/getsentry/craft/actions/artifacts/60233710/zip',
+              expired: false,
+              created_at: '2021-05-12T21:50:35Z',
+              updated_at: '2021-05-12T21:50:38Z',
+              expires_at: '2021-08-10T21:50:31Z',
+            },
+            {
+              id: 60232691,
+              node_id: 'MDg6QXJ0aWZhY3Q2MDIzMjY5MQ==',
+              name: 'e4bcfe450e0460ec5f20b20868664171effef6f9',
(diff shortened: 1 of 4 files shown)


src/artifact_providers/github.ts
```diff
@@ -94,13 +94,16 @@ export class GithubArtifactProvider extends BaseArtifactProvider {
...
-      artifact.name === revision && result.created_at < artifact.created_at
-        ? artifact
-        : result
+    const foundArtifact = artifacts.reduce(
+      (result, artifact) =>
+        artifact.name === revision &&
+        (!result || result.created_at < artifact.created_at)
+          ? artifact
+          : result,
+      null as ArtifactItem | null
```
chadwhitacre, line 106: Why use reduce here instead of a for loop? A for loop would be easier to reason about imo.

chadwhitacre (changes requested):
Also, I'm gonna play hardball with tests on this one. Complicated logic like this should have tests around it. We shouldn't rely solely on code review.

src/artifact_providers/github.ts
chadwhitacre: https://kentcdodds.com/blog/array-reduce-vs-chaining-vs-for-loop 🤔

chadwhitacre: (tl;dr Kent prefers chaining. 😅)

BYK: Eh, I'll go with a loop, sure :)

chadwhitacre: Also, just below in a comment we say that we assume that artifacts are listed in date descending order on this endpoint. If we can assume that then we don't need to compare dates, do we? We can just take the first one we find? Something like this?

```javascript
    for (let artifact of artifacts) {
      if (artifact.name === revision) {
        return artifact
      }   
    );
```

src/artifact_providers/github.ts
```diff
@@ -124,7 +125,7 @@ export class GithubArtifactProvider extends BaseArtifactProvider {
       // the artifacts are listed in descending date order on this endpoint.
       // There is no public documentation on this but the observed data and
       // common-sense logic suggests that this is a reasonably safe assumption.
```
chadwhitacre, line 127: De-dupe this comment with the above. As it stands its now redundant.

src/artifact_providers/__tests__/github.test.ts
```diff
@@ -0,0 +1,211 @@
+jest.mock('../../utils/githubApi.ts');
+import { getGithubClient } from '../../utils/githubApi';
+import { GithubArtifactProvider, ArtifactItem } from '../github';
+
+class TestGithubArtifactProvider extends GithubArtifactProvider {
+  public testGetRevisionArtifact(revision: string): Promise<ArtifactItem> {
+    return this.getRevisionArtifact(revision);
+  }
+}
```
chadwhitacre, line 9: What's the value here vs. using `GithubArtifactProvider` directly?

chadwhitacre (approved):

src/artifact_providers/__tests__/github.test.ts
BYK: It is not a value, it is a necessity if we want to keep the `getRevisionArtifact` method private on the class. I switched it to `protected` so it is exposed to child classes but not on the instances.

Issue #196, opened 2021-04-02
Replace request with a different library


BYK:
The `request` library we use is deprecated. Here are the alternatives: https://github.com/request/request/issues/3143

PR #238, opened 2021-05-17
ref(aws): Remove @aws-sdk/client-ec2 dependency

BYK:
The `@aws-sdk/client-ec2` dependency was only used to list all
available EC2 regions but it added more than 1 MB to the final
binary size.

This PR replaces that with a manual http request, relying on the
`request` library's [magic `aws` signing
feature](https://github.com/request/request#requestoptions-callback).

Even when we remove `request` (see #196), we can still do this quite
easily by directly using the [aws4](https://yarnpkg.com/package/aws4)
library.


BYK:
I manually tested this but would love to get confirmation from @iker-barriocanal. The bundle size is ~5.1MB before the patch and ~3.3MB after the patch.

iker-barriocanal (approved):

PR #241, opened 2021-05-19
feat(registry): Batch all registry targets

### LINKED ISSUE #48: registry target: batch updates if there are multiple packages to release
Case in question:
https://github.com/getsentry/sentry-dotnet/blob/master/.craft.yml#L11-L30

Ideally, there should be a way to batch those changes into one push to release-registry.

BYK:
Fixes #48. This PR automatically batches all registry operations in a
concurrency-safe manner. It checks out a single, canonical copy of the
registry, adds all the commits there and pushes all of them at once
instead of doing piecemeal check-out and pushes.

The goals when preparing the patch were as follows:
- Do not require any existing config to change
- Do not require any other target to support batch-mode
- Do not add any RegistryTarget-specific code outside of that class
- Keep RegistryTarget class usable w/o a batch-helper or factory pattern


BYK:
The locking logic here is broken so do not merge and ignore that part for now when reviewing.

src/targets/registry.ts
```diff
@@ -408,20 +427,67 @@ export class RegistryTarget extends BaseTarget {
...
+      localRepo = RegistryTarget.localRepo;
+    } else {
+      // XXX(BYK): This should NEVER happen
+      throw new Error(
+        `Local registry missing, it should have been cloned at this stage!`
+      );
+    }
+    this.commitVersionToRegistry(localRepo, version, revision);
+
+    if (this.isLastToPublish()) {
```
jan-auer, line 468: Please also add documentation around execution order. It looks like the _last_ target triggers the actual publish. While it makes sense, it might not necessarily be what a user expects. (knowing how it's implemented it's the only sensible approach though)

src/targets/registry.ts
```diff
@@ -81,6 +109,7 @@ export class RegistryTarget extends BaseTarget {
     this.github = getGithubClient();
     this.githubRepo = githubRepo;
     this.registryConfig = this.getRegistryConfig();
+    RegistryTarget.instances.add(this);
```
jan-auer, line 112: This singleton and the constructor side effect seem surprising from the outside. Could you lift this out of the registry target into generic code? Here's a suggestion:

- `BaseTarget` interface class gets something like `supportsBatching(): bool` and `addConfig()`
- The driver asks the target (class) whether it supports batching, and adds configs rather than publishing right away
- At some point defined by the driver -- probably after the last target of that kind -- it issues the `publish()` call

jan-auer (commented):
From user perspective, I like this approach very much. Rather than having the user think about whether they should create one entry or multiple entries, the target itself has support for batched operation. This can also be interesting for other targets.

To confirm: Does this still work with the lockfile? Effectively, `registry` now operates as a single target, but you'd still want all of the original target configs marked in the lockfile for partial releases.

src/targets/registry.ts
BYK: This is the first approach I thought about and I decided against it as it complicates code outside of registry too much and without a good reason. It is unclear if any other target needs or can use batching so to restate my goals in the PR description:

1. Do not require any existing config to change
2. Do not require any other target to support batch-mode
3. **Do not add any RegistryTarget-specific code outside of that class**
4. Keep RegistryTarget class usable w/o a batch-helper or factory pattern

This approach would violate items 2 and 3 if the logic is moved to the base class. I'd only move the logic to the base class if at least one or two more targets would need/benefit from batch mode.

BYK:
>To confirm: Does this still work with the lockfile? Effectively, registry now operates as a single target, but you'd still want all of the original target configs marked in the lockfile for partial releases. Edit: Just noticed your comment.

I think we are talking about two different things here. The "locking mechanism" I mentioned above was about limiting concurrency but I think you are referring to the state file, introduced in #192. I think I need to look deeper into the state file as we should mark all registry targets done at once in the state file, which shouldn't be the case here, great catch.

src/targets/registry.ts
jan-auer: > It is unclear if any other target needs or can use batching

This is a good point. I spent some time thinking about it and came to these conclusions:

1. The registry target also does not need batching, strictly speaking. The inconsistent releases mentioned in the original issue are actually solved by resuming the failed release using the lockfile, so we already have a solution for that. It's a nice-to-have addition that we only create a single commit. That said, it's extremely sensible to only do a single commit, and run all of this in batch.
2. For the same reason, `upm` and `brew` would benefit from batching. In practice, we don't have a use for this, however.
3. All other targets we repeat require individual uploads or actions, so there's no point in batching. 

When writing my comment originally, I assumed that we want to make this a generic concept: When a target is repeated, its operation is batched. If we want to keep this specific to the registry target, I would rather propose to change the target config to accept an explicit list of specs and update the 3 SDKs where we need that. This gets rid of implicit behavior.

BYK: >It's a nice-to-have addition that we only create a single commit. That said, it's extremely sensible to only do a single commit, and run all of this in batch.

Mind you, this PR makes a single push but still has multiple commits for clarity and ease of implementation.

>For the same reason, upm and brew would benefit from batching. In practice, we don't have a use for this, however.

I'd be interested to hear more details, especially for `upm` as I cannot understand how it can benefit from batching.

>When writing my comment originally, I assumed that we want to make this a generic concept: When a target is repeated, its operation is batched.

This is how I started and eventually decided that this seems like a special-case for the registry target rather than being a general concept applicable to other targets.

src/targets/registry.ts
```diff
@@ -408,20 +445,76 @@ export class RegistryTarget extends BaseTarget {
...
-      directory =>
-        this.pushVersionToRegistry(directory, remote, version, revision),
-      true,
-      'craft-release-registry-'
+    const startBatch = RegistryTarget.lock.isLocked();
+    const result = RegistryTarget.lock.runExclusive(() =>
+      this.doPublish(version, revision)
     );
-    this.logger.info('Release registry updated');
+    if (startBatch) {
```
BYK, line 457: Verified this logic with https://gist.github.com/BYK/b7f26582d0a2674f6785b5c80c38de43

src/targets/registry.ts
```diff
@@ -59,18 +61,52 @@ export interface RegistryConfig {
...
+
+interface ArtifactData {
+  url?: string;
+  checksums?: {
+    [key: string]: string;
+  };
+}
+
 /**
  * Target responsible for publishing static assets to GitHub pages
```
chadwhitacre, line 77: This comment is a lie, right?

chadwhitacre:
Probably should update the README?

https://github.com/getsentry/craft#sentry-release-registry-registry

src/targets/registry.ts
```diff
@@ -59,18 +61,52 @@ export interface RegistryConfig {
...
  */
 export class RegistryTarget extends BaseTarget {
+  /** A set of all created registry instances to be able to batch them */
+  private static instances: Set<RegistryTarget> = new Set();
+  // A promise-based lock to ensure non-parallel publishing when batching
```
chadwhitacre, line 82:
```suggestion
  // A promise-based lock to serialize publishing when batching
```

src/targets/registry.ts
```diff
@@ -265,18 +312,14 @@ export class RegistryTarget extends BaseTarget {
     this.logger.info(
       'Adding extra data (checksums, download links) for available artifacts...'
     );
-    const files: { [key: string]: any } = {};
 
+    const files: { [key: string]: any } = {};
     await mapLimit(artifacts, MAX_DOWNLOAD_CONCURRENCY, async artifact => {
       const fileData = await this.getArtifactData(artifact, version, revision);
-      if (Object.keys(fileData).length > 0) {
```
chadwhitacre, line 272: This protection is no longer necessary because the new condition above (`this.registryConfig.checksums.length === 0`) is equivalent, yes?

src/targets/registry.ts
```diff
@@ -59,18 +61,52 @@ export interface RegistryConfig {
...
   public readonly registryConfig: RegistryConfig;
   /** Github client */
   public readonly github: Github;
   /** Github repo configuration */
   public readonly githubRepo: GithubGlobalConfig;
+  private published = false;
+
+  public isPublished(): boolean {
+    return this.published;
+  }
```
chadwhitacre, line 99: Why do we need this? Who is going to consume this outside of this class? We already access `this.published` directly below.

chadwhitacre: Hrm, nevermind. In `isLastToPublish` we're dealing with sibling instances, not ourself.

src/targets/registry.ts
chadwhitacre: But in the gist I see `!lock.isLocked()` and here I see `lock.isLocked()` ... the opposite.

chadwhitacre: Since we have test code can we bring it into CI here?

src/targets/registry.ts
```diff
@@ -408,20 +445,76 @@ export class RegistryTarget extends BaseTarget {
...
-
-    await withTempDir(
-      directory =>
-        this.pushVersionToRegistry(directory, remote, version, revision),
-      true,
-      'craft-release-registry-'
+    const startBatch = RegistryTarget.lock.isLocked();
+    const result = RegistryTarget.lock.runExclusive(() =>
+      this.doPublish(version, revision)
     );
```
chadwhitacre, line 456: Suggest flipping these two lines and possibly otherwise reformatting to make it clearer that we always `doPublish` for ourself, and only kick off publishing for everyone else if we're the first in the batch.


```suggestion
    const result = RegistryTarget.lock.runExclusive(() =>
      this.doPublish(version, revision)
    );
    const startBatch = RegistryTarget.lock.isLocked();
```

src/targets/registry.ts
```diff
@@ -408,20 +445,76 @@ export class RegistryTarget extends BaseTarget {
...
+
+    if (this.isLastToPublish()) {
+      // Push!
+      if (!isDryRun()) {
+        this.logger.info(`Pushing the changes...`);
+        // Ensure we are still up to date with upstream
+        await withRetry(() =>
+          localRepo.git
+            .pull('origin', 'master', ['--rebase'])
+            .push('origin', 'master')
```
chadwhitacre, line 502: Rebasing minimizes but doesn't eliminate the race condition with other releases that might be going on. Probably low enough traffic that this is fine for now, might matter some day.

chadwhitacre:
From https://github.com/getsentry/craft/issues/48#issuecomment-790884404 ...

> One downside of the current approach is that we end up releasing the registry with unmatching versions for monorepos.

How does this happen? What does the screenshot mean? Is it the 1.0.2 vs. 1.0.3 that matters? Where is that screenshot from? How does this PR address @bruno-garcia's use case?

src/targets/registry.ts
BYK: Lol, let me fix that while I'm at this.

src/targets/registry.ts
BYK: Yup 👍🏻

src/targets/registry.ts
```diff
@@ -371,27 +397,38 @@ export class RegistryTarget extends BaseTarget {
...
+    );
+    await git.clone(remote.getRemoteStringWithAuth(), directory, [
+      '--filter=tree:0',
+      '--single-branch',
+    ]);
+    return git;
   }
 
   /**
    * Pushes an archive with static HTML web assets to the configured branch
```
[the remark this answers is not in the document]
BYK: Big time.

src/targets/registry.ts
BYK: Cannot do it the way you suggested as as soon as we run `RegistryTarget.lock.runExclusive`, `RegistryTarget.lock.isLocked()` will always return `true` 🙂

src/targets/registry.ts
BYK: Yup you caught a bug! I'll try to add tests here.

src/targets/registry.ts
BYK: Note the `withRetry` that's wrapping this so we try up to 3 times. See #232 for more context.

src/targets/registry.ts
chadwhitacre: (Discussing race condition in private Slack ...)

BYK: This is no longer an issue with the eager `this.published = true` command.

PR #244, opened 2021-05-21
ref(registry): Refactor RegistryTarget

BYK:
This PR salvages the non-controversial parts of #241:

- Better, more modular code organization
- Better/more typing
- Do a single clone/checkout for all registry

### DIFF
--- a/src/targets/registry.ts
+++ b/src/targets/registry.ts
@@ -1,6 +1,7 @@
 import { mapLimit } from 'async';
 import * as Github from '@octokit/rest';
-import simpleGit from 'simple-git';
+import rimraf from 'rimraf';
+import simpleGit, { SimpleGit } from 'simple-git';
 import * as path from 'path';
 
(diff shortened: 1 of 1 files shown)


BYK:
>Do a single clone/checkout for all registry

This part is not really concurrency/thread safe but we don't run publish actions concurrently yet anyway.

src/targets/registry.ts
```diff
@@ -59,12 +60,26 @@ export interface RegistryConfig {
...
+interface ArtifactData {
+  url?: string;
+  checksums?: {
+    [key: string]: string;
+  };
+}
+
 /**
- * Target responsible for publishing static assets to GitHub pages
+ * Target responsible for publishing to Sentry's release registry: https://github.com/getsentry/sentry-release-registry/
```
chadwhitacre, line 76: 👍

chadwhitacre (approved):

PR #245, opened 2021-05-26
feat(publish): Ability to merge to non-default

### LINKED ISSUE #115: Pull after merging the release branch
After successfully publishing and merging the release branch, you still have to manually call `git pull` to get your local branch up to speed. Craft could do that automatically.

The only place where this probably shouldn't happen, is the GitHub Action.

### LINKED ISSUE #239: Preparing a release not from `master` branch -> Publish Action
When using our Prepare Action, one is able to choose a `branch` from which the release should be created from.

image

It correctly creates an issue in our publish repo https://github.com/getsentry/publish/issues/325 but the problem is it tries to merge it back to master instead of the branch you chose to prepare the release.

Solving this would it also make it possible, I think, to do beta release with our prepare action.

BYK:
This fixes #239 and fixes #115. This fix requires us to switch
from a purely GitHub API driven publish flow to a purely local
Git driven publish flow as it is not possible to compute the
nearest merge base via any GitHub API. We rely on `git log` and
some parsing to be able to determine this.

The patch also effectively reverts #220 as there is now no need
for that (we are operating on the repo directly).

Test run: https://github.com/getsentry/release-tester/commits/releases/0.4.x
Also see no merges to main: https://github.com/getsentry/release-tester/commits/main

Depends on getsentry/publish#334

TODO: re-read config once we switch branches.


tonyo (approved):

src/config.ts
```diff
@@ -261,18 +261,11 @@ export async function getGlobalGithubConfig(
...
-        path.parse(path.relative(gitRoot, configDir))
-      );
       repoGithubConfig = {
         owner: remoteUrl.owner,
         repo: remoteUrl.name,
-        projectPath,
       };
     }
-  } else if (!repoGithubConfig.projectPath) {
-    repoGithubConfig.projectPath = '.';
```
chadwhitacre, line 275: Still finding some other mentions of `projectPath`. Remove as well?

```
$ ag projectPath
README.md
287:`projectPath`. Although Craft can infer the relative path of your project, it
294:  projectPath: sentry-rails

CHANGELOG.md
23:- feat: Add projectPath config option to GitHub (#220)

src/schemas/project_config.ts
29:  projectPath?: string;

src/schemas/projectConfig.schema.ts
23:        projectPath: {
```

src/commands/prepare.ts
```diff
@@ -527,7 +511,7 @@ export async function prepareMain(argv: PrepareOptions): Promise<any> {
 
   if (argv.publish) {
     logger.success(`Release branch "${branchName}" has been pushed.`);
-    await execPublish(newVersion);
+    await execPublish(argv.remote, newVersion);
```
chadwhitacre, line 514: Why does `argv.remote` already exist here? Is it guaranteed to be defined and non-null?

src/utils/githubApi.ts
```diff
@@ -202,186 +167,3 @@ export async function getFile(
...
-      } else {
-        if (maxRetries > 0 && retryNum >= maxRetries) {
-          logger.error(`Maximum retries reached (${maxRetries})`);
-        }
-        logger.error(`Error code: ${status}`);
-        throw e;
-      }
-    }
-  }
-}
```
chadwhitacre, line 387: Looks like this was not actually used _anywhere_. Eh?

src/utils/system.ts
```diff
@@ -357,27 +356,6 @@ export async function extractZipArchive(
...
-  owner: string,
-  repo: string,
-  sha: string,
-  directory: string
-): Promise<void> {
-  const stream = await downloadSources(owner, repo, sha);
-  logger.info(`Extracting sources to ${directory}`);
-  return extractSourcesFromTarStream(stream, directory);
-}
-
```
chadwhitacre, line 380: Another sidecar removal—never used in the first place. Yes?

src/utils/__tests__/githubApi.test.ts
```diff
@@ -87,101 +80,3 @@ describe('getFile', () => {
...
-      .mockImplementationOnce(funcReturns);
-
-    try {
-      await retryHttp(f, { ...params, retryCodes: [500] });
-      throw Error('unreachable');
-    } catch (e) {
-      return expect(e).toEqual(errorCode(400));
-    }
-  });
-});
```
chadwhitacre, line 187: This is all loosely related dead code removal.

src/utils/git.ts
```diff
@@ -0,0 +1,42 @@
...
+): string {
+  const branchName = branch || '';
+  const remotePrefix = remoteName + '/';
+  if (branchName.startsWith(remotePrefix)) {
+    return branchName.slice(remotePrefix.length);
+  }
+  return branchName;
+}
+
+export async function getGitClient(): Promise<SimpleGit> {
```
chadwhitacre, line 30: Possible future refactor: replace direct `simpleGit` usage throughout the codebase with calls to `getGitClient`.

```
$ ag simpleGit
src/targets/ghPages.ts
5:import simpleGit from 'simple-git';
154:    await simpleGit().clone(remote.getRemoteStringWithAuth(), directory);
155:    const git = simpleGit(directory);

src/targets/upm.ts
2:import simpleGit from 'simple-git';
119:        const git = simpleGit(directory);

src/targets/crates.ts
4:import simpleGit from 'simple-git';
295:    const git = simpleGit(directory);

src/targets/awsLambdaLayer.ts
5:import simpleGit from 'simple-git';
149:        const git = simpleGit(directory);

src/targets/registry.ts
4:import simpleGit, { SimpleGit } from 'simple-git';
388:    const git = simpleGit(directory);

src/utils/git.ts
1:import simpleGit, { SimpleGit } from 'simple-git';
36:  const git = simpleGit(configFileDir);

src/config.ts
7:import simpleGit from 'simple-git';
250:    const git = simpleGit(configDir);
```

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.23.0
+
+- feat(publish): Ability to merge to non-default (#245)
```
chadwhitacre, line 5: Also note removal of #220?

src/commands/publish.ts
```diff
@@ -72,6 +72,17 @@ export const builder: CommandBuilder = (yargs: Argv) => {
...
+      alias: 'm',
+      description:
+        'Target branch to merge into. Uses the default branch from GitHub as a fallback',
+      type: 'string',
+    })
+    .option('remote', {
+      default: 'origin',
+      description: 'The git remote to use when pushing',
+      type: 'string',
+    })
```
chadwhitacre, line 85: If I'm reading things right there is already an `argv.remote` on master today. Am I reading things right? Where does it come from? Or is that prepare vs. publish ... 🤔

src/commands/publish.ts
```diff
@@ -312,58 +327,82 @@ async function checkRevisionStatus(
   await statusProvider.waitForTheBuildToSucceed(revision);
 }
 
+/**
+ * Determines the closest branch we can merge to from the current checkout
+ * Adapted from https://stackoverflow.com/a/55238339/90297
+ * @param git our local Git client
+ * @param github A GitHub client in case we need to fallback to the default branch
+ * @param githubRepo
```
chadwhitacre, line 335: Out of sync?

src/commands/publish.ts
```diff
@@ -312,58 +327,82 @@ async function checkRevisionStatus(
...
  *
  * Leave the release branch unmerged, or merge it but not delete it if the
  * corresponding flags are set.
  *
- * @param github Github client
- * @param githubConfig Github repository configuration
- * @param branchName Release branch name
- * @param skipMerge If set to "true", the branch will not be merged
+ * @param git Git client
+ * @param remoteName The git remote name to interact with
```
chadwhitacre, line 372: +branch

src/commands/publish.ts
```diff
@@ -552,19 +585,23 @@ export async function publishMain(argv: PublishOptions): Promise<any> {
     logger.info(' ');
   }
 
-  if (argv.rev) {
-    logger.info('Not merging any branches because revision was specified.');
+  if (argv.noMerge) {
```
chadwhitacre, line 588: Is `noMerge` a new option here?

chadwhitacre (commented):
First pass done, bunch of questions/comments.

For next time: consider doing all the dead code removals in a separate PR.

src/config.ts
BYK: The ones under schemas should stay to Craft not error out when this value is set. I'll remove the Readme reference.

CHANGELOG.md
BYK: Not sure it is note-worthy as we are not removing a feature. It is just obsolete.

src/commands/prepare.ts
BYK: >Is it guaranteed to be defined and non-null?

This is TypeScript, what do you think? 🙂 

https://github.com/getsentry/craft/pull/245/files#diff-233373e3f179f999c4d8c851082724070fc138206b6254f2d080d299254a53b1R74-R78

src/commands/publish.ts
BYK: >Or is that prepare vs. publish ... 🤔

This. We had this on `prepare` but not on `publish` since we were interacting with GitHub directly, not needing to know the remote for the local `git`. Now tables have turned.

src/commands/publish.ts
BYK: Yup, sorry and good catch.

src/utils/githubApi.ts
BYK: It was used in a bunch of places but was obsolete and even erroneous as we already had https://yarnpkg.com/package/@octokit/plugin-retry installed.

src/utils/system.ts
BYK: Possibly, don't recall this one but I think it was completely unused.

src/utils/git.ts
BYK: Cannot really do it as most of the other usages involve a `.clone()` making the `isRepo` check obsolete.

src/utils/__tests__/githubApi.test.ts
BYK: Yup.

src/config.ts
chadwhitacre: Bonus points: Comment in config schema that this is vestigial for backwards compatibility.

src/commands/publish.ts
BYK: Nope, it was inside `handleReleaseBranch` under the name `skipMerge`: https://github.com/getsentry/craft/pull/245/files#diff-adc8934db1a646fd72f3cdda4ede76ecfc3628ec0f8dd1f37403d0121183e94aL334

src/utils/githubApi.ts
chadwhitacre: Gotcha, sorry. Not sure how I missed it but seeing now that we removed three usages from `targets/github.ts`. 👍

src/commands/publish.ts
```diff
@@ -72,6 +72,17 @@ export const builder: CommandBuilder = (yargs: Argv) => {
         'Source revision (git SHA or tag) to publish (if not release branch head)',
       type: 'string',
     })
+    .option('merge-target', {
```
jan-auer, line 75: I've often seen "base branch" used as terminology for this, but "merge target" is equally clear.

src/commands/publish.ts
```diff
@@ -312,58 +327,82 @@ async function checkRevisionStatus(
...
   keepBranch = false
 ): Promise<void> {
-  if (!branchName || skipMerge) {
-    logger.info('Skipping the merge step.');
-    return;
+  if (!mergeTarget) {
+    mergeTarget = await getMergeTarget(git, remoteName);
   }
+  logger.debug(`Checking out merge target branch:`, mergeTarget);
+  await git.checkout(mergeTarget);
```
jan-auer, line 387: Does this function run _after_ dirty checks on the repository? In rare cases where we still run publish locally, it would be good to warn the same way as prepare does.

src/commands/publish.ts
```diff
@@ -312,58 +327,82 @@ async function checkRevisionStatus(
...
-    await mergeReleaseBranch(
-      github,
-      githubConfig.owner,
-      githubConfig.repo,
-      branchName
-    );
+    await git
+      .pull(remoteName, mergeTarget, ['--rebase'])
+      .merge(['--no-ff', '--no-edit', branch])
+      .push(remoteName);
```
jan-auer, line 394: Note: This requires the upstream branch to be set correctly. My assumption is that this is always the case; although it may be safer to provide the remote refspec, too.

jan-auer (commented):
Did a quick pass and looks good to me.

src/commands/publish.ts
BYK: Don't think we need dirty checks as `git checkout` itself fails when the working directory is not clean. That's what I relied on. If you think this is dangerous, I'll follow up with explicit checks tho.

src/commands/publish.ts
BYK: Oh, good catch, will update this.

src/commands/publish.ts
jan-auer: You're right, the checkout will bail in time. Thanks for clarifying 👍

PR #249, opened 2021-05-28
feat(registry-target): Allow batched updates w/ new config

### LINKED ISSUE #48: registry target: batch updates if there are multiple packages to release

BYK:
Fixes #48 by introducing a new registry target config structure while
keeping backward compatibility.

Now we expect two dicts under the names `apps` and `sdks` where the keys
the canonical names and the values are individual configs for these
canonical names.

It also does a single commit for all changes instead of one commit per
canonical name.

This new structure also has the side effect of making it very hard to
miss required configuration.

Tested with `craft` in dry-run mode.

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -1,7 +1,4 @@
-minVersion: '0.21.0'
-github:
-  owner: getsentry
-  repo: craft
+minVersion: '0.22.2'
 changelogPolicy: auto
 requireNames:
   - /^sentry-craft.*\.tgz$/
@@ -19,13 +16,12 @@ targets:
         metadata:
           cacheControl: 'public, max-age=300'
   - name: registry
-    type: app
-    urlTemplate: 'https://downloads.sentry-cdn.com/craft/{{version}}/{{file}}'
-    checksums:
-      - algorithm: sha256
-        format: hex
-    config:
-      canonical: 'app:craft'
+    apps:
+      'app:craft':
+        urlTemplate: 'https://downloads.sentry-cdn.com/craft/{{version}}/{{file}}'
+        checksums:
+          - algorithm: sha256
+            format: hex
   - name: docker
     source: us.gcr.io/sentryio/craft
     target: getsentry/craft
--- a/README.md
+++ b/README.md
@@ -803,14 +803,14 @@ _none_
 
 **Configuration**
 
-| Option             | Description                                                                                                                                                                                                                               |
-| ------------------ | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `apps`             | List of `app` configs as a dict, keyed by their canonical names (example: `app:craft`)                                                                                                                                                                                               |
-| `sdks` | List of `sdk` configs as a dict, keyed by their canonical names (example: `maven:io.sentry:sentry`)                                                                                                                             |
-| `(sdks|apps).urlTemplate`      | **optional** URL template that will be used to generate download links for "app" package type.                                                                                                                                            |
-| `(sdks|apps).linkPrereleases`  | **optional** Update package versions even if the release is a preview release, "false" by default.                                                                                                                                        |
-| `(sdks|apps).checksums`        | **optional** A list of checksums that will be computed for matched files (see `includeNames`). Every checksum entry is an object with two attributes: algorithm (one of `sha256`, `sha384`, and `sha512`) and format (`base64` and `hex`). |
-| `(sdks|apps).onlyIfPresent`    | **optional** A file pattern. The target will be executed _only_ when the matched file is found.                                                                                                                                           |
+| Option | Description                                                                                         |
+| ------ | --------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `apps` | List of `app` configs as a dict, keyed by their canonical names (example: `app:craft`)              |
+| `sdks` | List of `sdk` configs as a dict, keyed by their canonical names (example: `maven:io.sentry:sentry`) |
+| `(sdks | apps).urlTemplate`                                                                                  | **optional** URL template that will be used to generate download links for "app" package type.                                                                                                                                             |
+| `(sdks | apps).linkPrereleases`                                                                              | **optional** Update package versions even if the release is a preview release, "false" by default.                                                                                                                                         |
+| `(sdks | apps).checksums`                                                                                    | **optional** A list of checksums that will be computed for matched files (see `includeNames`). Every checksum entry is an object with two attributes: algorithm (one of `sha256`, `sha384`, and `sha512`) and format (`base64` and `hex`). |
+| `(sdks | apps).onlyIfPresent`                                                                                | **optional** A file pattern. The target will be executed _only_ when the matched file is found.                                                                                                                                            |
 
 **Example**
 
--- a/README.md
+++ b/README.md
@@ -803,14 +803,14 @@ _none_
 
 **Configuration**
 
-| Option | Description                                                                                         |
-| ------ | --------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
-| `apps` | List of `app` configs as a dict, keyed by their canonical names (example: `app:craft`)              |
-| `sdks` | List of `sdk` configs as a dict, keyed by their canonical names (example: `maven:io.sentry:sentry`) |
-| `(sdks | apps).urlTemplate`                                                                                  | **optional** URL template that will be used to generate download links for "app" package type.                                                                                                                                             |
-| `(sdks | apps).linkPrereleases`                                                                              | **optional** Update package versions even if the release is a preview release, "false" by default.                                                                                                                                         |
-| `(sdks | apps).checksums`                                                                                    | **optional** A list of checksums that will be computed for matched files (see `includeNames`). Every checksum entry is an object with two attributes: algorithm (one of `sha256`, `sha384`, and `sha512`) and format (`base64` and `hex`). |
-| `(sdks | apps).onlyIfPresent`                                                                                | **optional** A file pattern. The target will be executed _only_ when the matched file is found.                                                                                                                                            |
+| Option                         | Description                                                                                                                                                                                                                                |
+| ------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `apps`                         | List of `app` configs as a dict, keyed by their canonical names (example: `app:craft`)                                                                                                                                                     |
+| `sdks`                         | List of `sdk` configs as a dict, keyed by their canonical names (example: `maven:io.sentry:sentry`)                                                                                                                                        |
+| `(sdks\|apps).urlTemplate`     | **optional** URL template that will be used to generate download links for "app" package type.                                                                                                                                             |
+| `(sdks\|apps).linkPrereleases` | **optional** Update package versions even if the release is a preview release, "false" by default.                                                                                                                                         |
+| `(sdks\|apps).checksums`       | **optional** A list of checksums that will be computed for matched files (see `includeNames`). Every checksum entry is an object with two attributes: algorithm (one of `sha256`, `sha384`, and `sha512`) and format (`base64` and `hex`). |
+| `(sdks\|apps).onlyIfPresent`   | **optional** A file pattern. The target will be executed _only_ when the matched file is found.                                                                                                                                            |
 
 **Example**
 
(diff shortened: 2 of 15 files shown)


src/artifact_providers/base.ts
```diff
@@ -240,23 +240,23 @@ export abstract class BaseArtifactProvider {
...
     this.logger.debug(`Fetching artifact list for revision \`${revision}\`.`);
     // check the cache first
-    const cached = this.fileListCache[revision];
-    if (cached) {
+    if (this.fileListCache[revision]) {
       this.logger.debug(`Found list in cache.`);
-      return cached;
+    } else {
+      // Cache the promise immediately to cause any subsequent calls during the
+      // fetch to use the pending promise instead of fetching again in parallel
```
BYK, line 247: This change was required as now there can be concurrent calls to this method. I'd argue it should have been this way from the start though :)

src/artifact_providers/base.ts
```diff
@@ -240,23 +240,23 @@ export abstract class BaseArtifactProvider {
...
+      this.fileListCache[revision] = this.doListArtifactsForRevision(revision);
     }
 
-    // the data wasn't in the cache, so now we have to go get it
-    let artifacts;
+    let artifacts: RemoteArtifact[];
     try {
-      artifacts = await this.doListArtifactsForRevision(revision);
+      artifacts = (await this.fileListCache[revision]) || [];
```
BYK, line 253: The `|| []` fallback is not really necessary but TypeScript insisted that this may be `undefined`. It clearly cannot read the `if` block above 😀

src/targets/awsLambdaLayer.ts
```diff
@@ -7,7 +7,7 @@ import {
   getAuthUsername,
   getGithubApiToken,
   getGithubClient,
-  GithubRemote,
+  GitHubRemote,
```
BYK, line 10: I was just annoyed by the incorrect capitalization but now I realize there is more to fix. I'll revert this change.

src/utils/registry.ts
```diff
@@ -1,10 +1,19 @@
-import * as fs from 'fs';
+import { promises as fsPromises, existsSync } from 'fs';
```
BYK, line 1: The changes in this file are to make these operations async so we can concurrently read/write these manifest files.

src/artifact_providers/base.ts
BYK: Turns out this was due to the incorrect type def above. Fixing...

chadwhitacre (commented):
A couple questions during my first pass here ...

I see that `listArtifactsForRevision` occurs in `stores/zeus.ts`. Looks like that file is copy/pasta, and I see that it can go away at any time (right? and actually I think _all_ of Zeus can go away basically now-ish?). Removing entirely is out of scope for this PR, of course, but just wanted to double check that we don't need to update the logic in the store here. Do we? :)

Secondly, I don't see any README update. Surely with a new config file format we should have a README update here, no?

.craft.yml
```diff
@@ -1,7 +1,4 @@
-minVersion: '0.21.0'
-github:
-  owner: getsentry
-  repo: craft
+minVersion: '0.22.2'
```
chadwhitacre, line 1: I think this should be `0.22.3`.

BYK: Good guess but no 🙂 

The latest released version of Craft is `0.22.2` and this new version will be `0.23.0`. That said, even then we cannot set this to `0.23.0` as the validation of the config would fail at the `craft prepare` stage when we try to release `0.23.0`. Ironically, we use the latest master of Craft when publishing so we will be able to use this new feature _before_ it is released in `0.23.0`.

I'll follow up with a patch to bring this number to `0.23.0` after the release tho.

BYK:
>I see that listArtifactsForRevision occurs in stores/zeus.ts. Looks like that file is copy/pasta, and I see that it can go away at any time (right? and actually I think all of Zeus can go away basically now-ish?). Removing entirely is out of scope for this PR, of course, but just wanted to double check that we don't need to update the logic in the store here. Do we? :)

Eh, great catch. I think Zeus stuff predates the `BaseArtifactProvider` days. I'll do a separate pass to DRY that up. I need to do a full audit to remove Zeus and I don't have enough time for that right now. Will schedule tho.

>Secondly, I don't see any README update. Surely with a new config file format we should have a README update here, no?

For sure, great catch thanks!

.craft.yml
chadwhitacre: In that case why bother bumping `minVersion` here at all? Is that strictly necessary?

src/targets/registry.ts
```diff
@@ -101,7 +98,34 @@ export class RegistryTarget extends BaseTarget {
   /**
    * Extracts Registry target options from the raw configuration.
    */
-  public getRegistryConfig(): RegistryConfig {
+  public getRegistryConfig(): RegistryConfig[] {
+    if (!this.config.sdks && !this.config.apps) {
+      this.logger.warn(
+        'You are using a deprecated registry traget config, please update.'
```
chadwhitacre, line 104:
```suggestion
        'You are using a deprecated registry target config, please update.'
```

src/targets/registry.ts
```diff
@@ -101,7 +98,34 @@ export class RegistryTarget extends BaseTarget {
...
+    }
+    for (const canonicalName of Object.keys(this.config.apps || {})) {
+      items.push({
+        ...this.config.apps[canonicalName],
+        type: RegistryPackageType.APP,
+        canonicalName,
+      });
+    }
+
+    return items;
```
chadwhitacre, line 125: How is it that we can now skip all of the validation that's now in `getLegacyRegistryConfig`? I don't see that we added type-checking that wasn't already there, e.g.

src/targets/registry.ts
```diff
@@ -93,6 +83,13 @@ export class RegistryTarget extends BaseTarget {
...
+    const remote = this.config.remote;
+    if (remote) {
+      const [owner, repo] = remote.split('/', 2);
+      this.remote = new GithubRemote(owner, repo);
+    } else {
+      this.remote = DEFAULT_REGISTRY_REMOTE;
+    }
     this.github = getGithubClient();
     this.githubRepo = githubRepo;
     this.registryConfig = this.getRegistryConfig();
```
chadwhitacre, line 95: I see in many places that we're removing usage of `this.registryConfig` ... yet this remains, and I see that it is still used once, in `getValidItems`. Is that intentional?

src/targets/registry.ts
chadwhitacre: Is the answer related to this in the PR description?

> This new structure also has the side effect of making it very hard to
miss required configuration.

.craft.yml
BYK: >In that case why bother bumping minVersion here at all? Is that strictly necessary?

Yes as I remove the top-level `github` key which forces Craft to infer it, which is only introduced in `0.22.0`.

src/targets/registry.ts
BYK: Yup. Since this happens in the constructor, we get an early warning/validation on the config by doing this.

One can argue `getValidItems` can be merged into `publish` as its only used there though.

src/targets/registry.ts
```diff
@@ -396,78 +423,88 @@ export class RegistryTarget extends BaseTarget {
     return git;
   }
 
+  public async getValidItems(
```
chadwhitacre, line 426: Could be `private`?

src/targets/registry.ts
BYK: lol will fix thanks!

src/targets/registry.ts
BYK: >Is the answer related to this in the PR description?

Yes! We should still type-check this but I'm hoping to do that in a future diff where I enable strict target config type checks through `validateConfig` in `config.ts`.

.craft.yml
chadwhitacre: > which forces Craft to infer it

Sorry ... to infer what?

I guess more generally, why do we ever need to specify a `minVersion` here if we will always use the latest version anyway? Maybe I'm misunderstanding ...

src/targets/registry.ts
BYK: Eh?

.craft.yml
BYK: See #208 re "infer what".

>why do we ever need to specify a minVersion here if we will always use the latest version anyway? Maybe I'm misunderstanding

We _hope_ to use the latest version at all times but we don't. `minVersion` ensures capabilities and sometimes config stability as in the case of #202.

src/targets/registry.ts
chadwhitacre: This could be a private method, yes? Since it's only used from `this.publish`?

src/targets/registry.ts
```diff
@@ -396,78 +424,88 @@ export class RegistryTarget extends BaseTarget {
...
-    if (!RegistryTarget.localRepo) {
-      await withTempDir(
-        async dir => {
-          RegistryTarget.localRepo = {
-            dir,
-            git: await this.cloneRegistry(dir),
-          };
-          process.on('beforeExit', () => {
-            RegistryTarget.localRepo = undefined;
-            rimraf(localRepo.dir, () => {
```
chadwhitacre, line 438: No more tempdir cleanup?

src/targets/registry.ts
BYK: Yes it can be. I just don't really see a reason to discuss/change this. Making this public would make testing easier.

src/targets/registry.ts
BYK: `withTempDir` does it automatically now: https://github.com/getsentry/craft/pull/249/files#diff-c6afb4a580905af2619fdd866dc35a21d6fc15ff37364697a4355f6d2b833d4aR506 (passing `true` there means "clean up once the callback finishes)

src/utils/registry.ts
```diff
@@ -13,18 +22,26 @@ import { GithubRemote } from './githubApi';
...
   }
   const packageManifestPath = path.join(baseDir, packageDirPath, 'latest.json');
-  logger.debug('Reading the current configuration from "latest.json"...');
-  return JSON.parse(fs.readFileSync(packageManifestPath).toString()) || {};
+  logger.debug('Reading the current configuration from', packageManifestPath);
+  return {
+    versionFilePath,
+    packageManifest:
+      JSON.parse(
+        await fsPromises.readFile(packageManifestPath, { encoding: 'utf-8' })
```
chadwhitacre, line 42: For my own learning: this is equivalent to using `readFileSync`, yes?

BYK: Codewise, yes. Runtime, it should be slightly slower as it adds async cycles but should allow other things to happen in the meantime (like other I/O).

src/targets/registry.ts
```diff
@@ -396,78 +424,88 @@ export class RegistryTarget extends BaseTarget {
...
+        await Promise.all(
+          items.map(registryConfig =>
+            this.updateVersionInRegistry(
+              registryConfig,
+              localRepo,
+              version,
+              revision
+            )
+          )
+        );
```
chadwhitacre, line 485: So this is the crucial part, where we update the filesystem for all registry sub-targets in parallel, and block on completion of all via `await Promise.all`. 👍

BYK: Exactly 🎉

src/targets/registry.ts
chadwhitacre: > would make testing easier

I _did_ notice a lack of tests. 😜

BYK: Hopefully there will be. For now we just rely on the extensive manual testing that I did. There were no tests anyway so I'm not leaving this place worse than I found it? 😝

README.md
```diff
@@ -794,38 +794,37 @@ the corresponding package directory can be found inside "packages" directory of
...
+| `(sdks | apps).linkPrereleases`                                                                              | **optional** Update package versions even if the release is a preview release, "false" by default.                                                                                                                                         |
+| `(sdks | apps).checksums`                                                                                    | **optional** A list of checksums that will be computed for matched files (see `includeNames`). Every checksum entry is an object with two attributes: algorithm (one of `sha256`, `sha384`, and `sha512`) and format (`base64` and `hex`). |
+| `(sdks | apps).onlyIfPresent`                                                                                | **optional** A file pattern. The target will be executed _only_ when the matched file is found.                                                                                                                                            |
```
chadwhitacre, line 813: Is this GitHub's viewer being wonky, or bad formatting?

(P.S. [Grip](https://github.com/joeyespo/grip) can be useful here.)

Screen Shot 2021-06-01 at 5 44 17 PM

BYK: It was bad formatting, fixed now! 😁

chadwhitacre (approved):
Ship it! 👍

PR #251, opened 2021-06-02
ref(logging): Overhaul debug level logging

BYK:
We are planning to set logging level to `Debug` over at `getsentry/publish` but it tends to be too noisy. This PR goes over all `logger.debug()` calls, tidies them up, and moves some of them to `logger.trace()` level to make debug-logging more usable.

### DIFF
--- a/src/utils/env.ts
+++ b/src/utils/env.ts
@@ -40,7 +40,7 @@ export interface RequiredConfigVar {
 function envHasVar(envVar: RequiredConfigVar): boolean {
   const { name, legacyName } = envVar;
 
-  logger.debug(`Checking for environment variable ${name}`);
+  logger.debug('Checking for environment variable:', name);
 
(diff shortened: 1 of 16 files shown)


src/utils/env.ts
```diff
@@ -1,6 +1,8 @@
 import { existsSync, statSync } from 'fs';
 import { join } from 'path';
-const os = require('os');
+// XXX(BYK): This is to be able to spy on `homedir()` in tests
+// TODO(BYK): Convert this to ES6 imports
```
chadwhitacre, line 4: Why is this hard?

chadwhitacre (approved):

src/utils/env.ts
BYK: Eh, multiple layers of issues. Will explain later.

PR #253, opened 2021-06-03
ref(zeus): Remove all Zeus support

BYK:
See getsentry/publish#360 for more context.

This PR also dropped the binary size to 3MB from (3.3MB)


src/commands/publish.ts
```diff
@@ -295,12 +295,10 @@ async function checkRequiredArtifacts(
   }
 }
 
-// TODO there is at least one case that is not covered: how to detect Zeus builds
-// that have unknown status (neither failed nor succeeded)
 /**
- * Checks statuses of all builds on Zeus for the provided revision
+ * Checks statuses of all builds on status provider for the provided revision
```
chadwhitacre, line 299:
```suggestion
 * Checks statuses of all builds on statusProvider for the provided revision
```

Or:

```suggestion
 * Checks statuses of all builds on the status provider for the provided revision
```

src/config.ts
```diff
@@ -296,17 +267,7 @@ export async function getArtifactProviderFromConfig(): Promise<BaseArtifactProvi
...
-      artifactProviderName = ArtifactProviderName.Github;
-    } else {
-      logger.warn(
-        `You are relying on the default artifact provider, which has changed Craft v${DEFAULTS_EPOCH_VERSION}.`
-      );
-      logger.warn(
-        `This will affect you when you set your \`minVersion\` in your config to ${DEFAULTS_EPOCH_VERSION} or later.`
-      );
-      artifactProviderName = ArtifactProviderName.Zeus;
-    }
```
chadwhitacre, line 309: Good thing we're pre-1.0! 😬

src/schemas/projectConfig.schema.ts
```diff
@@ -60,8 +60,8 @@ const projectConfigJsonSchema = {
           title: 'StatusProviderName',
           description: 'Name of the status provider',
           type: 'string',
-          enum: ['zeus', 'github'],
-          tsEnumNames: ['Zeus', 'Github'],
+          enum: ['github'],
+          tsEnumNames: ['Github'],
```
chadwhitacre, line 64: If an option only has one option, and it's required, is it an option? 🤔

chadwhitacre (changes requested):
Tiny comment tweak plz, otherwise this checks out. 👍

```
$ ag zeus
CHANGELOG.md
5:- ref(zeus): Remove all Zeus support (#253)
59:    name: zeus
61:    name: zeus
64:  Support for Zeus will be dropped in a future release and we highly recommend
167:- fix(zeus): Don't force ZEUS_API_TOKEN when Zeus is not used (#97)
198:- fix: Default status provider should still be Zeus
280:- Always print Zeus links when publishing
322:- Add a missing check for `ZEUS_API_TOKEN`
```

src/schemas/projectConfig.schema.ts
BYK: Good point but I think it is good to keep this for the reasons below:
1. There's nothing stopping us from adding more status providers (GitLab, GCB etc.)
2. Removing the option would require another sweeping config change

chadwhitacre (approved):

tonyo (approved):
This guy approves.

image

BYK:
good bye

PR #258, opened 2021-06-09
feat(maven): Add `maven` target to deploy to Maven Central

iker-barriocanal:
The goal of the new `maven` target is to automate the publishing of [`sentry-java`](https://github.com/getsentry/sentry-java) (Android and Java) to Maven Central.


AbhiPrasad (commented):
Looks good so far! 

We might have to adjust the maven upload command to be a little more configurable but I think that’s a non-issue right now

src/targets/maven.ts
```diff
@@ -0,0 +1,163 @@
...
+import { join } from 'path';
+import * as fs from 'fs';
+
+// TODO: add docs to the readme
+
+const GIT_REPO_OWNER = 'getsentry';
+const GIT_REPO_NAME = 'sentry-java';
+const FILES_TO_COMMIT = ['gradle.properties'];
+
+const USER_GRADLE_PROPS_FILE = join(homedir(), '/.gradle/gradle.properties');
```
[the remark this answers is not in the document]
iker-barriocanal: IMO, no. This file contains the secrets of Maven Central, so it shouldn't be in the project directory. If you see the [Gradle Build Environment](https://docs.gradle.org/current/userguide/build_environment.html), the choices if not using the project directory are 1) `$GRADLE_USER_HOME` and the default 2) `$USER_HOME/.gradle`; so I've added the option to use the former (instead of using directly the default value). This should allow more customization.

src/targets/maven.ts
```diff
@@ -0,0 +1,163 @@
...
+    if (process.env[envVar]) {
+      return process.env[envVar] as string; // `as string` to make TS happy
+    }
+    throw new ConfigurationError(
+      `Cannot publish to Maven Central: missing credentials.
+      Please, use the ${envVar} environment variable.`
+    );
+  }
+
+  private getEnvVarOrDefault(envVar: string): string {
```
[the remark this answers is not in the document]
iker-barriocanal: I don't understand what you're proposing. Can you give an example, and tell how it improves it? IMO typing it as `string` is fine.

src/targets/maven.ts
```diff
@@ -0,0 +1,163 @@
...
+    if (process.env[envVar]) {
+      return process.env[envVar] as string; // `as string` to make TS happy
+    }
+    return DEFAULT_ENV_VARIABLES[envVar] as string;
+  }
+
+  public async publish(version: string, _revison: string): Promise<void> {
```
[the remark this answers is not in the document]
iker-barriocanal: Yes, have a look at the [`version` module](https://github.com/getsentry/craft/blob/master/src/utils/version.ts).

src/targets/maven.ts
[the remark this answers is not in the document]
AbhiPrasad: Like we put DISTRIBUTIONS_PATH, MAVEN_REPO_ID in a string enum, then we type check to make sure the envVar is in the enum.

src/targets/maven.ts
```diff
@@ -0,0 +1,172 @@
...
+        await this.createUserGradlePropsFile();
+        const mavenReleaser = new MavenReleaser(
+          this.mavenConfig.distributionsPath,
+          this.mavenConfig.settingsPath,
+          this.mavenConfig.mavenRepoUrl,
+          this.mavenConfig.mavenRepoId,
+          this.mavenConfig.mavenCliPath,
+          this.mavenConfig.gradleCliPath
+        );
+        mavenReleaser.release();
```
AbhiPrasad, line 130: Should `mavenReleaser` operate differently in dry mode?

AbhiPrasad (commented):
Last set of comments, will then wait for undraft to finish review

src/targets/maven.ts
```diff
@@ -0,0 +1,172 @@
...
+
+        await this.createUserGradlePropsFile();
+        const mavenReleaser = new MavenReleaser(
+          this.mavenConfig.distributionsPath,
+          this.mavenConfig.settingsPath,
+          this.mavenConfig.mavenRepoUrl,
+          this.mavenConfig.mavenRepoId,
+          this.mavenConfig.mavenCliPath,
+          this.mavenConfig.gradleCliPath
+        );
```
BYK, line 129: This indicates that `MavenReleaser` should not be a separate class as you are passing so many arguments related to the current instance to this class. It seems like this should just be part of the current target class?

src/targets/maven.ts
```diff
@@ -0,0 +1,172 @@
...
+} from '../utils/githubApi';
+import simpleGit from 'simple-git';
+import { homedir } from 'os';
+import { join } from 'path';
+import * as fs from 'fs';
+
+// TODO: add docs to the readme
+
+const GIT_REPO_OWNER = 'getsentry';
+const GIT_REPO_NAME = 'sentry-java';
```
BYK, line 23: None of this should be hard-coded, see my comment below regarding git operations.

BYK (changes requested):

src/targets/maven.ts
```diff
@@ -0,0 +1,229 @@
...
+  mavenUsername: string;
+  mavenPassword: string;
+  // Optional env variables (have a default value)
+  distributionsPath: string;
+  settingsPath: string;
+  mavenRepoUrl: string;
+  mavenRepoId: string;
+  mavenCliPath: string;
+  gradleCliPath: string;
+}
```
AbhiPrasad, line 32:
```suggestion
/** Config options for the "maven" target. */
type MavenTargetConfig = {
    [key in keyof ((typeof RequiredConfig) & (typeof OptionalConfig))]: string
}
```

iker-barriocanal: This seems easier than merging enum types 😁.

AbhiPrasad (changes requested):
@bruno-garcia might also wanna take a look, esp around the maven command because I lack some context there.

README.md
```diff
@@ -988,6 +989,36 @@ targets:
...
+| ------------------------ | --------------------------------------------------------------------------------------------------------------------------------- |
+| `OSSRH_USERNAME`         | Username of Sonatype repository.                                                                                                  |
+| `OSSRH_PASSWORD`         | Password of Sonatype repository.                                                                                                  |
+| `MAVEN_CENTRAL_USERNAME` | Username of Maven Central.                                                                                                        |
+| `MAVEN_CENTRAL_PASSWORD` | Password of Maven Central.                                                                                                        |
+| `DISTRIBUTIONS_PATH`     | **optional** Path to the directory with unzipeed distribution packages. Defaults to `distributions/`.                             |
+| `SETTINGS_PATH`          | **optional** Path to the Maven `settings.xml`, containing Maven Central username and API key. Defaults to `scripts/settings.xml`. |
```
BYK, line 1005: These names are too generic. We should scope them to `MAVEN_` or `CRAFT_` (or both)

README.md
```diff
@@ -988,6 +989,36 @@ targets:
...
+| `MAVEN_REPO_URL`         | **optional** URL of the Maven repository. Defaults to `https://oss.sonatype.org/service/local/staging/deploy/maven2/`.            |
+| `MAVEN_REPO_ID`          | **optional** ID of the Maven server in the `settings.xml`. Defaults to `ossrh`.                                                   |
+| `MAVEN_CLI_PATH`         | **optional** Path to the Maven CLI. Defaults to `scripts/mvnw.cmd`.                                                               |
+| `GRADLE_CLI_PATH`        | **optional** Path to the Gradle CLI. Defaults to `gradlew`.                                                                       |
+
+**Configuration**
+
+_none_
```
BYK, line 1013: ? If there doesn't need to be anything, then just leave this section out.

I'd argue, maybe some of the ENV config can actually be moved here.

src/targets/maven.ts
```diff
@@ -1,43 +1,38 @@
...
+type MavenTargetConfig = {
+  [key in keyof (typeof RequiredConfig & typeof OptionalConfig)]: string;
+};
+
+enum RequiredConfig {
+  ossrhUsername = 'OSSRH_USERNAME',
+  ossrhPassword = 'OSSRH_PASSWORD',
+  mavenUsername = 'MAVEN_CENTRAL_USERNAME',
+  mavenPassword = 'MAVEN_CENTRAL_PASSWORD',
+}
```
BYK, line 27: Just make these a `Set`?
```suggestion
const REQUIRED_ENV_CONFIG = new Set([
  'OSSRH_USERNAME',
  'OSSRH_PASSWORD',
  'MAVEN_CENTRAL_USERNAME',
  'MAVEN_CENTRAL_PASSWORD',
]);
```

src/targets/maven.ts
```diff
@@ -102,48 +93,115 @@ export class MavenTarget extends BaseTarget {
     );
   }
 
-  private getEnvVarOrDefault(envVar: string): string {
+  private getEnvVarOrDefault(envVar: OptionalConfig): string {
```
BYK, line 96: Might be the time to use this: https://humanwhocodes.com/blog/2021/02/introducing-env-javascript-environment-variables/

src/targets/maven.ts
```diff
@@ -102,48 +93,115 @@ export class MavenTarget extends BaseTarget {
...
-          this.mavenConfig.mavenCliPath,
-          this.mavenConfig.gradleCliPath
-        );
-        mavenReleaser.release();
-
-        await git.add(FILES_TO_COMMIT);
-        await git.commit(`craft(maven): Deployed ${version} to Maven Central.`);
-        if (this.shouldPush()) {
-          await git.push();
+  public async publish(_version: string, _revison: string): Promise<void> {
```
BYK, line 103: If you are not using `_version` and `_revision`, something feels very off to me>

src/targets/maven.ts
```diff
@@ -0,0 +1,222 @@
...
+  }
+
+  public async publish(_version: string, _revison: string): Promise<void> {
+    this.createUserGradlePropsFile();
+    if (isDryRun()) {
+      this.logger.info('[dry-run] Not uploading to Maven.');
+      return;
+    }
+    this.upload();
+    this.closeAndRelease();
```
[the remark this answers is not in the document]
BYK: `execCommand` already does this.

src/targets/maven.ts
```diff
@@ -102,48 +93,115 @@ export class MavenTarget extends BaseTarget {
...
+      this.mavenConfig.distributionsPath
+    );
+    for (const distDir of distributionsDirs) {
+      const moduleName = path.parse(distDir).base;
+      const androidFile = this.getAndroidDistributionFile(distDir);
+      const targetFile = androidFile
+        ? androidFile
+        : path.join(distDir, `${moduleName}.jar`);
+      const javadocFile = path.join(distDir, `${moduleName}-javadoc.jar`);
+      const sourcesFile = path.join(distDir, `${moduleName}-sources.jar`);
```
BYK, line 128: Where are we getting these from? These should be fetched from the artifact provider from the release build?

This is why you need the `version` and `revision` arguments in `publish`: to get the associated artifacts.

src/targets/maven.ts
```diff
@@ -102,48 +93,115 @@ export class MavenTarget extends BaseTarget {
...
+
+  /**
+   * Deploys to Maven Central the distribution packages.
+   * Note that after upload, this must be `closeAndRelease`.
+   */
+  public upload(): void {
+    const distributionsDirs = fs.readdirSync(
+      this.mavenConfig.distributionsPath
+    );
+    for (const distDir of distributionsDirs) {
```
BYK, line 121: Looks like these don't depend on each other and can all be done in parallel using the
```js
await Promise.all(distributionsDirs.map(async distDir => {
    ...
}));
```

pattern.

src/targets/maven.ts
```diff
@@ -102,48 +93,115 @@ export class MavenTarget extends BaseTarget {
...
+    });
+  }
+
+  /**
+   * Returns the path to the first Android distribution file, if any.
+   */
+  public getAndroidDistributionFile(
+    distributionDir: string
+  ): string | undefined {
+    const files = fs.readdirSync(distributionDir);
```
BYK, line 166: Can this be part of the target config? Directory walking operations are quite expensive and you seem to be doing a lot of them.

src/targets/maven.ts
```diff
@@ -102,48 +93,115 @@ export class MavenTarget extends BaseTarget {
...
+    return (
+      `./${this.mavenConfig.mavenCliPath} gpg:sign-and-deploy-file ` +
+      `-Dfile=${targetFile} ` +
+      `-Dfiles=${javadocFile},${sourcesFile} ` +
+      `-Dclassifiers=javadoc,sources ` +
+      `-Dtypes=jar,jar ` +
+      `-DpomFile=${pomFile} ` +
+      `-DrepositoryId=${this.mavenConfig.mavenRepoId} ` +
+      `-Durl=${this.mavenConfig.mavenRepoUrl} ` +
+      `--settings ${this.mavenConfig.settingsPath} `
```
BYK, line 197: This looks like a ticking bomb without proper shell escaping. The `exec` function should be doing this, why are you building a giant string instead of using a proper array so the function can do proper shell escaping?

BYK (changes requested):
Everything @AbhiPrasad said + my comments :)

Seems like you are going in the right direction though.

src/targets/maven.ts
```diff
@@ -0,0 +1,222 @@
...
+   * Returns the command to be executed, using the given parameters.
+   */
+  public getMavenUploadCmd(
+    targetFile: string,
+    javadocFile: string,
+    sourcesFile: string,
+    pomFile: string
+  ): string {
+    return (
+      `./${this.mavenConfig.mavenCliPath} gpg:sign-and-deploy-file ` +
```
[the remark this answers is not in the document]
iker-barriocanal: Yes, this requires they have [Maven CLI](https://github.com/getsentry/sentry-java/blob/main/scripts/mvnw) and `GnuPG`. If any of these is missing, the target will fail. For now, the plan is to make releases manually to see how it goes and then cover these.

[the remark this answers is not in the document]
BYK: Let's check this before we start the process and bail out early?

src/targets/maven.ts
iker-barriocanal: Why? Using an `enum` allows to not hardcode the variable multiple times, e.g. [here](https://github.com/getsentry/craft/blob/48380191fb7551311e53cca541e89d5784dc15f3/src/targets/maven.ts#L71-L84); and to have stronger typing (instead of weaker types and checking afterward), e.g. [here](https://github.com/getsentry/craft/blob/48380191fb7551311e53cca541e89d5784dc15f3/src/targets/maven.ts#L88).

src/targets/maven.ts
```diff
@@ -0,0 +1,250 @@
...
+enum OptionalConfig {
+  distributionsPath = 'MAVEN_DISTRIBUTIONS_PATH',
+  settingsPath = 'MAVEN_SETTINGS_PATH',
+  mavenRepoUrl = 'MAVEN_REPO_URL',
+  mavenRepoId = 'MAVEN_REPO_ID',
+  mavenCliPath = 'MAVEN_CLI_PATH',
+  gradleCliPath = 'GRADLE_CLI_PATH',
+}
+
+// Paths should be relative to the root of the repository
```
BYK, line 39: Correction: paths should be relative to the `.craft.yml` file (the config file)

src/targets/maven.ts
```diff
@@ -0,0 +1,250 @@
...
+      mavenPassword: this.getEnvVarValue(RequiredConfig.mavenPassword),
+      // Optional env variables
+      distributionsPath: this.getEnvVarOrDefault(
+        OptionalConfig.distributionsPath
+      ),
+      settingsPath: this.getEnvVarOrDefault(OptionalConfig.settingsPath),
+      mavenRepoUrl: this.getEnvVarOrDefault(OptionalConfig.mavenRepoUrl),
+      mavenRepoId: this.getEnvVarOrDefault(OptionalConfig.mavenRepoId),
+      mavenCliPath: this.getEnvVarOrDefault(OptionalConfig.mavenCliPath),
+      gradleCliPath: this.getEnvVarOrDefault(OptionalConfig.gradleCliPath),
```
BYK, line 84: The repetition here is killing me. This is why I didn't want these to be enums. I'll try to find a better way for this.

src/targets/maven.ts
BYK: Did you have a look at this? Also, you can use `checkEnvForPrerequisite` from the `utils/env` module for required env variables.

src/targets/maven.ts
```diff
@@ -0,0 +1,250 @@
...
+    return DEFAULT_ENV_VARIABLES[envVar];
+  }
+
+  /**
+   * Checks whether the required software to run this target is available
+   * in the system. It assumes the config for this target to be available.
+   */
+  private checkRequiredSoftware(): void {
+    this.logger.debug('Checking Maven CLI is available...');
+    checkExecutableIsPresent(this.mavenConfig.mavenCliPath);
```
BYK, line 111: Adding the paths that are being checked to the debug log would make them way more useful.

src/targets/maven.ts
BYK: This comment still applies.

src/targets/maven.ts
BYK: This comment needs an answer :)

src/targets/maven.ts
```diff
@@ -0,0 +1,250 @@
+import { TargetConfig } from '../schemas/project_config';
+import { BaseArtifactProvider } from '../artifact_providers/base';
+import { BaseTarget } from './base';
+import { ConfigurationError } from '../utils/errors';
+import { homedir } from 'os';
+import * as path from 'path';
+import * as fs from 'fs';
```
BYK, line 7: Seems like you only use `promises` and `readdirSync`
```suggestion
import {promises, readdirSync} from 'fs';
```

src/targets/maven.ts
```diff
@@ -0,0 +1,250 @@
...
+import { exec } from 'child_process';
```
BYK, line 8: Use `spawnProcess` from `system/utils`
```suggestion
import { spawnProcess } from '../utils/system';
```

src/targets/maven.ts
```diff
@@ -0,0 +1,250 @@
...
+        exec(command, (error, _stdout, _stderr) => {
+          if (error) {
+            throw new Error(`Cannot upload ${distDir} to Maven:\n` + error);
+          }
+        });
+        // Not handling an exception forces `withRetry` to try executing the
+        // function again. In the very minute this point has been reached,
+        // there've been no errors running the command, so there's no need to
+        // return any other thing than resolving the promise.
+        return Promise.resolve();
```
BYK, line 161: This is not a great pattern. Let's use `spawnProcess` which already uses promises.

src/targets/maven.ts
```diff
@@ -0,0 +1,250 @@
...
+      });
+    }
+  }
+
+  /**
+   * Finishes the release flow.
+   */
+  public closeAndRelease(): void {
+    const gradleCliAbsPath = path.resolve(this.mavenConfig.gradleCliPath);
```
BYK, line 170: Instead of relying on the current working directory, I strongly recommend basing all your path operations to the config file path. `config` module has a convenience function that gives you the config file directory.

src/targets/maven.ts
```diff
@@ -0,0 +1,250 @@
...
+            throw new Error(`Cannot close and release to Maven:\n` + error);
+          }
+        }
+      );
+      // Not handling an exception forces `withRetry` to try executing the
+      // function again. In the very minute this point has been reached,
+      // there've been no errors running the command, so there's no need to
+      // return any other thing than resolving the promise.
+      return Promise.resolve();
+    });
```
BYK, line 185: Pro-tip: if you see yourself repeating a questionable pattern, at least consider encapsulating it in a helper function ;)

src/targets/maven.ts
BYK: Still applies.

BYK (changes requested):
Please address all my comments before requesting a re-review :)

Also, I'd wait till we get the main logic right before starting working on tests.

README.md
```diff
@@ -988,6 +989,36 @@ targets:
     releaseRepoName: 'unity'
 ```
 
+### Maven central (`maven`)
+
+PGP signs and publishes packages to Maven Central.
+
+**Environment**
```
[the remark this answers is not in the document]
bruno-garcia: I'll defer to @marandaneto here who has been doing these releases for the last year and a half :)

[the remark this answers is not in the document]
marandaneto: right now the signing is done via the Maven plugin `gpg:sign-and-deploy-file`
see this issue https://github.com/getsentry/sentry-java/issues/1490
ideally, we'd work on this issue first, so those values could also be configurable via Craft maybe, right now when calling `gpg:sign-and-deploy-file` pops up a keychain looking up for your generated gpg signature

src/targets/maven.ts
marandaneto: all the files except the `.asc` files which are generated during the gpg signing should be downloaded from the artifact provider.

README.md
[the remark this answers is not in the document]
BYK: >maybe, right now when calling gpg:sign-and-deploy-file pops up a keychain looking up for your generated gpg signature

That is a no-go as the opreation should work in attended mode.

src/targets/maven.ts
```diff
@@ -0,0 +1,389 @@
...
+
+  /**
+   * Checks whether the required software to run this target is available
+   * in the system. It assumes the config for this target to be available.
+   * If there's required software missing, raises an error.
+   */
+  private checkRequiredSoftware(): void {
+    this.logger.debug(
+      `Checking if Maven CLI is available on ${this.mavenConfig.mavenCliPath}...`
+    );
```
BYK, line 146:
```suggestion
    this.logger.debug(
      'Checking if Maven CLI is available:', this.mavenConfig.mavenCliPath
    );
```

src/targets/maven.ts
```diff
@@ -0,0 +1,389 @@
...
+   * `/myRegex/`
+   */
+  private makeRegexpFromConfiguration(): void {
+    this.config.androidDistDirPattern = new RegExp(
+      this.config.androidDistDirPattern
+    );
+    this.config.androidFileSearchPattern = new RegExp(
+      this.config.androidFileSearchPattern
+    );
+  }
```
BYK, line 170: Maybe you can use this:

https://github.com/getsentry/craft/blob/150310a11c4896e86df8ca26ef36e10ca5a44c8d/src/utils/filters.ts#L17-L26

src/targets/maven.ts
```diff
@@ -0,0 +1,389 @@
...
+
+  /**
+   * Creates the required user's `gradle.properties` file.
+   *
+   * If there's an existing one, it's overwritten.
+   * TODO: control when it's overwritten with an option.
+   */
+  private async createUserGradlePropsFile(): Promise<void> {
+    await fsPromises.writeFile(
```
BYK: You can save a cycle by directly returning the `writeFile` promise:
```suggestion
  private async createUserGradlePropsFile(): Promise<number> {
    return fsPromises.writeFile(
```

That said it may affect the stacktrace if an error happens.

src/targets/maven.ts
```diff
@@ -0,0 +1,389 @@
...
+      join(this.getGradleHomeDir(), GRADLE_PROPERTIES_FILENAME),
+      // Using `` instead of string concatenation makes all the lines but the
+      // first one to be indented to the right. To avoid that, these lines
+      // shouldn't have that much space at the beginning, something the linter
+      // doesn't agree with (and the code would be harder to read).
+      `mavenCentralUsername=${this.mavenConfig.MAVEN_CENTRAL_USERNAME}\n` +
+        `mavenCentralPassword=${this.mavenConfig.MAVEN_CENTRAL_PASSWORD}`
```
BYK, line 197:
```suggestion
      [
        'mavenCentralUsername=' + this.mavenConfig.MAVEN_CENTRAL_USERNAME,
        'mavenCentralPassword=' + this.mavenConfig.MAVEN_CENTRAL_PASSWORD,
      ].join('\n')
```

src/targets/maven.ts
```diff
@@ -0,0 +1,389 @@
...
+
+  /**
+   * Retrieves the Gradle Home path.
+   *
+   * See https://docs.gradle.org/current/userguide/build_environment.html#sec:gradle_environment_variables
+   *
+   * @returns the gradle home path.
+   */
+  public getGradleHomeDir(): string {
+    if (process.env.GRADLE_USER_HOME) {
```
BYK, line 209: Why not use `checkEnvForPrerequisite` like above?

src/targets/maven.ts
```diff
@@ -0,0 +1,389 @@
...
+      return process.env.GRADLE_USER_HOME;
+    }
+
+    return join(homedir(), '.gradle');
```
BYK, line 213: Would prefer if `'.gradle'` was abstracted out into a const:
```js
const DEFAULT_GRADLE_HOMEDIR = '.gradle';
```

src/targets/maven.ts
```diff
@@ -0,0 +1,389 @@
...
+        await Promise.all(
+          artifacts.map(
+            async artifact => await this.uploadArtifact(artifact, dir)
+          )
+        );
+      },
+      true,
+      'craft-release-maven-'
+    );
+  }
```
BYK, line 236: I don't see a value in having this being a separate method. What's the use case?

src/targets/maven.ts
```diff
@@ -0,0 +1,389 @@
...
+    const artifacts = await this.getArtifactsForRevision(revision);
+
+    await withTempDir(
+      async dir => {
+        await Promise.all(
+          artifacts.map(
+            async artifact => await this.uploadArtifact(artifact, dir)
+          )
+        );
+      },
```
BYK, line 232: You are overusing `await`. You can just pass promises around (like pointers in C) which would save cycles. Every `await` means another cycle on the event loop to wait until the promise is resolved only to return yet another result with an empty value in this case
```suggestion
      dir => Promise.all(artifacts.map(artifact => this.uploadArtifact(artifact, dir))),
```

src/targets/maven.ts
```diff
@@ -0,0 +1,389 @@
...
+   *
+   * @param artifact the remote artifact to be uploaded.
+   * @param dir directory where the artifact can be extracted.
+   */
+  private async uploadArtifact(
+    artifact: RemoteArtifact,
+    dir: string
+  ): Promise<void> {
+    await this.extractArtifact(artifact, dir);
+    const pkgName = basename(artifact.filename, '.zip');
```
BYK, line 249: Why are we assuming a `.zip` extension? A comment here would be great.

src/targets/maven.ts
```diff
@@ -0,0 +1,389 @@
...
+  ): Promise<void> {
+    this.logger.debug(`Downloading ${artifact.filename}...`);
+    const downloadedPkgPath = await this.artifactProvider.downloadArtifact(
+      artifact
+    );
+    this.logger.debug(
+      `Extracting ${artifact.filename} to ${downloadedPkgPath}...`
+    );
+    await extractZipArchive(downloadedPkgPath, dir);
+  }
```
BYK, line 272: Feels like you can just bundle this up in `uploadArtifact`?

src/targets/maven.ts
```diff
@@ -0,0 +1,389 @@
...
+    dir: string
+  ): Promise<void> {
+    this.logger.debug(`Downloading ${artifact.filename}...`);
+    const downloadedPkgPath = await this.artifactProvider.downloadArtifact(
+      artifact
+    );
+    this.logger.debug(
+      `Extracting ${artifact.filename} to ${downloadedPkgPath}...`
+    );
+    await extractZipArchive(downloadedPkgPath, dir);
```
BYK, line 271: Extra `await` again :D

src/targets/maven.ts
```diff
@@ -0,0 +1,389 @@
...
+   * @returns the target file name.
+   */
+  private getTargetFilename(distDir: string): string {
+    const moduleName = parse(distDir).base;
+    const isAndroidDistDir = this.config.androidDistDirPattern.test(moduleName);
+    if (isAndroidDistDir) {
+      return moduleName.replace(
+        this.config.androidFileSearchPattern,
+        this.config.androidFileReplaceStr
+      );
```
BYK, line 346: Maybe we can group these Android-related settings under a namespace:
```suggestion
    const isAndroidDistDir = this.config.android.distDirPattern.test(moduleName);
    if (isAndroidDistDir) {
      return moduleName.replace(
        this.config.android.fileSearchPattern,
        this.config.android.fileReplaceStr
      );
```

src/targets/maven.ts
```diff
@@ -0,0 +1,389 @@
...
+   * @param processFn function to be retried.
+   * @param actionName name of the action the function is performing.
+   */
+  private async retrySpawnProcess(
+    processFn: () => Promise<any>,
+    actionName: string
+  ): Promise<void> {
+    let retryDelay = RETRY_DELAY_SECS;
+    await withRetry(
+      () => processFn(),
```
BYK, line 367:
```suggestion
      processFn,
```

src/targets/maven.ts
```diff
@@ -0,0 +1,389 @@
...
+   *
+   * The function is retried `MAX_PUBLISHING_ATTEMPTS` times.
+   * The delay after each call starts at `RETRY_DELAY_SECS`, and is incremented
+   * exponentially by `RETRY_EXP_FACTOR`.
+   *
+   * @param processFn function to be retried.
+   * @param actionName name of the action the function is performing.
+   */
+  private async retrySpawnProcess(
+    processFn: () => Promise<any>,
```
BYK, line 362: This is not a great typing pattern

src/targets/maven.ts
```diff
@@ -0,0 +1,389 @@
...
+   * Retries executing the function, delaying ongoing calls for some time.
+   *
+   * The function is retried `MAX_PUBLISHING_ATTEMPTS` times.
+   * The delay after each call starts at `RETRY_DELAY_SECS`, and is incremented
+   * exponentially by `RETRY_EXP_FACTOR`.
+   *
+   * @param processFn function to be retried.
+   * @param actionName name of the action the function is performing.
+   */
+  private async retrySpawnProcess(
```
BYK, line 361: I'd say move this into `utils/system` and make the function signature very similar to `spawnProcess`

src/targets/maven.ts
```diff
@@ -0,0 +1,389 @@
...
+  ): Promise<void> {
+    let retryDelay = RETRY_DELAY_SECS;
+    await withRetry(
+      () => processFn(),
+      MAX_PUBLISHING_ATTEMPTS,
+      async err => {
+        this.logger.warn(
+          `${actionName} failed. Trying again in ${retryDelay}s.`
+        );
+        this.logger.debug(`${actionName} error: ${err}`);
```
BYK, line 373:
```suggestion
        this.logger.debug(`${actionName} error:`, err);
```

src/targets/maven.ts
```diff
@@ -0,0 +1,389 @@
...
+  }
+
+  /**
+   * Finishes the release flow, by closing and releasing to the repository.
+   * Note that this is a required step to make a release, but it doesn't
+   * perform any release on its own. Required files must have previously been
+   * uploaded accordingly.
+   */
+  public async closeAndRelease(): Promise<void> {
+    this.retrySpawnProcess(
```
BYK, line 388: Why are we retrying this one? Comments around why we do retries would be very useful.

BYK (commented):
Haven't checked the tests yet and need to do another pass but looking VERY good. Great job! 👏🏻

src/targets/maven.ts
```diff
@@ -0,0 +1,389 @@
...
+] as const;
+type OptionsType = typeof targetOptions[number];
+
+type TargetSettingType = SecretsType | OptionsType;
+
+/**
+ * Config options for the "maven" target.
+ */
+type MavenTargetConfig = {
+  [key in TargetSettingType]: string;
```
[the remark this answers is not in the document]
iker-barriocanal: `TargetSettingType` is used later (in `getTargetSecrets` and `getTargetSettings`), so I thought creating a type to avoid unions would help reading. Wdyt?

src/targets/maven.ts
iker-barriocanal: Because it's not a prerequisite and there aren't any alternative env vars. If it's not defined, Craft should use another path. `checkEnvForPrerequisite` requires the variable to be defined, and raises an error if not.

src/targets/maven.ts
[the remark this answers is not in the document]
BYK: Oh, right. Missed that sorry.

src/targets/maven.ts
iker-barriocanal: To improve code readability. By having a quick look at `publish`, it seems easier to understand the steps it does: create the gradle file, upload, and release. I used the same reasoning to extract `closeAndRelease` and not include its code in `publish`.

src/targets/maven.ts
```diff
@@ -0,0 +1,389 @@
...
+
+  private getTargetSecrets(): Record<TargetSettingType, string> {
+    const secrets = targetSecrets.map(secret => {
+      checkEnvForPrerequisite({ name: secret });
+      return {
+        name: secret,
+        value: process.env[secret],
+      };
+    });
+    return this.reduceConfig(secrets);
```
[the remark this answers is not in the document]
iker-barriocanal: I don't think this makes it easier. `every` returns a boolean so it should be apart, and the magic with types doesn't get along with the `reduce` without a prior `map` (or any other similar approach).

src/targets/maven.ts
```diff
@@ -55,14 +59,21 @@ const targetOptions = [
 ] as const;
 type OptionsType = typeof targetOptions[number];
 
+type AndroidFields = {
+  android: {
```
BYK, line 63: Isn't this an optional field?
```suggestion
  android?: {
```

src/targets/maven.ts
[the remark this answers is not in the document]
BYK: I mean, you can still do all this stuff in a single `reduce` call. It works as long as you define the type of your initial value. Something like this:
```js
targetSecrets
  .reduce((result, name) => {
    checkEnvForPrerequisite({ name });
    result[name] = process.env[name];
    return result;
  }, {}: Record<TargetSettingType, string>);
```

src/targets/maven.ts
```diff
@@ -135,14 +147,34 @@ export class MavenTarget extends BaseTarget {
...
+
   /**
    * Checks whether the required software to run this target is available
    * in the system. It assumes the config for this target to be available.
    * If there's required software missing, raises an error.
    */
   private checkRequiredSoftware(): void {
     this.logger.debug(
-      `Checking if Maven CLI is available on ${this.mavenConfig.mavenCliPath}...`
+      `Checking if Maven CLI is available: ${this.mavenConfig.mavenCliPath}`
```
BYK, line 177: Having the value as a separate argument has a different meaning and makes debugging easier, that's why I'm advocating for that. This makes a difference especially for objects.
```suggestion
      'Checking if Maven CLI is available:', this.mavenConfig.mavenCliPath
```

src/targets/maven.ts
```diff
@@ -269,6 +275,10 @@ export class MavenTarget extends BaseTarget {
       `Extracting ${artifact.filename} to ${downloadedPkgPath}...`
     );
     await extractZipArchive(downloadedPkgPath, dir);
+    // All artifacts downloaded from GitHub are ZIP files.
+    const pkgName = basename(artifact.filename, '.zip');
+    const distDir = join(dir, pkgName);
+    await this.uploadDistribution(distDir);
```
BYK: Still not following here. We already unpacked the `zip` archive at this point, right? Why do we still need that file or its filename?

src/targets/maven.ts
```diff
@@ -284,7 +294,9 @@ export class MavenTarget extends BaseTarget {
       pomFile,
     } = this.getFilesForMavenCli(distDir);
 
-    this.retrySpawnProcess(
+    // Maven central is very flaky, so retrying with an exponential delay in
+    // in case it fails.
+    retrySpawnProcess(
```
BYK, line 299: This is not the API I was trying to convey, the function signature should be same as `spawnProcess`:
```js
await retrySpawnProcess(
  this.mavenConfig.mavenCliPath,
  [
    'gpg:sign-and-deploy-file',
    `-Durl=${this.mavenConfig.mavenRepoUrl}`,
    `--settings ${this.mavenConfig.mavenSettingsPath}`,
  ],
  {
    maxRetries: MAX_PUBLISHING_ATTEMPTS,
  }
);
```

Note that you are missing an `await` in front of this function, which will cause all kinds of confusing issues.

The `RETRY_DELAY_SECS` and `RETRY_EXP_FACTOR` constants should live in the `utils/system` module and be the defaults for this helper.

src/utils/system.ts
```diff
@@ -191,6 +192,38 @@ export async function spawnProcess(
...
+ * next call; starting from `initialRetryDelay` and being incremented
+ * exponentially by `retryExpFactor`.
+ *
+ * @param fnToRetry function to retry
+ * @param actionName name of the action the function is performing
+ * @param maxPublishingAttemps maximum number of attemps
+ * @param initialRetryDelay delay between retries of running the function, in seconds
+ * @param retryExpFactor exponential backoff applied to the retry delay
+ */
+export async function retrySpawnProcess(
```
BYK, line 208: See my comments above. We'll also need a test for this now :)

BYK (changes requested):
Going well, skipped reviewing tests again for the other parts.

README.md
```diff
@@ -988,6 +989,52 @@ targets:
...
+**Configuration**
+
+| Option              | Description                                                           |
+| ------------------- | --------------------------------------------------------------------- |
+| `gradleCliPath`     | Path to the Gradle CLI. It must be executable by the calling process. |
+| `mavenCliPath`      | Path to the Maven CLI. It must be executable by the calling process.  |
+| `mavenSettingsPath` | Path to the Maven `settings.xml` file.                                |
+| `mavenRepoId`       | ID of the Maven server in the `settings.xml`.                         |
+| `mavenRepoUrl`      | URL of the Maven repository.                                          |
+| `android`           | Structure containing the data available below.                        |
```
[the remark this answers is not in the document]
iker-barriocanal: By looking at configs of other targets I see only optional settings are marked, and all settings are required here. I can do that, but I don't think it's a good idea to have a different style in this target.

src/targets/maven.ts
iker-barriocanal: No, the android configuration is required. It's the only way to get the release files (without traversing through the directories).

src/targets/maven.ts
iker-barriocanal: In order to publish a distribution, one of the files we need to find is the release file; a `.aar` for Android releases and a `.jar` for the others. Instead of traversing the whole directory, we can identify the file by doing some string replacement in the distribution name, since names are pretty similar. We need the name of the distribution for that, and the filename of the `zip` is the distribution directory.

src/targets/maven.ts
iker-barriocanal: Moved to `utils/async` instead, since mocks in `utils/system` were interfering, and IMO `async` was the most appropriate module to move to.

BYK: Eh, okay 😀

src/targets/maven.ts
BYK: This sounds like a great piece of comment that can live inside the code.

src/targets/maven.ts
```diff
@@ -0,0 +1,352 @@
...
+  public readonly mavenConfig: MavenTargetConfig;
+
+  public constructor(
+    config: TargetConfig,
+    artifactProvider: BaseArtifactProvider
+  ) {
+    super(config, artifactProvider);
+    this.mavenConfig = this.getMavenConfig();
+    this.checkRequiredSoftware();
+    this.makeRegexpFromConfiguration();
```
BYK, line 74: nit: This should be part of the `getMavenConfig` method.

src/targets/maven.ts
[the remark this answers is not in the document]
BYK: I don't like this `reduceConfig` method but won't push it harder until I write a better version myself.

src/targets/maven.ts
```diff
@@ -0,0 +1,352 @@
...
+      'Checking if Maven CLI is available: ',
+      this.mavenConfig.mavenCliPath
+    );
+    checkExecutableIsPresent(this.mavenConfig.mavenCliPath);
+    this.logger.debug(
+      'Checking if Gradle CLI is available: ',
+      this.mavenConfig.gradleCliPath
+    );
+    checkExecutableIsPresent(this.mavenConfig.gradleCliPath);
+    this.logger.debug('Checking if GPG is available in the path...');
```
BYK, line 166:
```suggestion
    this.logger.debug('Checking if GPG is available: ', this.mavenConfig.gradleCliPath);
```

src/targets/maven.ts
```diff
@@ -0,0 +1,352 @@
...
+   * Retrieves the Gradle Home path.
+   *
+   * @returns the gradle home path.
+   */
+  public getGradleHomeDir(): string {
+    if (process.env.GRADLE_USER_HOME) {
+      return process.env.GRADLE_USER_HOME;
+    }
+
+    return DEFAULT_GRADLE_USER_HOME;
```
BYK, line 220:
```suggestion
    return process.env.GRADLE_USER_HOME || DEFAULT_GRADLE_USER_HOME;
```

src/targets/maven.ts
```diff
@@ -0,0 +1,352 @@
...
+  }
+
+  /**
+   * Uploads the artifacts with the required files. This is a required step
+   * to make a release, but this doesn't perform any releases; after upload,
+   * the flow must finish with `closeAndRelease`.
+   */
+  public async upload(revision: string): Promise<void> {
+    this.logger.debug('Fetching artifact list...');
```
BYK, line 229: No need for this as we already have a more detailed debug log from the base function: https://github.com/getsentry/craft/blob/150310a11c4896e86df8ca26ef36e10ca5a44c8d/src/targets/base.ts#L85
```suggestion
```

src/targets/maven.ts
```diff
@@ -0,0 +1,352 @@
...
+    const artifacts = await this.getArtifactsForRevision(revision);
+
+    await withTempDir(
+      dir =>
+        Promise.all(
+          artifacts.map(artifact => this.uploadArtifact(artifact, dir))
+        ),
+      true,
+      'craft-release-maven-'
+    );
```
BYK, line 239: This logic has issues and is not concurrency-friendly:

1. The `withTempDir` logic should be abstracted away inside the `this.uploadArtifact` method
2. Since we reuse the same temp directory for all artifacts and these run concurrently, they may interfere with each other.

I'll push a change to address this issue.

src/targets/maven.ts
```diff
@@ -0,0 +1,352 @@
...
+    const targetFile = join(distDir, this.getTargetFilename(distDir));
+    const javadocFile = join(distDir, `${moduleName}-javadoc.jar`);
+    const sourcesFile = join(distDir, `${moduleName}-sources.jar`);
+    const pomFile = join(distDir, 'pom-default.xml');
+    return {
+      targetFile,
+      javadocFile,
+      sourcesFile,
+      pomFile,
+    };
```
BYK, line 310:
```suggestion
    return {
      targetFile: join(distDir, this.getTargetFilename(distDir)),
      javadocFile: join(distDir, `${moduleName}-javadoc.jar`),
      sourcesFile: join(distDir, `${moduleName}-sources.jar`),
      pomFile: join(distDir, 'pom-default.xml'),
    };
```

src/targets/maven.ts
```diff
@@ -0,0 +1,352 @@
...
+   * perform any release on its own. Required files must have previously been
+   * uploaded accordingly.
+   */
+  public async closeAndRelease(): Promise<void> {
+    // Maven central is very flaky, so retrying with an exponential delay in
+    // in case it fails.
+    await retrySpawnProcess(this.mavenConfig.gradleCliPath, [
+      'closeAndReleaseRepository',
+    ]);
+  }
```
BYK, line 351: I really don't like this one-line function to be abstracted out. I'll move this into the main `publish` in a push.

src/utils/__tests__/async.test.ts
```diff
@@ -4,6 +4,41 @@ import { logger } from '../../logger';
...
 
+import { retrySpawnProcess } from '../async';
+import { spawnProcess } from '../system';
+
+jest.mock('../system', () => {
+  const original = jest.requireActual('../system');
+  return {
+    ...original,
+    spawnProcess: jest.fn(original.spawnProcess),
+  };
```
BYK, line 15: I don't think we need this dance here. Fully mocking the `system` module should be enough.

src/utils/async.ts
```diff
@@ -1,4 +1,78 @@
...
+  retryOptions: RetryOptions = {},
+  spawnOptions: SpawnOptions = {},
+  spawnProcessOptions: SpawnProcessOptions = {}
+): Promise<Buffer | undefined> {
+  const argsString = args.map(arg => `"${arg}"`).join(' ');
+
+  if (isDryRun() && !spawnProcessOptions.enableInDryRunMode) {
+    logger.info('[dry-run] Not spawning process:', `${command} ${argsString}`);
+    return undefined;
+  }
```
BYK, line 58: Ideally, we wouldn't repeat this logic between here and `spawnProcess`. I'll look into ways of removing this.

src/utils/async.ts
```diff
@@ -1,4 +1,78 @@
...
+
+  const maxRetries = retryOptions.maxRetries || MAX_RETRIES;
```
BYK, line 60: This would use `MAX_RETRIES` when one sets `retryOptions.maxRetries` to `0`.
```suggestion
  const maxRetries = retryOptions.maxRetries ?? MAX_RETRIES;
```

src/utils/async.ts
```diff
@@ -1,4 +1,78 @@
...
+  const retryExpFactor = retryOptions.retryExpFactor || RETRY_EXP_FACTOR;
+  let retryDelay = retryOptions.retryDelay || RETRY_DELAY_SECS;
```
BYK, line 62:
```suggestion
  let retryDelay = retryOptions.retryDelay ?? RETRY_DELAY_SECS;
```

src/utils/__tests__/async.test.ts
```diff
@@ -4,6 +4,41 @@ import { logger } from '../../logger';
...
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  test('resolves before max retries', async () => {
+    const res = await retrySpawnProcess('ls', [], { maxRetries: 1 });
+    expect(res).toBeInstanceOf(Buffer);
+  });
+
+  test('hits max retries and exits', async () => {
```
BYK, line 28: Where are my tests that are actually testing the retry delay, exponential backoff etc? You can get some inspiration from `withRetry` tests.

src/targets/__tests__/maven.test.ts
```diff
@@ -0,0 +1,174 @@
...
+import { NoneArtifactProvider } from '../../artifact_providers/none';
+import { MavenTarget, MavenTargetConfig, targetOptions } from '../maven';
+import { retrySpawnProcess } from '../../utils/async';
+
+jest.mock('fs', () => ({
+  ...jest.requireActual('fs'),
+  promises: {
+    writeFile: async () => {
+      /* do nothing */
+    },
```
BYK, line 13: This should be `jest.fn()` so we can check the calls on it?

src/targets/__tests__/maven.test.ts
```diff
@@ -0,0 +1,174 @@
...
+  ...jest.requireActual('../../utils/async'),
+  retrySpawnProcess: jest.fn(),
+}));
+
+const targetSecrets: string[] = [
+  'OSSRH_USERNAME',
+  'OSSRH_PASSWORD',
+  'MAVEN_CENTRAL_USERNAME',
+  'MAVEN_CENTRAL_PASSWORD',
+];
```
BYK, line 33: Import this from the `maven` module, don't repeat.

src/targets/__tests__/maven.test.ts
```diff
@@ -0,0 +1,174 @@
...
+  beforeEach(() => removeTargetSecretsFromEnv());
+
+  function getExpectedValueOfConfigKey(configKey: string): string | RegExp {
+    if (
+      configKey.includes('androidDistDirPattern') ||
+      configKey.includes('androidFileSearchPattern')
+    ) {
+      return new RegExp(DEFAULT_OPTION_VALUE);
+    }
+    return DEFAULT_OPTION_VALUE;
```
BYK, line 87: Not a fan of this abstraction, a bit confusing.

src/targets/__tests__/maven.test.ts
```diff
@@ -0,0 +1,174 @@
...
+
+    await mvnTarget.upload('r3v1s10n');
+    expect(retrySpawnProcess).toBeCalledTimes(1);
+  });
+
+  test('close and release', async () => {
+    const mvnTarget = createMavenTarget(getTargetOptions());
+    await mvnTarget.closeAndRelease();
+    expect(retrySpawnProcess).toHaveBeenCalledTimes(1);
+  });
```
BYK, line 150: This is just testing the implementation, not a very useful test. I'd rather remove this case.

src/targets/__tests__/maven.test.ts
```diff
@@ -0,0 +1,174 @@
...
+    const mvnTarget = createMavenTarget(getTargetOptions());
+    mvnTarget.getArtifactsForRevision = jest
+      .fn()
+      .mockResolvedValueOnce([{ filename: 'mockArtifact.zip' }]);
+    mvnTarget.artifactProvider.downloadArtifact = jest
+      .fn()
+      .mockResolvedValueOnce('artifact/download/path');
+
+    await mvnTarget.upload('r3v1s10n');
+    expect(retrySpawnProcess).toBeCalledTimes(1);
```
BYK, line 143: Checking the arguments of `retrySpawnProcess` would be more useful than checking how many times it is called.

BYK (changes requested):

README.md
```diff
@@ -988,6 +989,52 @@ targets:
...
+PGP signs and publishes packages to Maven Central.
+
+**Environment**
+
+| Name                     | Description                      |
+| ------------------------ | -------------------------------- |
+| `OSSRH_USERNAME`         | Username of Sonatype repository. |
+| `OSSRH_PASSWORD`         | Password of Sonatype repository. |
+| `MAVEN_CENTRAL_USERNAME` | Username of Maven Central.       |
+| `MAVEN_CENTRAL_PASSWORD` | Password of Maven Central.       |
```
marandaneto, line 1003: sonatype repo and maven central are the same, so ideally only one configuration for user and pass

marandaneto: the env. var. name could be also generic as it depends on `mavenRepoUrl`.
for example, mavenRepoUrl could be pointing somewhere else hence the name OSSRH or MAVEN_CENTRAL won't make sense, it would still work though.
maybe MAVEN_USERNAME and MAVEN_PASSWORD only

README.md
```diff
@@ -988,6 +989,52 @@ targets:
...
+  - name: maven
+    gradleCliPath: ./gradlew
+    mavenCliPath: scripts/mvnw.cmd
+    mavenSettingsPath: scripts/settings.xml
+    mavenRepoId: ossrh
+    mavenRepoUrl: https://oss.sonatype.org/service/local/staging/deploy/maven2/
+    android:
+      distDirRegex: /^sentry-android-.*$/
+      fileReplaceeRegex: /\d\.\d\.\d(-SNAPSHOT)?/
+      fileReplacerStr: release.aar
```
marandaneto, line 1035: just a heads up, https://github.com/getsentry/sentry-android-gradle-plugin publishes to maven central but it's not an Android project, it's a pure Java file (jars), so the `android` config. should be optional.

src/targets/maven.ts
```diff
@@ -0,0 +1,329 @@
...
+   */
+  public async publish(_version: string, revison: string): Promise<void> {
+    await this.createUserGradlePropsFile();
+    await this.upload(revison);
+
+    // Maven central is very flaky, so retrying with an exponential delay in
+    // in case it fails.
+    await retrySpawnProcess(this.mavenConfig.gradleCliPath, [
+      'closeAndReleaseRepository',
+    ]);
```
marandaneto, line 183: right now a Gradle plugin does it (close and promote the releases): https://github.com/vanniktech/gradle-maven-publish-plugin/tree/master/src/main/kotlin/com/vanniktech/maven/publish/nexus

if we do those HTTP requests directly via Craft, we could remove the Gradle plugin for good.

iker-barriocanal: There's no support for the Gradle plugin yet, it will be supported at a later stage. This is targeting https://github.com/getsentry/sentry-java.

README.md
iker-barriocanal: Support for https://github.com/getsentry/sentry-android-gradle-plugin will come later.

src/targets/maven.ts
marandaneto: that's not really related to our sentry android gradle plugin, but rather the graadle plugin that calls `closeAndReleaseRepository`.
if we replicate what that plugin does via craft, we can eliminate the usage of 3rd party plugins to release our things

marandaneto: the maven cli does not have the ability to close the repo and promote the releases on sonatype, thats why we use this plugin, otherwise, we'd need to do it manually.

iker-barriocanal: Added a TODO, we can address this later.

src/utils/__tests__/async.test.ts
iker-barriocanal: We need it for this test: https://github.com/getsentry/craft/blob/iker/feat%2Fmaven-target/src/utils/__tests__/async.test.ts#L23-L28.

src/targets/__tests__/maven.test.ts
```diff
@@ -0,0 +1,212 @@
...
+    const callArgs = (retrySpawnProcess as jest.MockedFunction<
+      typeof retrySpawnProcess
+    >).mock.calls[0];
+
+    expect(callArgs).toHaveLength(2);
+    expect(callArgs[0]).toBeDefined();
```
BYK, line 154: You don't need this as you already have the `toEqual` check below.
```suggestion
```

src/targets/__tests__/maven.test.ts
```diff
@@ -0,0 +1,212 @@
...
+    expect(callArgs[0]).toEqual(DEFAULT_OPTION_VALUE);
+
+    const cmdArgs = callArgs[1];
+    expect(cmdArgs).toBeDefined();
```
BYK, line 158: Similar to above, better to drop this assertion.
```suggestion
```

src/targets/__tests__/maven.test.ts
```diff
@@ -0,0 +1,212 @@
...
+    expect(cmdArgs).toHaveLength(10);
+    // @ts-ignore `cmdArgs[*]` possibly undefined
```
BYK, line 160: I wonder if there's a blanket thing we can do to avoid repeating this. Maybe like `cmdArgs = cmdArgs as string[10]` or something?

src/targets/__tests__/maven.test.ts
```diff
@@ -0,0 +1,212 @@
...
+    expect(cmdArgs[0]).toEqual('gpg:sign-and-deploy-file');
+    // @ts-ignore `cmdArgs[*]` possibly undefined
+    expect(cmdArgs[1]).toMatch(new RegExp(`-Dfile=${tmpDirName}.*`));
```
BYK, line 163: One worry: the order of these arguments does not matter in real life does it? But then a snapshot test would have the same issue.

Better to be more explicit about we expect _something_ after the dirname.
```suggestion
    expect(cmdArgs[1]).toMatch(new RegExp(`-Dfile=${tmpDirName}.+`));
```

src/targets/__tests__/maven.test.ts
```diff
@@ -0,0 +1,212 @@
...
+    // @ts-ignore `cmdArgs[*]` possibly undefined
+    expect(cmdArgs[2]).toMatch(
+      new RegExp(
+        `-Dfiles=${tmpDirName}.*-javadoc\.jar,${tmpDirName}.*-sources\.jar`
```
BYK, line 167: `.*` pattern is not very efficient, if you know there's gonna be something there, use `.+` instead.
```suggestion
        `-Dfiles=${tmpDirName}.+-javadoc\.jar,${tmpDirName}.+-sources\.jar`
```

src/targets/__tests__/maven.test.ts
```diff
@@ -0,0 +1,212 @@
...
+      )
+    );
+    // @ts-ignore `cmdArgs[*]` possibly undefined
+    expect(cmdArgs[3]).toMatch(new RegExp(`-Dclassifiers=javadoc,sources`));
+    // @ts-ignore `cmdArgs[*]` possibly undefined
+    expect(cmdArgs[4]).toMatch(new RegExp(`-Dtypes=jar,jar`));
+    // @ts-ignore `cmdArgs[*]` possibly undefined
+    expect(cmdArgs[5]).toMatch(
+      new RegExp(`-DpomFile=${tmpDirName}.*pom-default\.xml`)
```
BYK, line 176:
```suggestion
      new RegExp(`-DpomFile=${tmpDirName}.+pom-default\.xml`)
```

src/targets/__tests__/maven.test.ts
```diff
@@ -0,0 +1,212 @@
...
+    // @ts-ignore `cmdArgs[*]` possibly undefined
+    expect(cmdArgs[6]).toMatch(
+      new RegExp(`-DrepositoryId=${DEFAULT_OPTION_VALUE}`)
+    );
+    // @ts-ignore `cmdArgs[*]` possibly undefined
+    expect(cmdArgs[7]).toMatch(new RegExp(`-Durl=${DEFAULT_OPTION_VALUE}`));
+    // @ts-ignore `cmdArgs[*]` possibly undefined
+    expect(cmdArgs[8]).toMatch(new RegExp(`--settings`));
+    // @ts-ignore `cmdArgs[*]` possibly undefined
+    expect(cmdArgs[9]).toMatch(DEFAULT_OPTION_VALUE);
```
BYK, line 187: I think these assertions can be a simple text equality with `toBe` or something?

src/targets/maven.ts
```diff
@@ -188,7 +175,17 @@ export class MavenTarget extends BaseTarget {
...
     await this.upload(revison);
-    await this.closeAndRelease();
+
+    // Maven central is very flaky, so retrying with an exponential delay in
+    // in case it fails.
+    // TODO: close the repository by doing the requests from Craft
+    // What the gradle plugin does is a / some requests to close the repo, see
+    // https://github.com/vanniktech/gradle-maven-publish-plugin/tree/master/src/main/kotlin/com/vanniktech/maven/publish/nexus
+    // If Craft did those requests, it wouldn't be necessary to rely on a third
+    // party plugin for releases.
```
BYK, line 185: Excellent comment 💯

src/targets/maven.ts
```diff
@@ -213,30 +211,29 @@ export class MavenTarget extends BaseTarget {
...
+   * the flow must finish with `closeAndReleaseRepository`.
    */
   public async upload(revision: string): Promise<void> {
-    this.logger.debug('Fetching artifact list...');
-    const artifacts = await this.getArtifactsForRevision(revision);
+    const artifacts = await this.getArtifactsForRevision(revision, {
+      includeNames:
+        this.config.includeNames === undefined
+          ? undefined
+          : stringToRegexp(this.config.includeNames),
```
BYK, line 227: Do you really need this? I thought `getArtifactsForRevision` does this automatically? Or this should be handled in the config init section, not here.

src/utils/__tests__/async.test.ts
BYK: I think this comment still holds but I'll give you a pass hoping you'll add these tests in the near future 🙂

BYK (approved):
We should probably add a `todo` somewhere about updating the Docker image to include the necessary binaries?

src/targets/__tests__/maven.test.ts
iker-barriocanal: Yes, it shouldn't matter. Since testing regex-s like this is the only way I managed to get it tested, I thought it wasn't worth it to do an algo for it.

src/targets/maven.ts
iker-barriocanal: `getArtifactForRevision` expects `RegExp | undefined`, `includeNames` is `string | undefined`, and `stringToRegexp` expects `string`. So there must be a check for `undefined` somewhere.

BYK (approved):

src/targets/__tests__/maven.test.ts
```diff
@@ -0,0 +1,198 @@
...
+    },
+  };
+}
+
+function createMavenTarget(
+  targetOptions?: Record<string, unknown>
+): MavenTarget {
+  const finalOptions = targetOptions ? targetOptions : getTargetOptions();
+  const mergedOptions = {
+    name: 'maven',
```
AbhiPrasad, line 61: Why do we set the name here? Isn’t that a class property?

src/targets/maven.ts
```diff
@@ -0,0 +1,334 @@
...
+      return {
+        name,
+        value: process.env[name],
+      };
+    });
+    return this.reduceConfig(secrets);
+  }
+
+  private reduceConfig(
+    config: { name: string; value: string | undefined }[]
```
AbhiPrasad, line 98: Nit: prefer Array type instead of [] if the generic is not a primitive.
```suggestion
    config: Array<{ name: string; value: string | undefined }>
```

src/targets/maven.ts
```diff
@@ -0,0 +1,334 @@
...
+    );
+    checkExecutableIsPresent(this.mavenConfig.mavenCliPath);
+    this.logger.debug(
+      'Checking if Gradle CLI is available: ',
+      this.mavenConfig.gradleCliPath
+    );
+    checkExecutableIsPresent(this.mavenConfig.gradleCliPath);
+    this.logger.debug(
+      'Checking if GPG is available: ',
+      this.mavenConfig.gradleCliPath
```
AbhiPrasad, line 165: Don’t think this path applies to gpg.

src/targets/maven.ts
```diff
@@ -0,0 +1,334 @@
...
+   * If there's an existing one, it's overwritten.
+   * TODO: control when it's overwritten with an option.
+   */
+  public createUserGradlePropsFile(): Promise<void> {
+    return fsPromises.writeFile(
+      join(this.getGradleHomeDir(), GRADLE_PROPERTIES_FILENAME),
+      [
+        // OSSRH and Maven Central credentials are the same
+        'mavenCentralUsername=' + this.mavenConfig.OSSRH_USERNAME,
+        'mavenCentralPassword=' + this.mavenConfig.OSSRH_PASSWORD,
```
AbhiPrasad, line 203: We can use template string here? Same thing applies for line above.
```suggestion
        `mavenCentralPassword=${this.mavenConfig.OSSRH_PASSWORD}`,
```

AbhiPrasad (commented):
Looks good

src/targets/__tests__/maven.test.ts
BYK: It is not. It is part of the config schema (since you need to define it in `.craft.yml` anyway). We enforce it to be a static string via TypeScript at compile-time and then use it for logger name: https://github.com/getsentry/craft/blob/e84f693f23ad2de33ca13c099460e2b46127a799/src/targets/base.ts#L36

We cannot use `constructor.name` as the code gets minified. See #228 and #274 for more details.

src/targets/maven.ts
BYK: I think a real improvement would come from extracting this type into an interface. Otherwise I don't see a huge difference in terms of readability.

src/targets/maven.ts
```diff
@@ -0,0 +1,334 @@
...
+    checkExecutableIsPresent(this.mavenConfig.mavenCliPath);
+    this.logger.debug(
+      'Checking if Gradle CLI is available: ',
+      this.mavenConfig.gradleCliPath
+    );
+    checkExecutableIsPresent(this.mavenConfig.gradleCliPath);
+    this.logger.debug(
+      'Checking if GPG is available: ',
+      this.mavenConfig.gradleCliPath
+    );
```
BYK, line 166:
```suggestion
    this.logger.debug('Checking if GPG is available');
```

src/targets/maven.ts
```diff
@@ -0,0 +1,334 @@
...
+   *
+   * If there's an existing one, it's overwritten.
+   * TODO: control when it's overwritten with an option.
+   */
+  public createUserGradlePropsFile(): Promise<void> {
+    return fsPromises.writeFile(
+      join(this.getGradleHomeDir(), GRADLE_PROPERTIES_FILENAME),
+      [
+        // OSSRH and Maven Central credentials are the same
+        'mavenCentralUsername=' + this.mavenConfig.OSSRH_USERNAME,
```
BYK, line 202:
```suggestion
        `mavenCentralUsername=${this.mavenConfig.OSSRH_USERNAME}`,
```

AbhiPrasad (approved):

PR #259, opened 2021-06-10
fix(registry): Fix onlyIfPresent config on batch

BYK:
Follow up to #249.

Fixes https://github.com/getsentry/publish/actions/runs/925201945


chadwhitacre:
What's the context here? What is the bug and how does this fix it?

chadwhitacre (approved):

PR #266, opened 2021-07-09
feat(symbol-collector): Add `symbol-collector` target

iker-barriocanal:
The goal of the `symbol-collector` target is to continue automating the release of [`sentry-java`](https://github.com/getsentry/sentry-java) (Android and Java). https://github.com/getsentry/craft/pull/258 automated publishing to Maven Central, and this PR automates uploading symbols to the Android Bucket.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,16 +8,16 @@ ENV DEBIAN_FRONTEND=noninteractive \
 
 RUN apt-get -qq update \
   && apt-get install -y --no-install-recommends \
-  apt-transport-https \
-  build-essential \
-  curl \
-  dirmngr \
-  gnupg \
-  git \
-  ruby-full \
-  twine \
-  jq \
-  unzip \
+    apt-transport-https \
+    build-essential \
+    curl \
+    dirmngr \
+    gnupg \
+    git \
+    ruby-full \
+    twine \
+    jq \
+    unzip \
   && curl -fsSL https://packages.microsoft.com/config/debian/10/packages-microsoft-prod.deb -o /tmp/packages-microsoft-prod.deb \
   && dpkg -i /tmp/packages-microsoft-prod.deb \
   && rm /tmp/packages-microsoft-prod.deb \
@@ -36,7 +36,7 @@ RUN apt-get -qq update \
   && gem update --no-document --system 3.1.5 \
   && gem install cocoapods \
   # Install https://github.com/getsentry/symbol-collector
-  && symbol_collector_url=$(curl -s https://api.github.com/repos/getsentry/symbol-collector/releases/tags/1.3.1 | \
+  && symbol_collector_url=$(curl -s https://api.github.com/repos/getsentry/symbol-collector/releases/tags/1.2.1 | \
   jq -r '.assets[].browser_download_url | select(endswith("symbolcollector-console-linux-x64.zip"))') \
   && curl -sL $symbol_collector_url -o "/tmp/sym-collector.zip" \
   && unzip /tmp/sym-collector.zip -d /usr/local/bin/ \
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,14 +8,16 @@ ENV DEBIAN_FRONTEND=noninteractive \
 
 RUN apt-get -qq update \
   && apt-get install -y --no-install-recommends \
-    apt-transport-https \
-    build-essential \
-    curl \
-    dirmngr \
-    gnupg \
-    git \
-    ruby-full \
-    twine \
+  apt-transport-https \
+  build-essential \
+  curl \
+  dirmngr \
+  gnupg \
+  git \
+  ruby-full \
+  twine \
+  jq \
+  unzip \
   && curl -fsSL https://packages.microsoft.com/config/debian/10/packages-microsoft-prod.deb -o /tmp/packages-microsoft-prod.deb \
   && dpkg -i /tmp/packages-microsoft-prod.deb \
   && rm /tmp/packages-microsoft-prod.deb \
@@ -23,8 +25,8 @@ RUN apt-get -qq update \
   && curl -fsSL https://download.docker.com/linux/debian/gpg | apt-key add - \
   && apt-get update -qq \
   && apt-get install -y --no-install-recommends \
-    dotnet-sdk-3.1 \
-    docker-ce-cli \
+  dotnet-sdk-3.1 \
+  docker-ce-cli \
   && apt-get clean \
   && rm -rf /var/lib/apt/lists/* \
   && curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s --  --profile minimal -y \
@@ -32,7 +34,13 @@ RUN apt-get -qq update \
   && cargo install cargo-hack \
   # Stick with 3.1.x as 3.2.x doesn't install on Debian Buster for some reason
   && gem update --no-document --system 3.1.5 \
-  && gem install cocoapods
+  && gem install cocoapods \
+  # Install https://github.com/getsentry/symbol-collector
+  && symbol_collector_url=$(curl -s https://api.github.com/repos/getsentry/symbol-collector/releases/tags/1.3.1 | \
+  jq -r '.assets[].browser_download_url | select(endswith("symbolcollector-console-linux-x64.zip"))') \
+  && curl -sL $symbol_collector_url -o "/tmp/sym-collector.zip" \
+  && unzip /tmp/sym-collector.zip -d /usr/local/bin/ \
+  && chmod +x /usr/local/bin/SymbolCollector.Console
 
 COPY dist/craft /usr/local/bin/craft
 RUN chmod +x /usr/local/bin/craft
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ then enforces a specific workflow for managing release branches, changelogs, art
   - [Ruby Gems Index (`gem`)](#ruby-gems-index-gem)
   - [AWS Lambda Layer (`aws-lambda-layer`)](#aws-lambda-layer-aws-lambda-layer)
   - [Unity Package Manager (`upm`)](#unity-package-manager-upm)
+  - [Maven central (`maven`)](#maven-central-maven)
   - [Symbol Collector (`symbol-collector`)](#symbol-collector-symbol-collector)
 - [Integrating Your Project with `craft`](#integrating-your-project-with-craft)
 - [Pre-release (Version-bumping) Script: Conventions](#pre-release-version-bumping-script-conventions)
@@ -83,7 +84,7 @@ npm install -g @sentry/craft
 
 ## Usage
 
-```plain
+```shell
 $ craft -h
 craft <command>
 
@@ -121,7 +122,7 @@ variables or by adding values to a configuration file (see below).
 All command line flags can be set through environment variables by prefixing
 them with `CRAFT_` and converting them to UPPERCASE_UNDERSCORED versions:
 
-```
+```shell
 CRAFT_LOG_LEVEL=Debug
 CRAFT_DRY_RUN=1
 CRAFT_NO_INPUT=0
@@ -159,7 +160,7 @@ Leading `export` is allowed.
 
 Example:
 
-```sh
+```shell
 # ~/.craft.env
 GITHUB_TOKEN=token123
 export NUGET_API_TOKEN=abcdefgh
@@ -175,7 +176,7 @@ that CI triggered by pushing this branch will result in release artifacts
 being built and uploaded to the artifact provider you wish to use during the
 subsequent `publish` step.
 
-```plain
+```shell
 craft prepare NEW-VERSION
 
 🚢 Prepare a new release branch
@@ -211,7 +212,7 @@ that branch. Once the checks pass, it downloads the release artifacts from the
 artifact provider configured in `.craft.yml` and uploads them to the targets named
 on the command line (and pre-configured in `.craft.yml`).
 
-```plain
+```shell
 craft publish NEW-VERSION
 
 🛫 Publish artifacts
@@ -989,6 +990,49 @@ targets:
     releaseRepoName: 'unity'
 ```
 
+### Maven central (`maven`)
+
+PGP signs and publishes packages to Maven Central.
+
+**Environment**
+
+| Name             | Description                      |
+| ---------------- | -------------------------------- |
+| `OSSRH_USERNAME` | Username of Sonatype repository. |
+| `OSSRH_PASSWORD` | Password of Sonatype repository. |
+
+**Configuration**
+
+| Option              | Description                                                           |
+| ------------------- | --------------------------------------------------------------------- |
+| `gradleCliPath`     | Path to the Gradle CLI. It must be executable by the calling process. |
+| `mavenCliPath`      | Path to the Maven CLI. It must be executable by the calling process.  |
+| `mavenSettingsPath` | Path to the Maven `settings.xml` file.                                |
+| `mavenRepoId`       | ID of the Maven server in the `settings.xml`.                         |
+| `mavenRepoUrl`      | URL of the Maven repository.                                          |
+| `android`           | Structure containing the data available below.                        |
+
+The `android` structure contains the following options:
+
+- `distDirRegex`: pattern of distribution directory names.
+- `fileReplaceeRegex` :pattern of substring of distribution module names to be replaced to get the Android distribution file.
+- `fileReplacerStr`: string to be replaced in the module names to get the Android distribution file.
+
+**Example**
+
+```yaml
+- name: maven
+  gradleCliPath: ./gradlew
+  mavenCliPath: scripts/mvnw.cmd
+  mavenSettingsPath: scripts/settings.xml
+  mavenRepoId: ossrh
+  mavenRepoUrl: https://oss.sonatype.org/service/local/staging/deploy/maven2/
+  android:
+    distDirRegex: /^sentry-android-.*$/
+    fileReplaceeRegex: /\d\.\d\.\d(-SNAPSHOT)?/
+    fileReplacerStr: release.aar
+```
+
 ### Symbol Collector (`symbol-collector`)
 
 Using the [`symbol-collector`](https://github.com/getsentry/symbol-collector) client, uploads native symbols.
@@ -1019,6 +1063,14 @@ Here is how you can integrate your GitHub project with `craft`:
 1. Set up a workflow that builds your assets and runs your tests. Allow building
    release branches (their names follow `release/{VERSION}` by default,
    configurable through `releaseBranchPrefix`).
+
+   ```yaml
+   on:
+     push:
+       branches:
+         - 'release/**'
+   ```
+
 2. Use the official `actions/upload-artifact@v2` action to upload your assets.
    Here is an example config (step) of an archive job:
 
@@ -1140,7 +1192,7 @@ configuration file) that contains a Sentry project's DSN.
 
 For example:
 
-```bash
+```shell
 export CRAFT_SENTRY_DSN='https://1234@sentry.io/2345'
 ```
 
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ then enforces a specific workflow for managing release branches, changelogs, art
   - [Ruby Gems Index (`gem`)](#ruby-gems-index-gem)
   - [AWS Lambda Layer (`aws-lambda-layer`)](#aws-lambda-layer-aws-lambda-layer)
   - [Unity Package Manager (`upm`)](#unity-package-manager-upm)
-  - [Java Symbols (`java-symbols`)](#java-symbols-java-symbols)
+  - [Symbol Collector (`symbol-collector`)](#symbol-collector-symbol-collector)
 - [Integrating Your Project with `craft`](#integrating-your-project-with-craft)
 - [Pre-release (Version-bumping) Script: Conventions](#pre-release-version-bumping-script-conventions)
 - [Post-release Script: Conventions](#post-release-script-conventions)
@@ -989,25 +989,25 @@ targets:
     releaseRepoName: 'unity'
 ```
 
-### Java Symbols (`java-symbols`)
+### Symbol Collector (`symbol-collector`)
 
 Using the [`symbol-collector`](https://github.com/getsentry/symbol-collector) client, uploads native symbols.
+The client needs to be available in the path.
 
 **Configuration**
 
-| Option           | Description                                                                              |
-| ---------------- | ---------------------------------------------------------------------------------------- |
-| `serverEndpoint` | The server endpoint.                                                                     |
-| `batchType`      | The batch type.                                                                          |
-| `bundleIdPrefix` | The prefix of the bundle ID. The new version will be appended to the end of this prefix. |
+| Option           | Description                                                                                  |
+| ---------------- | -------------------------------------------------------------------------------------------- |
+| `serverEndpoint` | **optional** The server endpoint. Defaults to `https://symbol-collector.services.sentry.io`. |
+| `batchType`      | The batch type of the symbols to be uploaded.                                                |
+| `bundleIdPrefix` | The prefix of the bundle ID. The new version will be appended to the end of this prefix.     |
 
 **Example**
 
 ```yaml
 targets:
-  - name: java-symbols
+  - name: symbol-collector
     includeNames: /libsentry(-android)?\.so/
-    serverEndpoint: my-server.com
     batchType: Android
     bundleIdPrefix: android-ndk-
 ```
--- a/README.md
+++ b/README.md
@@ -1037,14 +1037,14 @@ targets:
 ### Symbol Collector (`symbol-collector`)
 
 Using the [`symbol-collector`](https://github.com/getsentry/symbol-collector) client, uploads native symbols.
-The client needs to be available in the path.
+The `symbol-collector` needs to be available in the path.
 
 **Configuration**
 
 | Option           | Description                                                                                  |
 | ---------------- | -------------------------------------------------------------------------------------------- |
 | `serverEndpoint` | **optional** The server endpoint. Defaults to `https://symbol-collector.services.sentry.io`. |
-| `batchType`      | The batch type of the symbols to be uploaded.                                                |
+| `batchType`      | The batch type of the symbols to be uploaded. I.e: `Android`, `macOS`, `iOS`.                |
 | `bundleIdPrefix` | The prefix of the bundle ID. The new version will be appended to the end of this prefix.     |
 
 **Example**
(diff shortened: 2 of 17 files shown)


Dockerfile
```diff
@@ -8,31 +8,39 @@ ENV DEBIAN_FRONTEND=noninteractive \
...
   && rm -rf /var/lib/apt/lists/* \
   && curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s --  --profile minimal -y \
   && cargo --version \
   && cargo install cargo-hack \
   # Stick with 3.1.x as 3.2.x doesn't install on Debian Buster for some reason
   && gem update --no-document --system 3.1.5 \
-  && gem install cocoapods
+  && gem install cocoapods \
+  # Install https://github.com/getsentry/symbol-collector
+  && symbol_collector_url=$(curl -s https://api.github.com/repos/getsentry/symbol-collector/releases/tags/1.3.1 | \
```
iker-barriocanal, line 39: @bruno-garcia you may want to use another version of the Symbol Collector.

README.md
```diff
@@ -1019,18 +1020,40 @@ The `android` structure contains the following options:
...
+- name: maven
+  gradleCliPath: ./gradlew
+  mavenCliPath: scripts/mvnw.cmd
+  mavenSettingsPath: scripts/settings.xml
+  mavenRepoId: ossrh
+  mavenRepoUrl: https://oss.sonatype.org/service/local/staging/deploy/maven2/
+  android:
+    distDirRegex: /^sentry-android-.*$/
+    fileReplaceeRegex: /\d\.\d\.\d(-SNAPSHOT)?/
+    fileReplacerStr: release.aar
```
AbhiPrasad, line 1033:
```suggestion
targets:
  - name: maven
    gradleCliPath: ./gradlew
    mavenCliPath: scripts/mvnw.cmd
    mavenSettingsPath: scripts/settings.xml
    mavenRepoId: ossrh
    mavenRepoUrl: https://oss.sonatype.org/service/local/staging/deploy/maven2/
    android:
      distDirRegex: /^sentry-android-.*$/
      fileReplaceeRegex: /\d\.\d\.\d(-SNAPSHOT)?/
      fileReplacerStr: release.aar
```

README.md
```diff
@@ -1019,18 +1020,40 @@ The `android` structure contains the following options:
...
+```
+
+### Symbol Collector (`symbol-collector`)
+
+Using the [`symbol-collector`](https://github.com/getsentry/symbol-collector) client, uploads native symbols.
+The client needs to be available in the path.
```
AbhiPrasad, line 1039:
```suggestion
The `symbol-collector` client needs to be available in the path.
```

src/targets/symbolCollector.ts
```diff
@@ -0,0 +1,110 @@
...
+  private getSymbolCollectorConfig(): SymbolCollectorTargetConfig {
+    // The Symbol Collector should be available in the path
+    checkExecutableIsPresent(SYM_COLLECTOR_BIN_NAME);
+
+    if (!this.config.batchType || !this.config.bundleIdPrefix) {
+      throw new ConfigurationError(
+        'Required configuration not found in configuration file. ' +
+          'See the documentation for more details.'
+      );
+    }
```
AbhiPrasad, line 52: I would rather do a check for each and the error message include what config option is missing - makes it easier to debug.

src/targets/__tests__/symbolCollector.test.ts
```diff
@@ -0,0 +1,146 @@
...
+  customConfig?: Record<string, unknown>
+): SymbolCollector {
+  const config = customConfig
+    ? customConfig
+    : {
+        ['testKey']: 'testVal',
+      };
+  return new SymbolCollector(
+    {
+      name: 'aws-lambda-layer',
```
AbhiPrasad, line 35:
```suggestion
      name: 'symbol-collector',
```

src/targets/__tests__/symbolCollector.test.ts
```diff
@@ -0,0 +1,146 @@
...
+  };
+});
+
+const customConfig = {
+  batchType: 'batchType',
+  bundleIdPrefix: 'bundleIdPrefix-',
+};
+
+function getSymbolCollectorInstance(
+  customConfig?: Record<string, unknown>
```
AbhiPrasad, line 26: Setting a default value on the arg will mean we don't need the ternary expression below.

```suggestion
  customConfig: Record<string, unknown> = { testKey: 'testVal' } 
```

src/targets/symbolCollector.ts
```diff
@@ -0,0 +1,110 @@
...
+  batchType: string;
+  /** Prefix of the bundle ID to be uploaded. */
+  bundleIdPrefix: string;
+}
+
+export class SymbolCollector extends BaseTarget {
+  /** Target name */
+  public readonly name: string = 'symbol-collector';
+  /** Target options */
+  public readonly symbolCollectorConfig: SymbolCollectorTargetConfig;
```
AbhiPrasad, line 33: This seems like this should be private

AbhiPrasad: Let's also then not make tests directly assert on this object as well, feels fragile to me.

src/targets/__tests__/symbolCollector.test.ts
```diff
@@ -0,0 +1,146 @@
...
+    expect(
+      symCollector.artifactProvider.downloadArtifact
+    ).toHaveBeenCalledTimes(mockedArtifacts.length);
+
+    expect(spawnProcess).toHaveBeenCalledTimes(1);
+    const [cmd, args] = (spawnProcess as jest.MockedFunction<
+      typeof spawnProcess
+    >).mock.calls[0] as string[];
+    expect(cmd).toBe(SYM_COLLECTOR_BIN_NAME);
+    expect(args).toMatchInlineSnapshot(`
```
AbhiPrasad, line 131: Why inline snapshot over just asserting on the array?

src/utils/system.ts
```diff
@@ -273,6 +273,18 @@ function isExecutable(filePath: string): boolean {
   }
 }
 
+export function makeExecutable(filePath: string): boolean {
```
AbhiPrasad, line 276: Who uses this?

src/utils/githubApi.ts
```diff
@@ -69,6 +72,44 @@ export class GithubRemote {
         : '';
     return this.PROTOCOL_PREFIX + authData + this.GITHUB_HOSTNAME + this.url;
   }
+
+  public async getLatestRelease(): Promise<any> {
```
AbhiPrasad, line 76: Why have these methods been added to this file?

iker-barriocanal: Added them in the past and totally forgot to remove them lol.

src/utils/system.ts
iker-barriocanal: Same here, forgot to remove them.

Dockerfile
```diff
@@ -8,31 +8,39 @@ ENV DEBIAN_FRONTEND=noninteractive \
...
   && curl -fsSL https://packages.microsoft.com/config/debian/10/packages-microsoft-prod.deb -o /tmp/packages-microsoft-prod.deb \
   && dpkg -i /tmp/packages-microsoft-prod.deb \
   && rm /tmp/packages-microsoft-prod.deb \
   && echo 'deb [arch=amd64] https://download.docker.com/linux/debian buster stable' >> /etc/apt/sources.list \
   && curl -fsSL https://download.docker.com/linux/debian/gpg | apt-key add - \
   && apt-get update -qq \
   && apt-get install -y --no-install-recommends \
-    dotnet-sdk-3.1 \
-    docker-ce-cli \
+  dotnet-sdk-3.1 \
```
bruno-garcia, line 28: Given the image has .NET already (assuming it's because of `nuget` target), I can have a build of SymbolCollector that's 'runtime-dependant' which should reduce its size to just a few KBs.

Also, can we bump to 5.0?

```suggestion
  dotnet-sdk-5.0 \
```

Dockerfile
bruno-garcia: I think we need to go with 1.2.1 for now (as you found, there's an issue since the .NET 5 upgrade on symbol collector native CLI builds:

```suggestion
  && symbol_collector_url=$(curl -s https://api.github.com/repos/getsentry/symbol-collector/releases/tags/1.2.1 | \
```

README.md
```diff
@@ -1019,18 +1020,40 @@ The `android` structure contains the following options:
...
+
+**Configuration**
+
+| Option           | Description                                                                                  |
+| ---------------- | -------------------------------------------------------------------------------------------- |
+| `serverEndpoint` | **optional** The server endpoint. Defaults to `https://symbol-collector.services.sentry.io`. |
+| `batchType`      | The batch type of the symbols to be uploaded.                                                |
```
bruno-garcia, line 1046:
```suggestion
| `batchType`      | The batch type of the symbols to be uploaded. I.e: `Android`, `macOS`, `iOS`.                                                |
```

bruno-garcia (commented):
sweet! Left a couple suggestions. Thanks @iker-barriocanal

Dockerfile
```diff
@@ -8,31 +8,39 @@ ENV DEBIAN_FRONTEND=noninteractive \
...
+  apt-transport-https \
+  build-essential \
+  curl \
+  dirmngr \
+  gnupg \
+  git \
+  ruby-full \
+  twine \
+  jq \
+  unzip \
```
BYK, line 20: Let's bring back the indent.

Dockerfile
```diff
@@ -8,31 +8,39 @@ ENV DEBIAN_FRONTEND=noninteractive \
...
   && curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s --  --profile minimal -y \
   && cargo --version \
   && cargo install cargo-hack \
   # Stick with 3.1.x as 3.2.x doesn't install on Debian Buster for some reason
   && gem update --no-document --system 3.1.5 \
-  && gem install cocoapods
+  && gem install cocoapods \
+  # Install https://github.com/getsentry/symbol-collector
+  && symbol_collector_url=$(curl -s https://api.github.com/repos/getsentry/symbol-collector/releases/tags/1.3.1 | \
+  jq -r '.assets[].browser_download_url | select(endswith("symbolcollector-console-linux-x64.zip"))') \
```
BYK, line 40: I don't know how much overhead `jq` brings but we can replace this with some `awk` and `grep` magic if we ever need to 😀

Dockerfile
```diff
@@ -8,31 +8,39 @@ ENV DEBIAN_FRONTEND=noninteractive \
...
+  && curl -sL $symbol_collector_url -o "/tmp/sym-collector.zip" \
+  && unzip /tmp/sym-collector.zip -d /usr/local/bin/ \
```
BYK, line 42: I think you can just pipe this into `unzip` and avoid storing on disk?

src/targets/symbolCollector.ts
BYK: So much this.

src/targets/symbolCollector.ts
```diff
@@ -0,0 +1,110 @@
...
+    if (!this.config.batchType || !this.config.bundleIdPrefix) {
+      throw new ConfigurationError(
+        'Required configuration not found in configuration file. ' +
+          'See the documentation for more details.'
+      );
+    }
+
+    return {
+      serverEndpoint:
+        this.config.serverEndpoint || DEFAULT_SYM_COLLECTOR_SERVER_ENDPOINT,
```
BYK, line 56: Should this really be configurable? Also, we should probably not call this an endpoint but just a URL.

src/targets/symbolCollector.ts
```diff
@@ -0,0 +1,110 @@
...
+      batchType: this.config.batchType,
+      bundleIdPrefix: this.config.bundleIdPrefix,
+    };
+  }
+
+  public async publish(version: string, revision: string): Promise<any> {
+    const bundleId = this.symbolCollectorConfig.bundleIdPrefix + `${version}`;
```
BYK, line 63: ??
```suggestion
    const bundleId = this.symbolCollectorConfig.bundleIdPrefix + version;
```

src/targets/symbolCollector.ts
```diff
@@ -0,0 +1,110 @@
...
+    this.logger.debug('Fetching artifacts...');
```
BYK, line 64: You don't need this, `this.getArtifactsForRevision` already produces debug logs.

src/targets/symbolCollector.ts
```diff
@@ -0,0 +1,110 @@
...
+    const artifacts = await this.getArtifactsForRevision(revision, {
+      includeNames:
+        this.config.includeNames === undefined
+          ? undefined
+          : stringToRegexp(this.config.includeNames),
+    });
```
BYK, line 70: This is the second place we write the exact same code. I think this code should be moved into `getArtifactsForRevision` itself.

src/targets/symbolCollector.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    if (artifacts.length == 0) {
```
BYK, line 72:
```suggestion
    if (artifacts.length === 0) {
```

src/targets/symbolCollector.ts
```diff
@@ -0,0 +1,110 @@
...
+      this.logger.info(`Didn't found any artifacts after filtering`);
```
BYK, line 73: Shouldn't this be a warning?
```suggestion
      this.logger.warn(`Didn't found any artifacts after filtering`);
```

src/targets/symbolCollector.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    await withTempDir(async dir => {
+      // Download all artifacts in the same parent directory, where the symbol
+      // collector will recursively look for and deal with them.
+      // Since there are files with the same name, download them in different
+      // directories.
+      this.logger.debug('Downloading artifacts...');
+      await Promise.all(
+        artifacts.map(async (artifact, index) => {
+          const subdirPath = join(dir, index + '');
```
BYK, line 87: Don't be gimmicky please :)
```suggestion
          const subdirPath = join(dir, String(index));
```

Also, instead of using the index, can we not use `artifact.name` or something?

src/targets/symbolCollector.ts
```diff
@@ -0,0 +1,110 @@
...
+          dir,
+          '--batch-type',
+          this.symbolCollectorConfig.batchType,
+          '--bundle-id',
+          bundleId,
+          '--server-endpoint',
+          this.symbolCollectorConfig.serverEndpoint,
+        ])
+      )?.toString();
+      this.logger.debug('Process output: ', processOutput);
```
BYK, line 107: `spawnProcess` should be doing this already.

src/targets/__tests__/symbolCollector.test.ts
```diff
@@ -0,0 +1,146 @@
...
+      async cb => await cb('tmpDir')
+    );
+    (spawnProcess as jest.MockedFunction<
+      typeof spawnProcess
+    >).mockImplementation(() => Promise.resolve(undefined));
+
+    const mockedArtifacts = [
+      { filename: 'artifact1', storedFile: { lastUpdated: 'tomorrow' } },
+      { filename: 'artifact2', storedFile: { lastUpdated: 'next week' } },
+      { filename: 'artifact2', storedFile: { lastUpdated: 'in 5 years' } },
```
BYK, line 110: Are you sure these `lastUpdated` strings are valid? I think they should be ISO-8601 date strings?

src/targets/__tests__/symbolCollector.test.ts
BYK: Snapshots are much easier to manage 🙂

BYK (approved):
Didn't look too deep into tests but overall LGTM, great job!

src/targets/symbolCollector.ts
AbhiPrasad: We should turn on https://eslint.org/docs/rules/eqeqeq in craft

README.md
iker-barriocanal: Added this, but it'd be best to have that documentation somewhere else and point there. Updating Craft's docs because the symbol collector got e.g. a new feature and now supports a new type doesn't seem ideal.

Dockerfile
iker-barriocanal: `unzip` has a bug:

> Archives read from standard input are not yet supported, except with `funzip` (and then only the first member of the archive can be extracted).

`funzip` has some limitations:

> and it extracts the first member from the archive to stdout

(note the _first member_ and _to stdout_.)

I haven't found any alternatives _without_ storing on disk and directly piping, so suggestions are welcomed.

src/targets/symbolCollector.ts
iker-barriocanal: Yes, this should be configurable. The reason why this is called endpoint (and not just URL or any other way), is because the option in the symbol collector is `server-endpoint`. I think it'd be misleading to not be consistent here.

src/targets/symbolCollector.ts
iker-barriocanal: Do you have any suggestions to generate these unique names in a cheap way, without using the artifact names (and indexes), and before downloading the files?

src/targets/symbolCollector.ts
iker-barriocanal: All the configs in the targets are public, and that's the reason why this is too. If we're going to make them private, I think we should make them all private and not have some that are public and some others that aren't.

> Let's also then not make tests directly assert on this object as well, feels fragile to me.

Any suggestions to test whether the target has the required config parameters? By looking at the codebase, I don't think having getters is better; if we're changing the conventions, we shouldn't do it for a single target. Wdyt?

AbhiPrasad: Ok we should match the conventions - disregard my thoughts then. Maybe something to keep in mind if we decide to refactor the targets in the future.

In terms of the tests, I always prefer checking if the behaviour enabled by different config options is working rather than asserting on the config option itself. Asserting on resulting behaviour produces stronger tests and also matches real life usage much better - don’t find much value from just expecting the config property to be a certain value.

Dockerfile
iker-barriocanal: Opened another PR for that: https://github.com/getsentry/craft/pull/267

src/artifact_providers/base.ts
```diff
@@ -61,14 +62,47 @@ export interface LocalArtifact extends AbstractArtifact {
...
 }
+
+/**
+ * Returns parsed the given raw filters.
+ */
+export function parseFilterOptions(
+  rawFilters: RawFilterOptions
+): ParsedFilterOptions {
+  const parsedFilters: ParsedFilterOptions = {};
+  if (typeof rawFilters.includeNames !== 'undefined') {
```
AbhiPrasad, line 91: Do we need the undefined check if the properties (ex. `includesNames`) for `ParsedFilterOptions` are typed as optional?

src/artifact_providers/base.ts
```diff
@@ -61,14 +62,47 @@ export interface LocalArtifact extends AbstractArtifact {
 }
 
 /**
- * Fitlering options for artifacts
+ * Raw filtering options for artifacts
  */
-export interface FilterOptions {
+export interface RawFilterOptions {
```
AbhiPrasad, line 67: Please add a note that the raw filter options should be processed by `parseFilterOptions`.

src/targets/__tests__/symbolCollector.test.ts
```diff
@@ -0,0 +1,138 @@
...
+  };
+});
+
+const customConfig = {
+  batchType: 'batchType',
+  bundleIdPrefix: 'bundleIdPrefix-',
+};
+
+function getSymbolCollectorInstance(
+  config: Record<string, unknown> = { ['testKey']: 'testVal' }
```
AbhiPrasad, line 26: We don't need to use an explicit string key.

```suggestion
  config: Record<string, unknown> = { testKey: 'testVal' }
```

src/targets/symbolCollector.ts
iker-barriocanal: The behavior is either throwing an error (because the config wasn't correct) or successfully running the `publish` step. Not throwing an error is covered (there are tests with incorrect config expecting it to throw errors), and the `publish` step has its own tests too. Is there something that I'm missing?

AbhiPrasad (approved):
Nice! I would wait for Burak's answer about `unzip` before merging, but otherwise looks solid

Dockerfile
BYK: Oh, I wasn't aware of that. I guess I was thinking of `tar` as that's the more common format across Linux :)

src/artifact_providers/base.ts
```diff
@@ -61,14 +62,48 @@ export interface LocalArtifact extends AbstractArtifact {
...
+  /** Include files that match this pattern */
+  includeNames?: RegExp | string;
+  /** Exclude files that match this pattern */
+  excludeNames?: RegExp | string;
+}
+
+/**
+ * Parsed filtering options for artifacts
+ */
+export interface ParsedFilterOptions {
```
BYK, line 78: ❤️

src/artifact_providers/base.ts
```diff
@@ -61,14 +62,48 @@ export interface LocalArtifact extends AbstractArtifact {
...
   /** Include files that match this regexp */
   includeNames?: RegExp;
   /** Exclude files that match this regexp */
   excludeNames?: RegExp;
 }
+
+/**
+ * Returns parsed the given raw filters.
+ */
+export function parseFilterOptions(
```
BYK, line 88: ❤️ ❤️

src/artifact_providers/__tests__/base.test.ts
```diff
@@ -0,0 +1,41 @@
...
+
+describe('parseFilterOptions', () => {
+  test('empty object', () => {
+    const rawFilters: RawFilterOptions = {};
+    const parsedFilters = parseFilterOptions(rawFilters);
+    expect(parsedFilters).not.toHaveProperty('includeNames');
+    expect(parsedFilters).not.toHaveProperty('excludeNames');
+  });
+
+  test('undefined properties', () => {
```
BYK, line 12: I think we should also have one undefined one string combination.

Moreover, you can probably use `test.each` to clean up some of these cases. This is optional but the additional test cases is something we should definitely add.

src/targets/symbolCollector.ts
BYK: As much as I agree with @AbhiPrasad in a general sense, I think for this specific use case, this is fine. As long as the config is readonly, it is good enough for now.

src/targets/symbolCollector.ts
```diff
@@ -0,0 +1,108 @@
...
+  public readonly name: string = 'symbol-collector';
+  /** Target options */
+  public readonly symbolCollectorConfig: SymbolCollectorTargetConfig;
+
+  public constructor(
+    config: TargetConfig,
+    artifactProvider: BaseArtifactProvider
+  ) {
+    super(config, artifactProvider);
+    this.symbolCollectorConfig = this.getSymbolCollectorConfig();
```
BYK, line 39: This seems like a common pattern so maybe we should abstract config parsing into the base class and make sure it is always called by the default constructor `parseTargetSpecificConfig()`

src/targets/symbolCollector.ts
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  private getSymbolCollectorConfig(): SymbolCollectorTargetConfig {
+    // The Symbol Collector should be available in the path
+    checkExecutableIsPresent(SYM_COLLECTOR_BIN_NAME);
```
BYK, line 44: This is not really checking a config unless you make the bin name configurable through an env variable or the config file itself.

src/targets/symbolCollector.ts
```diff
@@ -0,0 +1,108 @@
...
+        this.config.serverEndpoint || DEFAULT_SYM_COLLECTOR_SERVER_ENDPOINT,
+      batchType: this.config.batchType,
+      bundleIdPrefix: this.config.bundleIdPrefix,
+    };
+  }
+
+  public async publish(version: string, revision: string): Promise<any> {
+    const bundleId = this.symbolCollectorConfig.bundleIdPrefix + version;
+    const artifacts = await this.getArtifactsForRevision(revision, {
+      includeNames: this.config.includeNames,
```
BYK, line 70: Why not pass `excludeNames` too?

src/targets/symbolCollector.ts
```diff
@@ -0,0 +1,108 @@
...
+    });
+
+    if (artifacts.length === 0) {
+      this.logger.warn(`Didn't found any artifacts after filtering`);
+      return;
+    }
+
+    this.logger.debug(`Found ${artifacts.length} symbol artifacts.`);
```
BYK, line 78: Should this logic be part of the base target class?

src/targets/symbolCollector.ts
```diff
@@ -0,0 +1,108 @@
...
+      // Download all artifacts in the same parent directory, where the symbol
+      // collector will recursively look for and deal with them.
+      // Since there are files with the same name, download them in different
+      // directories.
+      this.logger.debug('Downloading artifacts...');
+      await Promise.all(
+        artifacts.map(async (artifact, index) => {
+          const subdirPath = join(dir, String(index));
+          await fsPromises.mkdir(subdirPath);
+          await this.artifactProvider.downloadArtifact(artifact, subdirPath);
```
BYK, line 90: Might be better to just return the final promise:

```js
return this.artifactProvider.downloadArtifact(artifact, subdirPath);
```

src/targets/__tests__/symbolCollector.test.ts
```diff
@@ -0,0 +1,138 @@
+import { withTempDir } from '../../utils/files';
+import { NoneArtifactProvider } from '../../artifact_providers/none';
+import { checkExecutableIsPresent, spawnProcess } from '../../utils/system';
+import { SymbolCollector, SYM_COLLECTOR_BIN_NAME } from '../symbolCollector';
+
+jest.mock('../../utils/files');
+jest.mock('../../utils/system');
+jest.mock('fs', () => {
```
BYK, line 8: We may wanna lift the common `fs` mocks into a global mock module as this pattern seems common enough now.

src/targets/__tests__/symbolCollector.test.ts
```diff
@@ -0,0 +1,138 @@
...
+    new NoneArtifactProvider()
+  );
+}
+
+describe('target config', () => {
+  test('symbol collector not present in path', () => {
+    (checkExecutableIsPresent as jest.MockedFunction<
+      typeof checkExecutableIsPresent
+    >).mockImplementationOnce(() => {
+      throw new Error('Checked for executable');
```
BYK, line 42: Would have been better to make the executable path configurable and make sure we actually throw on a missing file.

src/targets/__tests__/symbolCollector.test.ts
```diff
@@ -0,0 +1,138 @@
...
+    >) = jest.fn();
+
+    expect(getSymbolCollectorInstance).toThrowErrorMatchingInlineSnapshot(
+      '"The required `batchType` parameter is missing in the configuration file. ' +
+        'See the documentation for more details."'
+    );
+    expect(checkExecutableIsPresent).toHaveBeenCalledTimes(1);
+    expect(checkExecutableIsPresent).toHaveBeenCalledWith(
+      SYM_COLLECTOR_BIN_NAME
+    );
```
BYK, line 66: I'd lift these checks into their own case as this doesn't seem relevant to the case we are testing.

PR #268, opened 2021-07-13
ref(symbol-collector): Verbose output of command

iker-barriocanal:
Be verbose on the output of the symbol collector call.

### DIFF
--- a/src/targets/symbolCollector.ts
+++ b/src/targets/symbolCollector.ts
@@ -91,18 +91,22 @@ export class SymbolCollector extends BaseTarget {
         })
       );
 
-      await spawnProcess(SYM_COLLECTOR_BIN_NAME, [
-        '--upload',
-        'directory',
-        '--path',
-        dir,
-        '--batch-type',
-        this.symbolCollectorConfig.batchType,
-        '--bundle-id',
-        bundleId,
-        '--server-endpoint',
-        this.symbolCollectorConfig.serverEndpoint,
-      ]);
+      this.logger.info(
+        (
+          await spawnProcess(SYM_COLLECTOR_BIN_NAME, [
+            '--upload',
+            'directory',
+            '--path',
+            dir,
+            '--batch-type',
+            this.symbolCollectorConfig.batchType,
+            '--bundle-id',
+            bundleId,
+            '--server-endpoint',
+            this.symbolCollectorConfig.serverEndpoint,
+          ])
+        )?.toString()
+      );
     });
   }
 }
--- a/src/targets/symbolCollector.ts
+++ b/src/targets/symbolCollector.ts
@@ -91,22 +91,24 @@ export class SymbolCollector extends BaseTarget {
         })
       );
 
-      this.logger.info(
-        (
-          await spawnProcess(SYM_COLLECTOR_BIN_NAME, [
-            '--upload',
-            'directory',
-            '--path',
-            dir,
-            '--batch-type',
-            this.symbolCollectorConfig.batchType,
-            '--bundle-id',
-            bundleId,
-            '--server-endpoint',
-            this.symbolCollectorConfig.serverEndpoint,
-          ])
-        )?.toString()
-      );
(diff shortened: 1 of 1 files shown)


src/targets/symbolCollector.ts
```diff
@@ -91,18 +91,22 @@ export class SymbolCollector extends BaseTarget {
...
+            dir,
+            '--batch-type',
+            this.symbolCollectorConfig.batchType,
+            '--bundle-id',
+            bundleId,
+            '--server-endpoint',
+            this.symbolCollectorConfig.serverEndpoint,
+          ])
+        )?.toString()
+      );
```
kamilogorek, line 109:
```suggestion
      const output = await spawnProcess(SYM_COLLECTOR_BIN_NAME, [
        '--upload',
        'directory',
        '--path',
        dir,
        '--batch-type',
        this.symbolCollectorConfig.batchType,
        '--bundle-id',
        bundleId,
        '--server-endpoint',
        this.symbolCollectorConfig.serverEndpoint,
      ]);
      
      this.logger.info(output?.toString());
```

I think this way will be more obvious what `toString` and `?` refer to.

kamilogorek (changes requested):

src/targets/symbolCollector.ts
```diff
@@ -91,18 +91,22 @@ export class SymbolCollector extends BaseTarget {
...
-        '--batch-type',
-        this.symbolCollectorConfig.batchType,
-        '--bundle-id',
-        bundleId,
-        '--server-endpoint',
-        this.symbolCollectorConfig.serverEndpoint,
-      ]);
+      this.logger.info(
+        (
+          await spawnProcess(SYM_COLLECTOR_BIN_NAME, [
```
AbhiPrasad, line 96: Can we make sure we extract into a variable - IMO makes it easier to read.

```
const stdOut = await spawnProcess(...)
if (stdOut) {
  // we don't wanna log out undefined
  this.logger.info(stdOut); 
} else {
  // idk how to handle this case
}
```

Also from what I can see from https://github.com/getsentry/craft/blob/master/src/utils/system.ts#L127, the promise always resolves to a buffer, never undefined. Why is it typed that way?

AbhiPrasad: Oh wait sorry didn't see the early return case.

AbhiPrasad: Then we can make it:

```ts
const stdOut = await spawnProcess(...)
if (stdOut) {
  this.logger.info(stdOut); 
}
```

src/targets/symbolCollector.ts
```diff
@@ -103,6 +103,12 @@ export class SymbolCollector extends BaseTarget {
         '--server-endpoint',
         this.symbolCollectorConfig.serverEndpoint,
       ]);
+
+      if (cmdOutput) {
+        this.logger.info(cmdOutput.toString());
+      } else {
+        this.logger.info(`The command didn't have any output.`);
```
AbhiPrasad, line 110: Thinking about this more, cmdOutput returns either a `Buffer` or `undefined`. `undefined` is only returned in dry run mode, otherwise if the command has no output it'll return an empty array (`[]`).

Therefore we shouldn't log anything out in the else condition. We should check instead for an empty array to say that the command didn't have any output.

AbhiPrasad (approved):

kamilogorek (approved):

src/targets/symbolCollector.ts
```diff
@@ -103,6 +103,14 @@ export class SymbolCollector extends BaseTarget {
...
         this.symbolCollectorConfig.serverEndpoint,
       ]);
+
+      if (cmdOutput) {
+        if (cmdOutput.length === 0) {
+          this.logger.info(`The command didn't have any output.`);
+        } else {
+          this.logger.info('Command output:\n', cmdOutput.toString());
+        }
+      }
```
BYK, line 113: Feels like these should be `debug` instead of `info`?

PR #270, opened 2021-07-22
ref(maven): Support BOM files in `maven` target

iker-barriocanal:
Add support for BOM files in the `maven` target.

This is based on [previous Kotlin code](https://github.com/getsentry/sentry-java/blob/bf425d511746782ee51a9ac475fa353997d6f34e/scripts/release.kts), but that's incomplete. A BOM file is a POM file with a `<packaging>pom</packaging>` key.

An artifact (the `zip` downloaded from GHA) may contain a POM and a BOM, both with the same name (`pom-default.xml` is the default name). In case both are present, unzipping the artifact may lead to having an unknown name for the BOM file (e.g. the POM is the `pom-default.xml`, and the BOM is ??). Thus, if the default file isn't a BOM, all `xml` files in the artifact are checked to not miss the BOM. If no BOM is found, the distribution is uploaded as it is now (without this PR); and if a BOM is found, a different command to upload files is run. Once artifacts have been uploaded, the registry is closed.

### DIFF
--- a/src/targets/__tests__/maven.test.ts
+++ b/src/targets/__tests__/maven.test.ts
@@ -151,7 +151,7 @@ describe('publish', () => {
     mvnTarget.artifactProvider.downloadArtifact = jest
       .fn()
       .mockResolvedValueOnce('artifact/download/path');
-    mvnTarget.isBomFile = jest.fn().mockResolvedValueOnce(undefined);
+    mvnTarget.isBomFile = jest.fn().mockResolvedValueOnce(false);
 
     await mvnTarget.upload('r3v1s10n');
 
--- a/src/targets/__tests__/maven.test.ts
+++ b/src/targets/__tests__/maven.test.ts
@@ -1,7 +1,12 @@
 import { homedir } from 'os';
 import { join } from 'path';
 import { NoneArtifactProvider } from '../../artifact_providers/none';
-import { MavenTarget, targetOptions, targetSecrets } from '../maven';
+import {
+  MavenTarget,
+  POM_DEFAULT_FILENAME,
+  targetOptions,
+  targetSecrets,
+} from '../maven';
 import { retrySpawnProcess } from '../../utils/async';
 import { withTempDir } from '../../utils/files';
 
@@ -11,6 +16,11 @@ jest.mock('fs', () => ({
   ...jest.requireActual('fs'),
   promises: {
     writeFile: jest.fn(() => Promise.resolve()),
+    readFile: jest.fn((file: string) => file),
+    readdir: async () => Promise.resolve([]), // empty dir
+    access: jest.fn().mockImplementation(() => {
+      // do nothing
+    }),
   },
 }));
 
@@ -109,10 +119,8 @@ describe('publish', () => {
       async () => void callOrder.push('closeAndRelease')
     );
 
-    const version = '1.0.0';
     const revision = 'r3v1s10n';
-
-    await mvnTarget.publish(version, revision);
+    await mvnTarget.publish('1.0.0', revision);
     expect(gradlePropsMock).toHaveBeenCalledTimes(1);
     expect(uploadMock).toHaveBeenCalledTimes(1);
     expect(uploadMock).toHaveBeenLastCalledWith(revision);
@@ -127,7 +135,7 @@ describe('publish', () => {
     ]);
   });
 
-  test('upload', async () => {
+  test('upload POM', async () => {
     // simple mock to always use the same temporary directory,
     // instead of creating a new one
     (withTempDir as jest.MockedFunction<typeof withTempDir>).mockImplementation(
@@ -143,9 +151,11 @@ describe('publish', () => {
     mvnTarget.artifactProvider.downloadArtifact = jest
       .fn()
       .mockResolvedValueOnce('artifact/download/path');
+    mvnTarget.isBomFile = jest.fn().mockResolvedValueOnce(undefined);
 
     await mvnTarget.upload('r3v1s10n');
-    expect(retrySpawnProcess).toBeCalledTimes(1);
+
+    expect(retrySpawnProcess).toHaveBeenCalledTimes(1);
     const callArgs = (retrySpawnProcess as jest.MockedFunction<
       typeof retrySpawnProcess
     >).mock.calls[0];
@@ -171,6 +181,49 @@ describe('publish', () => {
     expect(cmdArgs[7]).toBe(`-Durl=${DEFAULT_OPTION_VALUE}`);
     expect(cmdArgs[8]).toBe('--settings');
     expect(cmdArgs[9]).toBe(DEFAULT_OPTION_VALUE);
+  });
+
+  test('upload BOM', async () => {
+    // simple mock to always use the same temporary directory,
+    // instead of creating a new one
+    (withTempDir as jest.MockedFunction<typeof withTempDir>).mockImplementation(
+      async cb => {
+        return await cb(tmpDirName);
+      }
+    );
+
+    const mvnTarget = createMavenTarget();
+    mvnTarget.getArtifactsForRevision = jest
+      .fn()
+      .mockResolvedValueOnce([{ filename: 'mockArtifact.zip' }]);
+    mvnTarget.artifactProvider.downloadArtifact = jest
+      .fn()
+      .mockResolvedValueOnce('artifact/download/path');
+    mvnTarget.isBomFile = jest.fn().mockResolvedValueOnce('path/to/bomfile');
+
+    await mvnTarget.upload('r3v1s10n');
+
+    expect(retrySpawnProcess).toHaveBeenCalledTimes(1);
+    const callArgs = (retrySpawnProcess as jest.MockedFunction<
+      typeof retrySpawnProcess
+    >).mock.calls[0];
+
+    expect(callArgs).toHaveLength(2);
+    expect(callArgs[0]).toEqual(DEFAULT_OPTION_VALUE);
+
+    const cmdArgs = callArgs[1] as string[];
+    expect(cmdArgs).toHaveLength(7);
+    expect(cmdArgs[0]).toBe('gpg:sign-and-deploy-file');
+    expect(cmdArgs[1]).toMatch(
+      new RegExp(`-Dfile=${tmpDirName}.+${POM_DEFAULT_FILENAME}`)
+    );
+    expect(cmdArgs[2]).toMatch(
+      new RegExp(`-DpomFile=${tmpDirName}.*${POM_DEFAULT_FILENAME}`)
+    );
+    expect(cmdArgs[3]).toBe(`-DrepositoryId=${DEFAULT_OPTION_VALUE}`);
+    expect(cmdArgs[4]).toBe(`-Durl=${DEFAULT_OPTION_VALUE}`);
+    expect(cmdArgs[5]).toBe('--settings');
+    expect(cmdArgs[6]).toBe(DEFAULT_OPTION_VALUE);
   });
 });
 
--- a/src/targets/maven.ts
+++ b/src/targets/maven.ts
@@ -22,7 +22,7 @@ const GRADLE_PROPERTIES_FILENAME = 'gradle.properties';
  */
 const DEFAULT_GRADLE_USER_HOME = join(homedir(), '.gradle');
 export const POM_DEFAULT_FILENAME = 'pom-default.xml';
-const POM_FILE_EXTNAME = '.xml'; // Must include the leading `.`
+const POM_FILE_EXT = '.xml'; // Must include the leading `.`
 const BOM_FILE_KEY_REGEXP = stringToRegexp('/<packaging>pom</packaging>/');
 
 export const targetSecrets = ['OSSRH_USERNAME', 'OSSRH_PASSWORD'] as const;
@@ -271,6 +271,10 @@ export class MavenTarget extends BaseTarget {
     }
   }
 
+  /**
+   * Returns the path to the BOM file in the given distribution directory, and
+   * `undefined` if there isn't any.
+   */
   private async getBomFileFomDist(
     distDir: string
   ): Promise<string | undefined> {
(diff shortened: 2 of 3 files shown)


src/targets/__tests__/maven.test.ts
```diff
@@ -109,10 +119,8 @@ describe('publish', () => {
       async () => void callOrder.push('closeAndRelease')
     );
 
-    const version = '1.0.0';
     const revision = 'r3v1s10n';
-
-    await mvnTarget.publish(version, revision);
+    await mvnTarget.publish('1.0.0', revision);
```
AbhiPrasad, line 123: Why change to directly passing in the string vs. a variable like before?

src/targets/maven.ts
```diff
@@ -21,6 +21,9 @@ const GRADLE_PROPERTIES_FILENAME = 'gradle.properties';
  * https://docs.gradle.org/current/userguide/build_environment.html#sec:gradle_environment_variables
  */
 const DEFAULT_GRADLE_USER_HOME = join(homedir(), '.gradle');
+export const POM_DEFAULT_FILENAME = 'pom-default.xml';
+const POM_FILE_EXTNAME = '.xml'; // Must include the leading `.`
```
AbhiPrasad, line 25: nit: name is redundant
```suggestion
const POM_FILE_EXT = '.xml'; // Must include the leading `.`
```

src/targets/maven.ts
```diff
@@ -258,12 +261,92 @@ export class MavenTarget extends BaseTarget {
...
+      );
+      return false;
+    }
+  }
+
+  private async uploadBomDistribution(bomFile: string): Promise<void> {
+    await retrySpawnProcess(this.mavenConfig.mavenCliPath, [
+      'gpg:sign-and-deploy-file',
+      `-Dfile=${bomFile}`,
+      `-DpomFile=${bomFile}`,
```
AbhiPrasad, line 335: don't need to set `-Dtypes=jar,jar` here?

src/targets/maven.ts
```diff
@@ -258,12 +261,92 @@ export class MavenTarget extends BaseTarget {
...
+        encoding: 'utf8',
+      });
+      const matchesRequiredKey = BOM_FILE_KEY_REGEXP.test(fileContents);
+      if (matchesRequiredKey) {
+        return true;
+      }
+      return false;
+    } catch (error) {
+      this.logger.warn(
+        `Error checking whether it's a BOM file: ${pomFilepath}\n`,
```
AbhiPrasad, line 324:
```suggestion
        `Could not determine if file path is BOM file: ${pomFilepath}\nError:\n`,
```

src/targets/maven.ts
```diff
@@ -258,12 +261,92 @@ export class MavenTarget extends BaseTarget {
...
+    try {
+      await fsPromises.access(pomFilepath, fsConstants.R_OK);
+      const fileContents = await fsPromises.readFile(pomFilepath, {
+        encoding: 'utf8',
+      });
+      const matchesRequiredKey = BOM_FILE_KEY_REGEXP.test(fileContents);
+      if (matchesRequiredKey) {
+        return true;
+      }
+      return false;
```
AbhiPrasad, line 321:
```suggestion
      return BOM_FILE_KEY_REGEXP.test(fileContents);
```

src/targets/maven.ts
```diff
@@ -258,12 +261,92 @@ export class MavenTarget extends BaseTarget {
...
+    // There may be several files in the ZIP-ed artifact with the same name,
+    // where the BOM may be one of them (there may not be a BOM). Files may be
+    // renamed when extracting the ZIP, so the default name (`pom-default.xml`)
+    // may not match. It's assumed that any renaming keeps the same extension,
+    // so all files with the same extension are checked to identify the BOM.
+    const filesInDir = await fsPromises.readdir(distDir);
+    const potentialPoms = filesInDir
+      .filter(f => extname(f) === POM_FILE_EXTNAME)
+      .filter(f => f !== POM_DEFAULT_FILENAME)
+      .map(f => join(distDir, f));
```
AbhiPrasad, line 291: If we wanna be fancy we can do this in a single `reduce` pass, but I like the explicit filter + map.

src/targets/maven.ts
```diff
@@ -258,12 +261,92 @@ export class MavenTarget extends BaseTarget {
...
+    if (bomFile) {
+      this.logger.debug('Found BOM: ', bomFile);
+      await this.uploadBomDistribution(bomFile);
+    } else {
+      this.logger.debug('Did not find a BOM.');
+      await this.uploadPomDistribution(distDir);
+    }
+  }
+
+  private async getBomFileFomDist(
```
AbhiPrasad, line 274: Would appreciate a doc string here.

AbhiPrasad (changes requested):

src/targets/__tests__/maven.test.ts
iker-barriocanal: The version isn't being used anywhere else, so IMO removes a bit of noise (like in others `upload` calls). Do you think it's easier to read by leaving the variable?

src/targets/maven.ts
iker-barriocanal: No, see https://github.com/getsentry/sentry-java/blob/bf425d511746782ee51a9ac475fa353997d6f34e/scripts/release.kts#L64-L70.

src/targets/maven.ts
iker-barriocanal: Yep, IMO it's easier to read.

src/targets/__tests__/maven.test.ts
AbhiPrasad: No I have no strong opinions, just wondering. I agree it helps reduce noise in the test file so I would keep it 👍

src/targets/__tests__/maven.test.ts
```diff
@@ -143,9 +151,11 @@ describe('publish', () => {
     mvnTarget.artifactProvider.downloadArtifact = jest
       .fn()
       .mockResolvedValueOnce('artifact/download/path');
+    mvnTarget.isBomFile = jest.fn().mockResolvedValueOnce(undefined);
```
AbhiPrasad, line 154: `isBomFile` returns `Promise<boolean>`
```suggestion
    mvnTarget.isBomFile = jest.fn().mockResolvedValueOnce(false);
```

src/targets/maven.ts
```diff
@@ -258,12 +261,91 @@ export class MavenTarget extends BaseTarget {
...
+      await this.uploadPomDistribution(distDir);
+    }
+  }
+
+  /**
+   * Returns the path to the BOM file in the given distribution directory, and
+   * `undefined` if there isn't any.
+   */
+  private async getBomFileInDist(distDir: string): Promise<string | undefined> {
+    const pomFilepath = join(distDir, POM_DEFAULT_FILENAME);
```
AbhiPrasad, line 279: Should this be something like `possibleBomFilePath`?

src/targets/maven.ts
```diff
@@ -258,12 +261,91 @@ export class MavenTarget extends BaseTarget {
...
+    try {
+      await fsPromises.access(pomFilepath, fsConstants.R_OK);
+      const fileContents = await fsPromises.readFile(pomFilepath, {
+        encoding: 'utf8',
+      });
+      return BOM_FILE_KEY_REGEXP.test(fileContents);
+    } catch (error) {
+      this.logger.warn(
+        `Could not determine if path corresponds to a BOM file: ${pomFilepath}\n` +
+          `Error:\n`,
```
AbhiPrasad, line 323:
```suggestion
          'Error:\n',
```

src/targets/maven.ts
```diff
@@ -258,12 +261,91 @@ export class MavenTarget extends BaseTarget {
...
+    const pomFilepath = join(distDir, POM_DEFAULT_FILENAME);
+    if (await this.isBomFile(pomFilepath)) {
+      return pomFilepath;
+    }
+
+    // There may be several files in the ZIP-ed artifact with the same name,
+    // where the BOM may be one of them (there may not be a BOM). Files may be
+    // renamed when extracting the ZIP, so the default name (`pom-default.xml`)
+    // may not match. It's assumed that any renaming keeps the same extension,
+    // so all files with the same extension are checked to identify the BOM.
```
AbhiPrasad, line 288: Can we simplify this logic by enforcing users of the target to pass in the bom file name as a craft option? I know we already wrote all of this out, but I'm scared of the regex test failing to account for all scenarios.

AbhiPrasad (commented):
Just wanted to get the comment about the zip behaviour address, and then I'll slam the ✅.

src/targets/maven.ts
iker-barriocanal: It could, but not sure whether it _should_. We don't know whether it's a BOM, but we do know it's a POM (a BOM is a type of POM). So I think making a reference to BOM makes it less accurate; besides, some logic below treats any `xml` file as a potential BOM, so that would be confusing.

AbhiPrasad: Got it, that reasoning sounds good to me.

src/targets/maven.ts
iker-barriocanal: No. Although the end goal is to run the target in unattended mode, it currently is meant to run in attended mode. When files with same name exist in a ZIP, OSX (where this is going to run for now, at least until running in unattended mode is fully supported) asks for a name to rename the file. Force the user to set the same filename set in the config file is a clear no for me.

Regarding other scenarios, you're right this isn't solid enough. The Linux file system (where the Craft image will run), OSX, and running in attended mode is a lot of scenarios. The thing that should never change is the file extension (even Linux shouldn't modify it), and that's the reason behind it. The little prose in the PR description is ambiguous, but I left it like that on purpose. Once the `maven` target gets more work (to support unattended mode), these scenarios should be handled.

I'll add a `TODO` in the code for this, which clearly missing.

AbhiPrasad (approved):
Thanks for the additional context + TODO 🚀

src/targets/__tests__/maven.test.ts
```diff
@@ -11,6 +16,11 @@ jest.mock('fs', () => ({
   ...jest.requireActual('fs'),
   promises: {
     writeFile: jest.fn(() => Promise.resolve()),
+    readFile: jest.fn((file: string) => file),
+    readdir: async () => Promise.resolve([]), // empty dir
+    access: jest.fn().mockImplementation(() => {
+      // do nothing
+    }),
```
BYK, line 23: `jest.fn()` should have been sufficient here?

src/targets/maven.ts
```diff
@@ -21,6 +21,9 @@ const GRADLE_PROPERTIES_FILENAME = 'gradle.properties';
  * https://docs.gradle.org/current/userguide/build_environment.html#sec:gradle_environment_variables
  */
 const DEFAULT_GRADLE_USER_HOME = join(homedir(), '.gradle');
+export const POM_DEFAULT_FILENAME = 'pom-default.xml';
+const POM_FILE_EXT = '.xml'; // Must include the leading `.`
+const BOM_FILE_KEY_REGEXP = stringToRegexp('/<packaging>pom</packaging>/');
```
BYK, line 26: Why use `stringToRegexp` here where you can simply write a regex???

```js
/<packaging>pom<\/packaging>/
```

FYI, it is also unsafe to try parsing an XML file with a RegEx.

src/targets/maven.ts
```diff
@@ -258,12 +261,95 @@ export class MavenTarget extends BaseTarget {
...
+   */
+  private async getBomFileInDist(distDir: string): Promise<string | undefined> {
+    const pomFilepath = join(distDir, POM_DEFAULT_FILENAME);
+    if (await this.isBomFile(pomFilepath)) {
+      return pomFilepath;
+    }
+
+    // There may be several files in the ZIP-ed artifact with the same name,
+    // where the BOM may be one of them (there may not be a BOM). Files may be
+    // renamed when extracting the ZIP, so the default name (`pom-default.xml`)
```
BYK, line 286: >Files may be renamed when extracting the ZIP

How so? _We_ control how we unzip, right?

src/targets/maven.ts
```diff
@@ -258,12 +261,95 @@ export class MavenTarget extends BaseTarget {
...
+    // may not match. It's assumed that any renaming keeps the same extension,
+    // so all files with the same extension are checked to identify the BOM.
+    // TODO: make sure all scenarios are considered and tested.
+    // Each file system may handle this case differently, and attended vs
+    // unattended mode also have different behaviours. It's not desired to get
+    // the BOM renamed in such a way that isn't handled by the target.
+    const filesInDir = await fsPromises.readdir(distDir);
+    const potentialPoms = filesInDir
+      .filter(f => extname(f) === POM_FILE_EXT)
+      .filter(f => f !== POM_DEFAULT_FILENAME)
```
BYK, line 296: Why not do these in a single `filter()` call?

```js
.filter(f => f !== POM_DEFAULT_FILENAME && extname(f) === POM_FILE_EXT)
```

src/targets/maven.ts
```diff
@@ -258,12 +261,95 @@ export class MavenTarget extends BaseTarget {
...
+      .map(f => join(distDir, f));
+
+    for (const f of potentialPoms) {
+      if (await this.isBomFile(f)) {
+        return f;
+      }
+    }
```
BYK, line 303: Strongly recommend using sync APIs for `isBomFile` and replacing this block with `potentialPoms.find(f => this.isBomFile(f))` if the number of files is small and the file contents are small. Would be much faster and simpler code.

src/targets/maven.ts
```diff
@@ -258,12 +261,95 @@ export class MavenTarget extends BaseTarget {
...
+   *
+   * A BOM file is a POM file with the following key:
+   * `<packaging>pom</packaging>`, usually named as `pom-default.xml`.
+   *
+   * @param pomFilepath path to the POM.
+   * @returns true if the POM is a BOM.
+   */
+  public async isBomFile(pomFilepath: string): Promise<boolean> {
+    try {
+      await fsPromises.access(pomFilepath, fsConstants.R_OK);
```
BYK, line 319: Why do we check this explicitly? `fsPromises.readFile` would throw an exception if we cannot read the file anyway?

src/targets/maven.ts
iker-barriocanal: Not sure how big the size of the contents can be, but went with the declarative way.

PR #271, opened 2021-07-22
feat(maven): Optional `android` config

iker-barriocanal:
The `maven` target is responsible for releases related to Maven. Currently, the Android configuration is required, but there are scenarios where it isn't used (e.g. https://github.com/getsentry/sentry-android-gradle-plugin). This PR makes the android configuration optional.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,8 @@
 
 - feat(maven): Add maven target to deploy to Maven Central (#258)
 - feat(symbol-collector): Add symbol-collector target (#266)
-- ref(maven): Support BOM files in `maven` target (#270)
-- ref(maven): Optional `android` config (#271)
+- feat(maven): Support BOM files in `maven` target (#270)
+- feat(maven): Optional `android` config (#271)
 
 ## 0.24.4
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - feat(maven): Add maven target to deploy to Maven Central (#258)
 - feat(symbol-collector): Add symbol-collector target (#266)
 - ref(maven): Support BOM files in `maven` target (#270)
+- ref(maven): Optional `android` config (#271)
 
 ## 0.24.4
 
--- a/src/targets/__tests__/maven.test.ts
+++ b/src/targets/__tests__/maven.test.ts
@@ -46,7 +46,7 @@ function removeTargetSecretsFromEnv(): void {
   }
 }
 
-function getAllTargetOptions() {
+function getFullTargetConfig(): any {
   return {
     OSSRH_USERNAME: DEFAULT_OPTION_VALUE,
     OSSRH_PASSWORD: DEFAULT_OPTION_VALUE,
@@ -63,7 +63,7 @@ function getAllTargetOptions() {
   };
 }
 
-function getRequiredTargetOptions() {
+function getRequiredTargetConfig(): any {
   return {
     OSSRH_USERNAME: DEFAULT_OPTION_VALUE,
     OSSRH_PASSWORD: DEFAULT_OPTION_VALUE,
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - feat(maven): Add maven target to deploy to Maven Central (#258)
 - feat(symbol-collector): Add symbol-collector target (#266)
 - ref(maven): Support BOM files in `maven` target (#270)
+- ref(maven): Optional `android` config (#271)
```
BYK, line 8: These are both new features, not refactors as refactors should not change existing behavior (thus, not documented in changelog 😄 )

```suggestion
- feat(maven): Support BOM files in `maven` target (#270)
- feat(maven): Optional `android` config (#271)
```

src/targets/maven.ts
```diff
@@ -401,14 +415,18 @@ export class MavenTarget extends BaseTarget {
...
+      );
+      if (isAndroidDistDir) {
+        return moduleName.replace(
           this.mavenConfig.android.fileReplaceeRegex,
           this.mavenConfig.android.fileReplacerStr
-        )
-      : `${moduleName}.jar`;
+        );
+      }
+    }
```
BYK, line 428: You should probably print a warning here according to your updated docs?

marandaneto:
it prints `Did not find a BOM` for every package, maybe let's change to: `found a BOM` when it's a BOM, since the majority of the packages aren't a BOM, this gives the impression that something is off, but it's the correct behavior.

src/targets/maven.ts
```diff
@@ -81,10 +81,24 @@ export class MavenTarget extends BaseTarget {
...
+      return {
+        ...targetSecrets,
+        ...outerTargetSettings,
+        ...androidSettings,
+      };
+    }
+
+    this.logger.warn(
+      'You may need the Android configuration and it was not found in the configuration file.'
+    );
```
marandaneto, line 98: not sure if this should be a `warn` level since the `android` config is optional.

marandaneto:
@iker-barriocanal @BYK there are a few comments here but the PR does whats supposed to do, feel free to address them, or not and merge and release Craft, I can confirm that it works :)

marandaneto (approved):

marandaneto:
> it prints `Did not find a BOM` for every package, maybe let's change to: `found a BOM` when it's a BOM, since the majority of the packages aren't a BOM, this gives the impression that something is off, but it's the correct behavior.

btw it already prints `found BOM`, so just removing `Did not find a BOM` or lowering log level would be fine.

src/targets/maven.ts
iker-barriocanal: There's a warning when checking for the android config ([line 96](https://github.com/getsentry/craft/blob/e4dc758b79cb9ce6b05ecfca9a80037036f4f3c9/src/targets/maven.ts#L96-L98)), I find it better than doing it for every distribution. Do you think there should be another warning here?

src/targets/maven.ts
iker-barriocanal: I find anything below a `warn` to not be enough when an `android` release may file because the config is missing.

src/targets/maven.ts
BYK: Nah, that's alright. Thanks for the info! 🙂

src/targets/maven.ts
BYK: Then this should be `info` or even less and then we should have a hard-fail when this is required but missing.

src/targets/__tests__/maven.test.ts
```diff
@@ -61,23 +61,108 @@ function getTargetOptions() {
...
+
+  test('no env vars and no options', () => {
+    removeTargetSecretsFromEnv();
+    expect(createMavenTarget).toThrowErrorMatchingInlineSnapshot(
+      `"Required value(s) OSSRH_USERNAME not found in configuration files or the environment. See the documentation for more details."`
+    );
+  });
+
+  test('env vars without options', () => {
+    expect(createMavenTarget.bind(this, {})).toThrowErrorMatchingInlineSnapshot(
```
BYK, line 101:
```suggestion
    expect(() => createMavenTarget({})).toThrowErrorMatchingInlineSnapshot(
```

src/targets/__tests__/maven.test.ts
```diff
@@ -61,23 +61,108 @@ function getTargetOptions() {
...
+      `"Required configuration gradleCliPath not found in configuration file. See the documentation for more details."`
+    );
+  });
+
+  test('no android config', () => {
+    const config = getRequiredTargetConfig();
+    delete config.android;
+    expect(
+      createMavenTarget.bind(this, config)
```
BYK, line 110:
```suggestion
      () => createMavenTarget(config)
```

src/targets/__tests__/maven.test.ts
```diff
@@ -61,23 +61,108 @@ function getTargetOptions() {
...
+    ).toThrowErrorMatchingInlineSnapshot(
+      `"Required Android configuration is incorrect or was not found in the configuration file. See the documentation for more details."`
```
BYK, line 112: Was it incorrect or not found? Why cannot be more specific?

src/targets/__tests__/maven.test.ts
```diff
@@ -61,23 +61,108 @@ function getTargetOptions() {
...
+    );
+  });
+
+  test('incorrect one-line android config', () => {
+    const config = getRequiredTargetConfig();
+    config.android = 'yes';
+    expect(
+      createMavenTarget.bind(this, config)
```
BYK, line 120:
```suggestion
      () => createMavenTarget(config)
```

src/targets/__tests__/maven.test.ts
```diff
@@ -61,23 +61,108 @@ function getTargetOptions() {
...
+      `"Required Android configuration is incorrect or was not found in the configuration file. See the documentation for more details."`
+    );
+  });
+
+  test('correct object android config, with additional props', () => {
+    const config = getFullTargetConfig();
+    config.android.additionalProp = 'not relevant';
+    const mvnTarget = createMavenTarget(config);
+    const androidConfig: any = mvnTarget.mavenConfig.android;
+    expect(androidConfig.distDirRegex).toBeDefined();
```
BYK, line 158: Why not https://jestjs.io/docs/expect#tomatchobjectobject here?

src/targets/__tests__/maven.test.ts
```diff
@@ -87,10 +172,26 @@ describe('Maven target configuration', () => {
...
+    targetOptions.map(secret =>
+      expect(mvnTarget.config).toEqual(
+        expect.objectContaining({
+          [secret]: DEFAULT_OPTION_VALUE,
+        })
+      )
+    );
+    expect(mvnTarget.config.android.distDirRegex).toStrictEqual(
+      expect.any(String)
+    );
```
BYK, line 187:
```suggestion
    expect(typeof mvnTarget.config.android.distDirRegex).toBe(
      'string'
    );
```

BYK:
```suggestion
    expect(mvnTarget.config.android.distDirRegex).toBeInstanceOf(
      String
    );
```

BYK (approved):
Definitely an improvement over the existing state so approving. That said I have some questions, concerns, and suggestions.

src/targets/__tests__/maven.test.ts
```diff
@@ -61,23 +61,108 @@ function getTargetOptions() {
...
+function getRequiredTargetConfig(): any {
+  return {
+    OSSRH_USERNAME: DEFAULT_OPTION_VALUE,
+    OSSRH_PASSWORD: DEFAULT_OPTION_VALUE,
+    gradleCliPath: DEFAULT_OPTION_VALUE,
+    mavenCliPath: DEFAULT_OPTION_VALUE,
+    mavenSettingsPath: DEFAULT_OPTION_VALUE,
+    mavenRepoId: DEFAULT_OPTION_VALUE,
+    mavenRepoUrl: DEFAULT_OPTION_VALUE,
+    android: false,
```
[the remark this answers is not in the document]
iker-barriocanal: My thinking about what this config field should answer is "I don't want to configure Android".

- I think of `null` as "there's no Android config". Why isn't there an Android config? Have you forgotten about it? This, to me, is like not answering a question.
- I think of `undefined` as "the Android config hasn't been defined". Why don't you define it then, if it's a config file? Do you want to set an `android` config or not? This, to me, is like giving a very ambiguous answer to a question.
- I think of `false` as "I don't want to set an Android config, but I'm saying it". This, to me, is like saying you don't want to set the configuration (answering with a nonambiguous answer).

Thus, using `false` seems the most appropriate for me. Do you have a different opinion about any (or all) of them? What do you think? Is it a best practice to use another type? (I might have thought about the field like `isAndroid` too.)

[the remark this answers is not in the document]
BYK: I understand and mostly agree with the reasoning. But requiring to be explicit can lead to unnecessary frustration or repetition, especially in configurations like these. I'm almost thinking about it might be better to split `android` and `maven` targets and make them share a common core to address your concerns here.

BYK (approved):
Looks good with the updates so unblocking.

I'm still not sold on the "explicitly set to `false`" idea though. If I only care about Java packages w/o Android, why should I need to _know_ (or worse, guess) to set `android` to false? Or does it even make sense setting `android: false` when it is actually required?

PR #275, opened 2021-08-17
ref(maven): Improve logging for `maven` target

iker-barriocanal:
- Don't log when not finding BOMs. Most of the packages aren't supposed to be BOM files, and logging this may give the impression that something is missing when it's not.
- Add a note to the target's docs to show command output when the log level is `trace`.

### DIFF
--- a/src/targets/maven.ts
+++ b/src/targets/maven.ts
@@ -13,7 +13,6 @@ import { withTempDir } from '../utils/files';
 import { ConfigurationError } from '../utils/errors';
 import { stringToRegexp } from '../utils/filters';
 import { checkEnvForPrerequisite } from '../utils/env';
-import { LogLevel } from 'consola';
 
 const GRADLE_PROPERTIES_FILENAME = 'gradle.properties';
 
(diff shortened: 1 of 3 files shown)


src/targets/maven.ts
```diff
@@ -411,4 +412,14 @@ export class MavenTarget extends BaseTarget {
         )
       : `${moduleName}.jar`;
   }
+
+  private logCmdOutput(output: Buffer | undefined): void {
+    if (output) {
+      if (output.length === 0) {
```
AbhiPrasad, line 418: We can combine this conditional.

src/targets/maven.ts
```diff
@@ -411,4 +412,14 @@ export class MavenTarget extends BaseTarget {
...
+        this.logger.debug(`The command didn't have any output.`);
```
AbhiPrasad, line 419: Hard to know what the command is from this statement, would be helpful to have some details about the command

src/targets/maven.ts
```diff
@@ -411,4 +412,14 @@ export class MavenTarget extends BaseTarget {
         )
       : `${moduleName}.jar`;
   }
+
+  private logCmdOutput(output: Buffer | undefined): void {
```
BYK, line 416: You shouldn't need any of this as `spawnProcess` does it underneath:

https://github.com/getsentry/craft/blob/a015c79de78bbc672570a38cb2177786e2fa93db/src/utils/system.ts#L174-L182

BYK (changes requested):

src/targets/maven.ts
```diff
@@ -331,16 +332,20 @@ export class MavenTarget extends BaseTarget {
...
+        'gpg:sign-and-deploy-file',
+        `-Dfile=${bomFile}`,
+        `-DpomFile=${bomFile}`,
+        `-DrepositoryId=${this.mavenConfig.mavenRepoId}`,
+        `-Durl=${this.mavenConfig.mavenRepoUrl}`,
+        '--settings',
+        this.mavenConfig.mavenSettingsPath,
+      ],
+      {},
+      { showStdout: this.logger.level >= LogLevel.Debug }
```
BYK, line 347: Why are we doing this?

BYK (changes requested):
I don't understand the motivation behind the `this.logger.level >= LogLevel.Debug` condition.

`spawnProcess` already logs with level `info` if this is `true` at all times. If we want to see the output for debugging purposes, we can always set the logging level to `trace` for that specific session, right?

BYK (approved):

Issue #272, opened 2021-07-28
Improve logging for the Maven target


marandaneto:
log the output of the maven CLI and gradlew CLI calls if `--log-level=debug` is enabled, so we can debug things if something goes wrong.

Craft also writes a file (~/.gradle/gradle.properties) with the exported env. vars `OSSRH_USERNAME` and `OSSRH_PASSWORD` values, required by Gradle, I suggest erasing these values (`mavenCentralUsername` and `mavenCentralPassword`) after Craft is executed, so we don't leak the user and pass.

there are also a few comments on https://github.com/getsentry/craft/pull/271

iker-barriocanal:
Logging has been improved on https://github.com/getsentry/craft/pull/275 and `~/.gradle/gradle.properties` is removed since https://github.com/getsentry/craft/pull/276.

PR #276, opened 2021-08-17
feat(maven): Remove Gradle properties after publishing

iker-barriocanal:
Removing the `gradle.properties` after running the maven target may prevent exposing it, which contains the OSSRH credentials (see https://github.com/getsentry/craft/issues/272). There also are minor changes to address some feedback in https://github.com/getsentry/craft/pull/270.

### DIFF
--- a/src/targets/__tests__/maven.test.ts
+++ b/src/targets/__tests__/maven.test.ts
@@ -18,6 +18,7 @@ jest.mock('fs', () => ({
     writeFile: jest.fn(() => Promise.resolve()),
     readFile: jest.fn((file: string) => file),
     readdir: async () => Promise.resolve([]), // empty dir
+    unlink: jest.fn(),
   },
 }));
 
@@ -104,10 +105,14 @@ describe('publish', () => {
   test('main flow', async () => {
     const callOrder: string[] = [];
     const mvnTarget = createMavenTarget();
-    const gradlePropsMock = jest.fn(
-      async () => void callOrder.push('gradleProps')
+    const createGradlePropsMock = jest.fn(
+      async () => void callOrder.push('createGradleProps')
     );
-    mvnTarget.createUserGradlePropsFile = gradlePropsMock;
+    mvnTarget.createUserGradlePropsFile = createGradlePropsMock;
+    const deleteGradlePropsMock = jest.fn(
+      async () => void callOrder.push('deleteGradleProps')
+    );
+    mvnTarget.deleteUserGradlePropsFile = deleteGradlePropsMock;
     const uploadMock = jest.fn(async () => void callOrder.push('upload'));
     mvnTarget.upload = uploadMock;
     (retrySpawnProcess as jest.MockedFunction<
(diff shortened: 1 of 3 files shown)


src/targets/maven.ts
```diff
@@ -292,17 +302,10 @@ export class MavenTarget extends BaseTarget {
...
       .map(f => join(distDir, f));
 
-    for (const f of potentialPoms) {
-      if (await this.isBomFile(f)) {
-        return f;
-      }
-    }
-
-    return undefined;
+    return potentialPoms.find(async f => await this.isBomFile(f));
```
BYK, line 308:
```suggestion
    return potentialPoms.find(f => this.isBomFile(f));
```

src/targets/maven.ts
```diff
@@ -205,6 +206,15 @@ export class MavenTarget extends BaseTarget {
     );
   }
 
+  /**
+   * Deletes the user's `gradle.properties` file.
+   */
+  public deleteUserGradlePropsFile(): Promise<void> {
+    return fsPromises.unlink(
+      join(this.getGradleHomeDir(), GRADLE_PROPERTIES_FILENAME)
```
BYK, line 214: Who creates this file? If it is the `closeAndReleaseRepository` script, then shouldn't this be its responsibility (to remove the file I mean)

BYK (changes requested):

src/targets/__tests__/maven.test.ts
```diff
@@ -121,17 +123,18 @@ describe('publish', () => {
 
     const revision = 'r3v1s10n';
     await mvnTarget.publish('1.0.0', revision);
-    expect(gradlePropsMock).toHaveBeenCalledTimes(1);
+    expect(createGradlePropsMock).toHaveBeenCalledTimes(1);
     expect(uploadMock).toHaveBeenCalledTimes(1);
     expect(uploadMock).toHaveBeenLastCalledWith(revision);
     expect(retrySpawnProcess).toHaveBeenCalledTimes(1);
     expect(retrySpawnProcess).toHaveBeenCalledWith(DEFAULT_OPTION_VALUE, [
```
BYK, line 130: Shouldn't we also add a check for `deleteGradleProps` to ensure it is deleting the correct file?

src/targets/maven.ts
BYK: Then at this point, I'd convert this into something like `withTempDir` with auto-clean up even in case of exceptions for safety. This is much like a [Python Context Manager](https://realpython.com/python-with-statement/#using-the-python-with-statement)

BYK (changes requested):
Pushing back for even safer code 😀

BYK (approved):
Postponing the context manager idea to a follow-up. Looks good as it is, thanks a lot!

marandaneto:
@iker-barriocanal my suggestion was not really to remove the file, but erase the content of the written properties, this is a global file that has also users configuration, if you delete the file, I will have to copy/paste all my user configuration again

BYK:
@marandaneto I think we can incorporate this feedback into the planned follow-up: use a context manager-like function which keeps a snapshot of the old file around and restores it back once the operation finishes.

iker-barriocanal:
@marandaneto if that user configuration is relevant for this target, the target itself should be able to get it in some way. When this gets moved to GHA anything that the target isn't able to get won't be in that config file.

marandaneto:
@iker-barriocanal no, its not relevant for the target, but the file is a global configuration file for Gradle itself, so it could be from different projects etc, eg:

a=1
b=2
c=3

now you add
mavenCentralUsername=user
mavenCentralPassword=pass

so:

a=1
b=2
c=3
mavenCentralUsername=user
mavenCentralPassword=pass

when you delete the file, my projects in need of `a, b and c` fails.

@BYK suggestion of keeping a snapshot of the file is great, restoring to the previous state, I'd rather say it's a severe bug.

PR #278, opened 2021-08-24
fix(maven): Snapshot gradle props instead of deleting

iker-barriocanal:
From https://github.com/getsentry/craft/pull/276 the gradle properties file was removed, causing the loss of any user config that may exist in that file. This PR makes a snapshot of the file at the beginning of the `maven` target execution and restores it at the end, so the user config shouldn't be affected.

### DIFF
--- a/src/targets/__tests__/maven.test.ts
+++ b/src/targets/__tests__/maven.test.ts
@@ -19,6 +19,8 @@ jest.mock('fs', () => ({
     readFile: jest.fn((file: string) => file),
     readdir: async () => Promise.resolve([]), // empty dir
     unlink: jest.fn(),
+    access: jest.fn(),
+    copyFile: jest.fn(),
   },
 }));
 
@@ -188,16 +190,22 @@ describe('publish', () => {
   beforeEach(() => jest.resetAllMocks());
 
   test('main flow', async () => {
+    (withTempDir as jest.MockedFunction<typeof withTempDir>).mockImplementation(
+      async cb => {
+        return await cb(tmpDirName);
+      }
+    );
+
     const callOrder: string[] = [];
     const mvnTarget = createMavenTarget();
-    const createGradlePropsMock = jest.fn(
-      async () => void callOrder.push('createGradleProps')
+    const makeSnapshotMock = jest.fn(
+      async () => void callOrder.push('makeSnapshot')
     );
-    mvnTarget.createUserGradlePropsFile = createGradlePropsMock;
-    const deleteGradlePropsMock = jest.fn(
-      async () => void callOrder.push('deleteGradleProps')
+    mvnTarget.createUserGradlePropsFile = makeSnapshotMock;
+    const recoverGradlePropsSnapshot = jest.fn(
+      async () => void callOrder.push('recoverSnapshot')
     );
-    mvnTarget.deleteUserGradlePropsFile = deleteGradlePropsMock;
+    mvnTarget.recoverGradlePropsSnapshot = recoverGradlePropsSnapshot;
     const uploadMock = jest.fn(async () => void callOrder.push('upload'));
     mvnTarget.upload = uploadMock;
     (retrySpawnProcess as jest.MockedFunction<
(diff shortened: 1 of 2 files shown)


src/targets/maven.ts
```diff
@@ -187,31 +187,117 @@ export class MavenTarget extends BaseTarget {
    * @param revision Git commit SHA to be published.
    */
   public async publish(_version: string, revison: string): Promise<void> {
-    await this.createUserGradlePropsFile();
-    await this.upload(revison);
+    await withTempDir(async dir => {
+      const snapshotPath = await this.makeGradlePropsSnapshot(dir);
```
BYK, line 191: I mean you've come all this way, it is trivial to convert this into something like `withTempDir` that wraps around the actions you want to run safely.

src/targets/maven.ts
```diff
@@ -187,31 +187,117 @@ export class MavenTarget extends BaseTarget {
...
+   * @param snapshotDir Directory to place the snapshot in.
+   * @returns The path to the snapshot in the given directory,
+   *  or `undefined` if the snapshot wasn't created.
+   */
+  public async makeGradlePropsSnapshot(
+    snapshotDir: string
+  ): Promise<string | undefined> {
+    const propsExpectedPath = this.getGradlePropsPath();
+    try {
+      await fsPromises.access(propsExpectedPath, fsConstants.R_OK);
```
BYK, line 227: You don't need this step. Just move this `try`-`catch` case down to the `copyFile` call and it is fine. This way you'll save an IO op _and_ a thread cycle.

src/targets/maven.ts
```diff
@@ -187,31 +187,117 @@ export class MavenTarget extends BaseTarget {
...
 
   /**
-   * Creates the required user's `gradle.properties` file.
+   * Restores the gradle properties snapshot.
    *
-   * If there's an existing one, it's overwritten.
-   * TODO: control when it's overwritten with an option.
+   * @param snapshotPath Path to the snapshot of the gradle properties file.
+   */
+  public async recoverGradlePropsSnapshot(
```
BYK, line 262:
```suggestion
  public async restoreGradleProps(
```

src/targets/maven.ts
```diff
@@ -187,31 +187,117 @@ export class MavenTarget extends BaseTarget {
...
+    snapshotPath: string | undefined
+  ): Promise<void> {
+    if (!snapshotPath) {
+      this.logger.debug('Deleting temporary gradle properties file...');
+      return this.deleteUserGradlePropsFile();
+    }
+
+    const gradlePropsPath = this.getGradlePropsPath();
+    try {
+      await fsPromises.access(gradlePropsPath, fsConstants.W_OK);
```
BYK, line 272: Again, drop this check, it is useless.

BYK (changes requested):

src/targets/maven.ts
```diff
@@ -187,31 +193,140 @@ export class MavenTarget extends BaseTarget {
...
+  ): Promise<string | undefined> {
+    const propsExpectedPath = this.getGradlePropsPath();
+    try {
+      accessSync(propsExpectedPath, fsConstants.R_OK);
+    } catch (error) {
+      this.logger.debug(
+        'Did not find a gradle properties file in: ',
+        propsExpectedPath
+      );
+      return undefined;
```
BYK, line 264:
```suggestion
      return null;
```

`null` might be a better option here to signify intentional empty return value.

src/targets/maven.ts
```diff
@@ -187,31 +193,140 @@ export class MavenTarget extends BaseTarget {
...
+      await this.createUserGradlePropsFile();
+
+      try {
+        return await cb();
+      } finally {
+        if (snapshotPath) {
+          await this.restoreGradleProps(snapshotPath);
+        } else {
+          this.logger.debug('Deleting temporary gradle properties file...');
+          this.deleteUserGradlePropsFile();
```
BYK, line 240: This logic already exists in `restoreGradleProps()` so I think you should:

1. Remove the conditional here and always call `await this.restoreGradleProps(snapshotPath)`
2. Remove the helper function `deleteUserGradlePropsFile` and bake its code into `restoreGradleProps`

src/targets/maven.ts
```diff
@@ -57,6 +61,8 @@ export type MavenTargetConfig = Record<TargetSettingType, string> &
 
 type PartialTargetConfig = { name: string; value: string | undefined }[];
 
+type CallbackFunction = () => Promise<any>;
```
BYK, line 64:
```suggestion
type AsyncCallback = () => Promise<unknown>;
```

BYK (commented):
Context manager looks great and see how much it simplifies the code!

src/targets/__tests__/maven.test.ts
```diff
@@ -338,3 +348,92 @@ describe('get gradle home directory', () => {
...
+        );
+
+      await mvnTarget.publish('v3rs10n', 'r3v1s10n');
+      await expect(fsPromises.access(expectedPropsPath)).rejects.toThrowError(
+        /ENOENT: no such file/
+      );
+    });
+  });
+
+  test('existent props file', async () => {
```
BYK, line 405: You are missing some test cases:
1. What if the callback function throws an error? You should test that we still restore the snapshot/delete our own props file
2. What if we simply cannot backup an existing snapshot file? Where's the test case for that?
3. What if we cannot restore a snapshot file for some reason?

BYK (changes requested):

src/targets/__tests__/maven.test.ts
```diff
@@ -338,3 +348,92 @@ describe('get gradle home directory', () => {
...
+      const mvnTarget = createMavenTarget(getRequiredTargetConfig());
+      mvnTarget.upload = jest
+        .fn()
+        .mockImplementationOnce(
+          async () => await testCorrectPropsFile(expectedPropsPath)
+        );
+
+      await mvnTarget.publish('v3rs10n', 'r3v1s10n');
+      await expect(fsPromises.access(expectedPropsPath)).rejects.toThrowError(
+        /ENOENT: no such file/
```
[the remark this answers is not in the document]
iker-barriocanal: Because `to` and `be` don't exist. Moving to `accessSync` since it's cleaner, but if you know an even cleaner way I can update it.

kamilogorek:
Superseded by #351

PR #282, opened 2021-09-06
upgrade(docker): Upgrade to Debain Bullseye and Node 14

BYK:
Prerequisite for #281.


HazAT (approved):
Let's hope this doesn't break anything else xD

PR #285, opened 2021-09-14
feat(prepare): Pass old version to bump-version

BYK:
With this patch, we do a crude determination of the previous release,
assuming the first tag before current tip  the previous release.
This is rather naive as we allow custom formatting of git tags but it is
a good start nevertheless.


iker-barriocanal (approved):

iker-barriocanal:
Note that we shouldn't only consider custom formatting, since following the semver formatting there's also the chance to take unwanted tags (e.g. prereleases).

PR #287, opened 2021-09-14
feat(changelog): Auto generate changelog when empty

BYK:


src/utils/git.ts
```diff
@@ -4,6 +4,14 @@ import { getConfigFileDir } from '../config';
 import { ConfigurationError } from './errors';
 import { logger } from '../logger';
 
+export interface GitChange {
+  hash: string;
+  message: string;
+  pr: string | null;
+}
+
+export const PRExtractor = /(?<=\(#)\d+(?=\)$)/;
```
iker-barriocanal, line 13: TIL.

We should add to the README the expected format of the commit messages in `auto` mode, to include the PR ID.

src/commands/prepare.ts
```diff
@@ -384,6 +391,14 @@ async function prepareChangelog(
       }
       if (!changeset.body) {
         replaceSection = changeset.name;
+        const changes = await getChangesSince(git, oldVersion);
+        changeset.body = changes
+          .map(change =>
+            change.pr
+              ? `- ${change.message}`
+              : `- ${change.message} (${change.hash.slice(0, 8)})`
```
iker-barriocanal, line 399: Any reason to set 8?

src/commands/prepare.ts
```diff
@@ -336,6 +341,8 @@ async function execPublish(remote: string, newVersion: string): Promise<never> {
  * @param changelogPath Path to the changelog file
  */
 async function prepareChangelog(
+  git: SimpleGit,
+  oldVersion: string,
```
iker-barriocanal, line 345: We should add these to the docstrings. What is `oldVersion` expecting? Is it the raw version number, or the revision of that version? Does it require a tag?

src/utils/changes.ts
```diff
@@ -72,9 +73,11 @@ function locateChangeset(
   ) {
     const matchedTitle = match[2] || match[3];
     if (predicate(matchedTitle)) {
+      const padSize = match?.[1]?.length || 0;
       return {
         end: HEADER_REGEX.exec(markdown),
         start: match,
+        padding: new Array(padSize + 1).join(' '),
```
iker-barriocanal, line 80: 👍

iker-barriocanal (commented):
Should we add tests for the PR ID format in the commit messages?

src/commands/prepare.ts
BYK: That's the common length for short SHAs on GitHub 🙂

src/utils/git.ts
BYK: I don't think that's a good idea as GitHub does this automatically and I don't see a reason for anyone to modify that (unless they are using Phabricator etc. which disqualifies them for this anyway)

src/commands/prepare.ts
BYK: Yeah, good catch. Will update 👍🏻

BYK:
>Should we add tests for the PR ID format in the commit messages?

Thought about that but I think a better test would be to test the new changelog generation logic: make sure we have a PR id at the end, if not fallback to short SHA etc. For that I need to construct a sample/fake repo in tests.

That said since this feature is ephemeral and will evolve very quickly to use GitHub milestones, I think delaying tests for that next step makes more sense.

iker-barriocanal (approved):
🚀

src/commands/prepare.ts
BYK: Eh, you are right about getting confused. Git defaults to `7` but this is known to cause collisions with larger projects (such as Sentry) so AFAIK the default for GitHub is 8+ based on some StackOverflow research: https://stackoverflow.com/questions/18134627/how-much-of-a-git-sha-is-generally-considered-necessary-to-uniquely-identify-a

Will add a comment in a follow-up.

src/utils/git.ts
```diff
@@ -22,6 +30,23 @@ export async function getLatestTag(git: SimpleGit): Promise<string> {
...
 }
 
+export async function getChangesSince(
+  git: SimpleGit,
+  rev: string
+): Promise<GitChange[]> {
+  const { all: commits } = await git.log({
+    from: rev,
+    to: 'HEAD',
+    symmetric: false,
```
chadwhitacre, line 40: I don't really understand this option. Why are we using it here?

https://www.npmjs.com/package/simple-git#git-log

https://git-scm.com/docs/gitrevisions#_dotted_range_notations

src/utils/git.ts
chadwhitacre: Would be good to comment on the GitHub behavior we're depending on here.

src/utils/git.ts
BYK: This option defaults to `true`, giving you a "triple-dot range" which lists all commits from both parts whereas we only need the commits reachable from `HEAD` that don't exist on `rev`.

Makes sense? One is `A - B` and the other is more like `A XOR B`. We want the former.

src/commands/prepare.ts
```diff
@@ -384,6 +393,14 @@ async function prepareChangelog(
...
       if (!changeset.body) {
         replaceSection = changeset.name;
+        const changes = await getChangesSince(git, oldVersion);
+        changeset.body = changes
+          .map(change =>
+            change.pr
+              ? `- ${change.message}`
+              : `- ${change.message} (${change.hash.slice(0, 8)})`
+          )
+          .join('\n');
```
chadwhitacre, line 403: So previous behavior was to leave `changeset.body` empty, and the new behavior is to populate it with the commit message. If there are no commit messages for the commits in question then `changelog.body` will remain empty.

How are newlines in commit messages handled? Gracefully?

src/commands/prepare.ts
chadwhitacre: > Will add a comment in a follow-up.

Bump ☝️ 

Also, I think the pr branch logic would've been cleaner to encapsulate in `getChangeSince` since the PR extraction logic is also down there, and what we're basically doing here is defaulting to a SHA in the case that we can't find a PR.

CHANGELOG.md
```diff
@@ -1,15 +1,5 @@
...
-## Unreleased
-
-- feat(maven): Add maven target to deploy to Maven Central (#258)
-- feat(symbol-collector): Add symbol-collector target (#266)
-- feat(maven): Support BOM files in `maven` target (#270)
-- feat(maven): Optional `android` config (#271)
-- feat(maven): Remove Gradle properties after publishing (#276)
-- upgrade(docker): Upgrade to Debain Bullseye and Node 14 (#282)
-- fix(cocoapods): Add COCOAPODS_ALLOW_ROOT=1 to Docker (#281)
-
```
chadwhitacre, line 12: Did we test that this PR generates this output?

src/utils/changes.ts
```diff
@@ -23,6 +23,7 @@ export interface Changeset {
 export interface ChangesetLoc {
   start: RegExpExecArray;
   end: RegExpExecArray | null;
+  padding: string;
```
chadwhitacre, line 26: If I'm reading this all right, this is just a refactor, not a behavior change with regard to padding, correct?

CHANGELOG.md
BYK: Of course I did 😀 I does much better than this 😎

src/utils/changes.ts
BYK: Correct 🙂

src/commands/prepare.ts
BYK: >So previous behavior was to leave changeset.body empty,

Not really, we filled it with the dreaded "No documented changes" text.

> the new behavior is to populate it with the commit message.
> How are newlines in commit messages handled? Gracefully?

Sort of. We only use the title, the first line of the commit message so no need to handle new lines. `simple-git` is a bit weird with its jargon: it uses `message` for the title and `body` for the actual long message.

>If there are no commit messages for the commits in question then changelog.body will remain empty

I think this is an interesting edge case and I think should be handled. Will follow up.

BYK: >I think this is an interesting edge case and I think should be handled. Will follow up.

Okay, false alarm. Just checked the code and when we get an empty commit list, that results in a blank text which then triggers the default text fallback which still says "No documented changes.".

src/utils/git.ts
BYK: Added some comments: https://github.com/getsentry/craft/pull/289

src/utils/git.ts
BYK: Added comments here: https://github.com/getsentry/craft/pull/289

PR #288, opened 2021-09-15
ref: Fix TypeScript type warning regarding catch

BYK:
TypeScript now defaults to `unknown` for caught error types instead of `any` causing us to get warnings when trying to access some standard `Error` properties. This PR changes the code accordingly so it is both safer, cleaner, and does not have this warning anymore.

### DIFF
--- a/src/commands/publish.ts
+++ b/src/commands/publish.ts
@@ -319,10 +319,10 @@ async function checkRevisionStatus(
     logger.debug('Repository info received');
     logger.trace(repositoryInfo);
   } catch (e) {
-    reportError(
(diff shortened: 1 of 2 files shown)


iker-barriocanal (commented):
Don't we need to set either `--strict` or `--useUnknownInCatchVariables`? See https://www.typescriptlang.org/tsconfig#useUnknownInCatchVariables

iker-barriocanal (approved):

BYK:
>Don't we need to set either --strict or --useUnknownInCatchVariables? See https://typescriptlang.org/tsconfig#useUnknownInCatchVariables

Seems like this became the default since TS 4.4. Not sure what I'm getting but this doesn't break anything and also makes our code future-proof so I'd just land 😀

PR #289, opened 2021-09-17
ref: Add comments for changelog extraction

BYK:
Follow up to #287

### DIFF
--- a/src/utils/git.ts
+++ b/src/utils/git.ts
@@ -43,8 +43,8 @@ export async function getChangesSince(
   const { all: commits } = await git.log({
     from: rev,
     to: 'HEAD',
-    // The symmetric option defaults to true, giving us all the differen commits
-    // between `from` and `to` whereas what we are interested is only the ones
+    // The symmetric option defaults to true, giving us all the different commits
+    // between `from` and `to` whereas what we are interested in is only the ones
     // accessible from `to` so we get a "changelog" kind of list.
     // See https://github.com/steveukx/git-js#git-log and
     // https://git-scm.com/docs/gitrevisions#_dotted_range_notations for more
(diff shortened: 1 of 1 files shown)


src/utils/git.ts
```diff
@@ -37,6 +43,11 @@ export async function getChangesSince(
   const { all: commits } = await git.log({
     from: rev,
     to: 'HEAD',
+    // The symmetric option defaults to true, giving us all the differen commits
```
chadwhitacre, line 46:
```suggestion
    // The symmetric option defaults to true, giving us all the different commits
```

src/utils/git.ts
```diff
@@ -37,6 +43,11 @@ export async function getChangesSince(
...
+    // between `from` and `to` whereas what we are interested is only the ones
```
chadwhitacre, line 47:
```suggestion
    // between `from` and `to` whereas what we are interested in is only the ones
```

src/utils/git.ts
```diff
@@ -37,6 +43,11 @@ export async function getChangesSince(
   const { all: commits } = await git.log({
     from: rev,
     to: 'HEAD',
+    // The symmetric option defaults to true, giving us all the different commits
+    // between `from` and `to` whereas what we are interested in is only the ones
+    // accessible from `to` so we get a "changelog" kind of list.
+    // See https://github.com/steveukx/git-js#git-log and
+    // https://git-scm.com/docs/gitrevisions#_dotted_range_notations for more
```
chadwhitacre, line 50: > commits between `from` and `to` whereas what we are interested in is only the ones

Sorry to be pedantic about a comment but this is really throwing me off, I think "between" here might be ambiguous? How about this ... ?

```suggestion
    // The symmetric option defaults to true, giving us all the different commits
    // reachable from both `from` and `to` whereas what we are interested in is only the ones
    // reachable from `to` and _not_ from `from` so we get a "changelog" kind of list. One is `A - B` and the other is more like `A XOR B`. We want the former.
    // See https://github.com/steveukx/git-js#git-log and
    // https://git-scm.com/docs/gitrevisions#_dotted_range_notations for more
```

chadwhitacre (approved):

PR #290, opened 2021-09-17
upgrade(octokit): Upgrade octokit to v18

BYK:
This is needed so we can use the GraphQL endpoints.

List of breaking changes: https://github.com/octokit/rest.js/releases/tag/v18.0.0

All tests pass and also manually verified locally by preparing and releasing https://github.com/getsentry/release-tester/releases/0.18.0

### DIFF
--- a/src/artifact_providers/github.ts
+++ b/src/artifact_providers/github.ts
@@ -1,4 +1,4 @@
-import Github from '@octokit/rest';
+import { Octokit } from '@octokit/rest';
 import * as fs from 'fs';
 import fetch from 'node-fetch';
 import * as path from 'path';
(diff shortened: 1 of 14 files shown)


src/artifact_providers/github.ts
```diff
@@ -81,14 +64,14 @@ export class GithubArtifactProvider extends BaseArtifactProvider {
     const per_page = 100;
 
     // https://docs.github.com/en/free-pro-team@latest/rest/reference/actions#artifacts
```
iker-barriocanal, line 66: Since we're not using this format anymore:
```suggestion

```

iker-barriocanal (approved):
Awesome to have types for GitHub endpoint methods!

src/artifact_providers/github.ts
BYK: I mean it is still the same endpoint 😀 I think it is good to keep the reference.

PR #277, opened 2021-08-17
test(symcollector): Add tests for filter options

iker-barriocanal:
- Clean up tests for the base artifact provider.
- Add tests to cover more edge-cases, such as having strings, regex, and undefined on the same call.
- Other minor changes.

PR #263, opened 2021-07-01
docs: Mention release/** branches on README

marandaneto:

PR #291, opened 2021-09-20
feat(changelog): GitHub Milestones based changelogs

BYK:
This patch implements the GitHub Milestones based automatic
changelog generation. It builds up on #285 and #287 and ties
them with GitHub milestone data to generate a changelog that
is composed of logical sections with references to relevant
PRs.

Any remaining items are put under "Various fixes & improvements"
section as an itemized list.


BYK:
Changelog generated for this commit:

<hr>

## Automated Changelog Generation (ongoing)
                                                                                     
We now automatically generate changelog entries for the `auto` changelog policy where none provided, instead of saying "No documented changes". The commits/PRs are grouped by their associated GitHub milestones and the milestone title and description are used in the changelog along with a list of related commits/PRs. Any unaccounted changes are grouped under the "Various improvements and fixes" section.

PRs: #290, #289, #287, #285

## Added Maven Target (ongoing)

Added the long-awaited Maven target, full with Android support.

PRs: #271, #275, #276, #270, #258

## Added symbol-collector Target (ongoing)

Added target for our very own [symbol-collector](https://github.com/getsentry/symbol-collector/) to collect and upload all native system symbols with Craft.

PRs: #284, #277, #269, #268, #267, #266

## Fixed Cocoapods Support (ongoing)

Turns out our Cocoapods target was a bit outdated and broken. We have fixed it in this release! 🥳

PRs: #281, #282

## Various fixes & improvements

- feat(changelog): GitHub Milestones based changelogs (750a7797)
- ref: Fix TypeScript type warning regarding catch (#288)
- build(deps): Bump set-value from 3.0.2 to 4.1.0 (#286)
- ci: Drop Node 10 support, use Node 14 by default (#283)
- build(deps-dev): Bump tar from 4.4.15 to 4.4.18 (#280)
- build(deps): Bump path-parse from 1.0.6 to 1.0.7 (#274)
- build(deps-dev): Bump tar from 4.4.8 to 4.4.15 (#273)
- docs: Consistent code samples for shell (e84f693f)
- docs: Mention release/** branches on README (#263)

src/utils/changes.ts
```diff
@@ -175,3 +181,152 @@ export function prependChangeset(
 
   return markdown.slice(0, startIdx) + newSection + markdown.slice(startIdx);
 }
+
+interface Commit {
+  hash: string;
+  title: string;
+  hasPRinTitle: boolean;
+  pr: string | null;
```
iker-barriocanal, line 189: If there's a `pr` field, why do we have another field saying whether there's a PR? If there isn't, `pr === null`; if there is, `pr` will be a string, right?

src/utils/changes.ts
```diff
@@ -175,3 +181,152 @@ export function prependChangeset(
...
+  milestone: string | null;
+}
+
+interface Milestone {
+  title?: string;
+  description?: string;
+  state?: string;
```
iker-barriocanal, line 196: What are the possible `state` values? Are they unlimited? Can we type this, or link to the place where these are documented?

src/utils/changes.ts
```diff
@@ -175,3 +181,152 @@ export function prependChangeset(
...
+  prs: string[];
+}
+
+// This is set to 8 since GitHub and GitLab prefer that over the default 7 to
+// avoid collisions.
```
iker-barriocanal, line 201: 🙌

src/utils/changes.ts
```diff
@@ -175,3 +181,152 @@ export function prependChangeset(
...
+const SHORT_SHA_LENGTH = 8;
+function formatCommit(commit: Commit): string {
```
iker-barriocanal, line 203: This seems coupled to the commit interface. What do you think about moving them into a class?

src/utils/changes.ts
```diff
@@ -175,3 +181,152 @@ export function prependChangeset(
...
+        number
+        milestone {
+          number
+        }
+      }
+    }
+  }`);
+
+  const milestones: {
+    [number: string]: Milestone;
```
iker-barriocanal, line 243: Personal preference to use [`Record`](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeystype) in these cases.

src/utils/changes.ts
```diff
@@ -175,3 +181,152 @@ export function prependChangeset(
...
+      hasPRinTitle: Boolean(gitCommit.pr),
+      milestone: null,
+    });
+    const commitInfo = githubCommits[`C${hash}`];
+    if (commitInfo) {
+      const pr = commitInfo.associatedPullRequests.nodes[0];
+      if (pr) {
+        commit.pr = pr.number;
+        commit.milestone = pr.milestone?.number;
+      }
```
iker-barriocanal, line 266: IMO, moving this before `commit` is created makes it easier to read (get the data and create the object from it, instead of creating it with some data and assigning them afterward). Same with the milestone logic below.

src/utils/changes.ts
```diff
@@ -175,3 +181,152 @@ export function prependChangeset(
...
+    } else {
+      missing.push(commit);
+    }
+
+    const milestoneNum = commit.milestone;
+    if (milestoneNum != null) {
+      const milestone = milestones[milestoneNum] || { prs: [] };
+      // We _know_ the PR exists as milestones are attached to PRs
```
iker-barriocanal, line 274: Q: is it possible to get a PR from GitHub that doesn't exist?

src/utils/changes.ts
```diff
@@ -175,3 +181,152 @@ export function prependChangeset(
...
+      ${milestoneQuery}
+    }
+  }
+
+    fragment MilestoneFragment on Milestone {
+      title
+      description
+      state
+    }
+  `);
```
iker-barriocanal, line 307: There are some GraphQL queries. What do you think about encapsulating them in methods and calling these methods, instead of doing them in here? I suggest placing them in `src/utils/githubApi`.

iker-barriocanal (commented):
I suggest renaming `changes.ts` to something closer to `gitChanges`. Without context, `changes` it's harder to understand what is referring to.

src/utils/changes.ts
BYK: It says hasPRin**Title** not just "hasPR" 😁 Maybe I can add a comment here:

Typically, GitHub automatically adds the PR number at the end of commit title in the format ` (#<PR_NO>)`. This allows us to infer the PR number from the commit log only, staying offline. However, it is possible to remove this line when merging PRs. Even more, it is very easy to just use the commit SHA to get from a commit to a pr to a milestone using GraphQL. This only matters when formatting the commit (or potentially if a commit is associated with multiple PRs but that's irrelevant for us for now) as we want to know whether we should append the PR number provided by GitHub at the end of the commit title or not.

Makes sense?

src/utils/changes.ts
BYK: I'll add typing. They can only be `OPEN` or `CLOSED`.

src/utils/changes.ts
BYK: Not sure if that's actually possible but I'll try.

src/utils/changes.ts
BYK: PR is a concept that GitHub invented, so no. 😁 Even more, the case here is we have a milestone already which can only be associated with a PR or an issue, not a commit directly.

Ideally, we'd have a conditional type here where one says `pr: string | null; milestone: null` and the other part says `pr: string; milestone: string | null` but I got lazy, and also thought this was harder to reason about.

Thoughts?

src/utils/changes.ts
BYK: These are very specific queries so I wouldn't put them into the very generic `src/utils/githubApi` but I thought about moving them into separate functions. Just held that off until I got some feedback from you folks to support that idea. Will do 😄

BYK:
>I suggest renaming changes.ts to something closer to gitChanges. Without context, changes it's harder to understand what is referring to.

I'll rename it to `changelog.ts` then.

src/utils/changes.ts
BYK: Maybe later, not a fan of classes 😄

src/utils/githubApi.ts
```diff
@@ -87,6 +87,8 @@ export function getGithubApiToken(): string {
   return githubApiToken;
 }
 
+const _GitHubClientCache: Record<string, Octokit> = {};
```
kamilogorek, line 90: Maybe use `Map` instead? It has a nicer API to work with caches (`has` vs. coercing with `!`)

src/utils/changelog.ts
```diff
@@ -175,3 +181,193 @@ export function prependChangeset(
...
+  title: string;
+  hasPRinTitle: boolean;
+  pr: string | null;
+  milestone: string | null;
+}
+
+interface Milestone {
+  title?: string;
+  description?: string;
+  state?: 'OPEN' | 'CLOSED';
```
kamilogorek, line 196: How about doing `enum State` instead? We use it in few places so would be easier to refactor in the future if necessary.

src/utils/githubApi.ts
BYK: Since we are using strings as keys, using a good-old-object is the better way forward. I agree with `Map` API but typing and perf around it is still an issue to this day :(

src/utils/changelog.ts
BYK: Since this code is fairly isolated here, I'd prefer deferring the enum to a later stage. If you feel strongly, I can do it of course. Just don't see the immediate or even mid-term benefits.

src/utils/githubApi.ts
```diff
@@ -95,25 +97,27 @@ export function getGithubApiToken(): string {
  * @param token Github authentication token
  * @returns Github client
  */
-export function getGithubClient(token = ''): Octokit {
+export function getGitHubClient(token = ''): Octokit {
```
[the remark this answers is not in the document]
BYK: I think we do have protection for that on the line right below. Do you have some other concerns?

kamilogorek (approved):
All the replies make sense, looks good to me

src/utils/changes.ts
iker-barriocanal: Ooops 😅

src/utils/changes.ts
iker-barriocanal: Sounds good! My next question is whether we want to have PRs not attached to any milestone and commits in the same section, but that's more a behavior question and definitely something we can think about later. Just leaving my thoughts.

iker-barriocanal (approved):
I'd move the changes related to node versions and `tsconfig` to different PRs, but not blocking.

BYK:
>I'd move the changes related to node versions and tsconfig to different PRs, but not blocking.

This was also on my mind so I'll get that done first before landing.

Thanks a lot @kamilogorek and @iker-barriocanal for the great reviews. Btw. I'll also add some tests before landing as nothing hits these lines of codes during tests and they were buggy during my refactors.

src/utils/changes.ts
BYK: >My next question is whether we want to have PRs not attached to any milestone and commits in the same section, but that's more a behavior question and definitely something we can think about later.

Yup, this was the initial design we agreed on. Change/PR does not matter if they are not tied to a larger milestone for now.

src/utils/__tests__/changelog.test.ts
```diff
@@ -0,0 +1,504 @@
...
+  const sampleChangeset = {
+    body: '- this is a test',
+    name: 'Version 1.0.0',
+  };
+
+  test.each([
+    [
+      'regular',
+      `# Changelog\n## ${sampleChangeset.name}\n${sampleChangeset.body}\n`,
+    ],
```
iker-barriocanal, line 27: What about prereleases? Can we add tests like this one to test them?

src/utils/__tests__/changelog.test.ts
```diff
@@ -0,0 +1,504 @@
...
+          message: 'Upgraded the kernel',
+          pr: null,
+        },
+      ]);
+      makeCommitResponse([]);
+      const changes = await generateChangesetFromGit(dummyGit, '1.0.0');
+      expect(changes).toMatchInlineSnapshot(`
+        "## Various fixes & improvements
+
+        - Upgraded the kernel (abcdef12)"
```
iker-barriocanal, line 323: Is it worth it to test the hash length (to be more than 7)?

src/utils/__tests__/changelog.test.ts
```diff
@@ -0,0 +1,504 @@
...
+      expect(changes).toMatchInlineSnapshot(`
+        "## Various fixes & improvements
+
+        - Upgraded the kernel (abcdef12)
+        - Upgraded the manifold (#123)
+        - Refactored the crankshaft (#456)"
+      `);
+    });
+  });
+  it('should handle group prs under milestones', async () => {
```
iker-barriocanal, line 419: This test isn't only testing group prs, but also group prs + commits without prs. This should be reflected in the naming.

```suggestion
  it('should handle group prs under milestones and individual commits', async () => {
```

We may also want to include prs that don't belong to any milestone (in a test where some prs belong to milestones).

src/utils/__tests__/changelog.test.ts
BYK: Not sure I understand the case you want to cover with a prerelease version. The title search here has nothing specific towards versions.

Btw these are old tests that already existed so not touching them in this PR. GitHub shows them as brand new as I renamed the test file to align with the main module (changes -> changelog)

src/utils/__tests__/changelog.test.ts
```diff
@@ -0,0 +1,504 @@
...
+    ## 1.0.1
+    newer
+
+    ## ${sampleChangeset.name} (2019-02-02)
+    ${sampleChangeset.body}
+
+    ## 0.9.0
+    older
+    `,
+    ],
```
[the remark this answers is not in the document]
BYK: We can have both. Again tho, these are already existing tests so not touching in this PR 🙂

src/utils/__tests__/changelog.test.ts
```diff
@@ -0,0 +1,504 @@
...
+
+    ${sampleChangeset.name}
+    -------
+    ${sampleChangeset.body}
+
+    ## 0.9.0
+    older
+    `,
+    ],
+  ])('should extract %s', (_testName, markdown) => {
```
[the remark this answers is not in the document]
BYK: Will look into this in a future PR: existing tests.

src/utils/__tests__/changelog.test.ts
BYK: We already do that on line 313, right? That's where this value comes from.

src/utils/__tests__/changelog.test.ts
BYK: >We may also want to include prs that don't belong to any milestone (in a test where some prs belong to milestones).

We have that case above on lines 414 and 415? We can add a similar PR here but I don't see the added benefit/coverage.

src/utils/__tests__/changelog.test.ts
iker-barriocanal: My comment was more about adding another check to prevent a regression be in the snapshots, but that's another thing 🙂.

iker-barriocanal (approved):

src/commands/prepare.ts
```diff
@@ -393,14 +389,7 @@ async function prepareChangelog(
...
         replaceSection = changeset.name;
-        const changes = await getChangesSince(git, oldVersion);
-        changeset.body = changes
-          .map(change =>
-            change.pr
-              ? `- ${change.message}`
-              : `- ${change.message} (${change.hash.slice(0, 8)})`
-          )
-          .join('\n');
+        changeset.body = await generateChangesetFromGit(git, oldVersion);
```
chadwhitacre, line 392: Nice to see this encapsulation. 👍

src/utils/githubApi.ts
```diff
@@ -95,25 +97,27 @@ export function getGithubApiToken(): string {
...
+    const { retry } = require('@octokit/plugin-retry');
+    const octokitWithRetries = Octokit.plugin(retry);
+    _GitHubClientCache[githubApiToken] = new octokitWithRetries(attrs);
   }
 
-  // eslint-disable-next-line @typescript-eslint/no-var-requires
-  const { retry } = require('@octokit/plugin-retry');
-  const octokitWithRetries = Octokit.plugin(retry);
-  return new octokitWithRetries(attrs);
+  return _GitHubClientCache[githubApiToken];
```
chadwhitacre, line 120: Any reason to introduce this cache on this particular PR, or is this more of an opportunistic optimization?

BYK: During my refactoring, I moved the calls to `getGitHubClient` function into the two helper functions because it felt stupid to call it from outside and keep passing it into the helpers. Turns out we never need a new client throughout the application unless we provide a new token so I made the optimization.

So it is both I guess 🙂

src/utils/__tests__/changelog.test.ts
```diff
@@ -0,0 +1,504 @@
...
+    });
+    it('should use short commit SHA for local commits w/o pull requests', async () => {
+      mockGetChangesSince.mockResolvedValueOnce([
+        {
+          hash: 'abcdef1234567890',
+          message: 'Upgraded the kernel',
+          pr: null,
+        },
+      ]);
+      makeCommitResponse([{ hash: 'abcdef1234567890' }]);
```
chadwhitacre, line 334: This test is confusing. It's _almost_ the same as the one just before except for this line, but it's not obvious how this line means we are testing "w/o pull requests," because in both this and the previous test we have `pr: null` in the call to `mockResolvedValueOnce`.

chadwhitacre: I infer from later tests that `mockResolvedValueOnce` simulates what's available locally while `makeCommitResponse` simulates what is available remotely. Not a big deal but maybe a comment or something could help to explain how the test suite is functioning.

chadwhitacre: I'll also say that I have a personal preference for building test abstractions that highlight the variations in what is being tested. There's a lot of boilerplate repeated from test to test here that masks the meaning of each.

chadwhitacre: But that would mean a solid refactor and I understand if you don't want to take that on. :)

src/utils/__tests__/changelog.test.ts
```diff
@@ -0,0 +1,504 @@
...
+        pr: '900',
+        milestone: '5',
+      },
+    ]);
+    mockClient.mockResolvedValueOnce({
+      repository: {
+        M1: {
+          title: 'Better drivetrain',
+          description:
+            'We have upgraded the drivetrain for a smoother and more performant driving experience. Enjoy!',
```
chadwhitacre, line 474: 😁

src/utils/__tests__/changelog.test.ts
chadwhitacre: Brainstorm would be a `changesWith` wrapper that takes the variable elements and does the mock dancing, so that tests look like this:

```javascript
    it('should use short commit SHA for local commits', async () => {                                          
      var localCommits = [
        { 
          hash: 'abcdef1234567890',
          message: 'Upgraded the kernel',
          pr: null,
        },
      ];

      // remoteCommits and milestones would have defaults in changesWith but are here for illustration ...
      var remoteCommits = [];
      var milestones = [];

      expect(changesWith(localCommits, remoteCommits, milestones).toMatchInlineSnapshot(`
        "## Various fixes & improvements

        - Upgraded the kernel (abcdef12)"
      `);
    });
```

Then from here I guess we would get to parameterized testing. 😁

BYK: Okay added a comment to alleviate the immediate confusion.

I agree with your suggestions but I'd defer that to a follow up PR as it indeed is a large refactor and I'd love to get this out first and better the tests later (as long as they cover the critical paths ofc).

src/utils/changelog.ts
```diff
@@ -0,0 +1,396 @@
...
+    logger.warn(
+      `The following commits were not found on GitHub:`,
+      missing.map(commit => `${commit.hash.slice(0, 8)} ${commit.title}`)
+    );
+  }
+
+  const milestonesInfo = await getMilestonesDetails(Object.keys(milestones));
+
+  const changelogSections = [];
+  for (const milestoneNum of Object.keys(milestones)) {
```
chadwhitacre, line 264: What does the sort end up being on this? Comment? Test?

src/utils/changelog.ts
```diff
@@ -0,0 +1,396 @@
...
+  }
+
+  if (missing.length > 0) {
+    logger.warn(
+      `The following commits were not found on GitHub:`,
+      missing.map(commit => `${commit.hash.slice(0, 8)} ${commit.title}`)
+    );
+  }
+
+  const milestonesInfo = await getMilestonesDetails(Object.keys(milestones));
```
chadwhitacre, line 261: nit: `milestonesInfo` since that's the variable name we assign to

src/utils/changelog.ts
```diff
@@ -0,0 +1,396 @@
...
+  return text;
+}
+
+export async function generateChangesetFromGit(
+  git: SimpleGit,
+  rev: string
+): Promise<string> {
+  const gitCommits = await getChangesSince(git, rev);
+
+  const githubCommits = await getMilestoneAndPRFromCommits(
```
chadwhitacre, line 222: nit: `getMilestonesAndPRsFromCommits` since we get plural milestones and PRs (right?) and we have plural in `getMilestonesDetails`

src/utils/changelog.ts
chadwhitacre: Sorry ... meant `getMilestonesInfo`. Either that or `milestonesDetails` for the variable, point is make them match

src/utils/changelog.ts
```diff
@@ -0,0 +1,396 @@
...
+      ${milestoneQuery}
+    }
+  }
+
+  fragment MilestoneFragment on Milestone {
+    title
+    description
+    state
+  }
+`)) as MilestonesDetailsResult).repository;
```
chadwhitacre, line 387: Maybe worth exploring [a query builder](https://www.npmjs.com/package/gql-query-builder) in the future if this becomes something we do more of?

src/utils/changelog.ts
chadwhitacre: Or since milestones are nested under PRs: `getPRsWithMilestonesFromCommits`?

chadwhitacre (changes requested):
Some remaining questions/nits/naming suggestions.

FTR was slightly harder to review because of the `changes.ts` -> `changelog.ts` rename. It's a good rename but doing the filename change on a separate PR would've saved me some manual diffing locally to understand what had changed, or perhaps a note calling out the additions.

src/utils/changelog.ts
BYK: Yup, totally agree with that. And a validator for tests. Overkill for now though.

src/utils/changelog.ts
BYK: The naming is due to the `commit -> PR -> milestone` mapping. 1 commit to 1 PR to 1 Milestone. I guess making the other one singular is the right way to go here.

src/utils/changelog.ts
BYK: No sort at the moment, don't see a strong need yet. JS dictionaries now implicit sorting based on the insertion order so this should always produce consistent results.

BYK:
>FTR was slightly harder to review because of the changes.ts -> changelog.ts rename.

Sorry about this, realized it too late that GitHub did not recognize the rename properly 🙁

src/utils/changelog.ts
chadwhitacre: Sgtm. Also, excellent footnote game. 😁

chadwhitacre (approved):
Seems we're talking past each other on function names. Happy to unblock this and revisit later if necessary.

src/utils/changelog.ts
iker-barriocanal: I rather `milestonesDetails`, `getMilestonesInfo` sounds more like a function.

chadwhitacre: lolsob so much confusion ... 

> Sorry ... meant `getMilestonesInfo`. Either that or `milestonesDetails` for the variable, point is make them match

"Either that" meaning either change `Details` to `Info` in the function name `getMilestones_________` or change `Info` to `Details` in the variable name `milestones__________`. Use the same noun in the variable and function names but yes `getFooBar` sounds like a function. 😜

PR #296, opened 2021-09-24
fix(changelog): Escape # in milestones

BYK:
We need to escape the `#` sign in milestone titles and bodies as it may interfere or get hidden with the markdown formatting (it uses `#` to identify headers). This patch fixes the issue and it also fixes another issue when the milestone description was empty.

### DIFF
--- a/src/utils/__tests__/changelog.test.ts
+++ b/src/utils/__tests__/changelog.test.ts
@@ -487,8 +487,7 @@ describe('generateChangesetFromGit', () => {
         },
         M5: {
           title: 'Better driver experience',
-          description:
-            'We are working on making your driving experience more pleasant and safer.',
+          description: '',
           state: 'OPEN',
         },
       },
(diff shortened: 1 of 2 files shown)


src/utils/__tests__/changelog.test.ts
```diff
@@ -487,8 +487,7 @@ describe('generateChangesetFromGit', () => {
         },
         M5: {
           title: 'Better driver experience',
-          description:
-            'We are working on making your driving experience more pleasant and safer.',
+          description: '',
```
chadwhitacre, line 490: Even more of an edge case (and not necessary for this PR) ... what if the description is `'  '` ... i.e., non-empty whitespace?

src/utils/__tests__/changelog.test.ts
```diff
@@ -513,4 +510,69 @@ describe('generateChangesetFromGit', () => {
...
+      {
+        hash: 'abcdef1234567890',
+        pr: '123',
+        milestone: '1',
+      },
+    ]);
+    mockClient.mockResolvedValueOnce({
+      repository: {
+        M1: {
+          title: 'Drivetrain #1 in town',
```
chadwhitacre, line 532: 😂

src/utils/changelog.ts
```diff
@@ -39,9 +39,13 @@ export interface ChangesetLoc {
...
 }
 
+function escapeMarkdownPound(text: string): string {
+  return text.replace(/#/g, '&#35;');
+}
+
 function markdownHeader(level: number, text: string): string {
   const prefix = new Array(level + 1).join('#');
-  return `${prefix} ${text}`;
+  return `${prefix} ${escapeMarkdownPound(text)}`;
```
chadwhitacre, line 48: YOU MEAN OCTOTHORPE :trollface:

chadwhitacre (approved):

src/utils/__tests__/changelog.test.ts
BYK: We can trim it but I'd assume GitHub would do that already.

src/utils/changelog.ts
BYK: LOL WUT?

chadwhitacre: https://en.wikipedia.org/wiki/Number_sign 😁

Screen Shot 2021-09-24 at 10 07 31 AM

PR #297, opened 2021-09-24
feat(changelog): Skip items with `#skip-changelog` in them

BYK:
We sometimes want to omit certain changes (such as internal changes in Relay and this patch brings the option to skip them from the changelog altogether.

### DIFF
--- a/src/utils/__tests__/changelog.test.ts
+++ b/src/utils/__tests__/changelog.test.ts
@@ -12,6 +12,7 @@ import {
   removeChangeset,
   prependChangeset,
   generateChangesetFromGit,
+  SKIP_CHANGELOG_MAGIC_WORD,
 } from '../changelog';
 
 describe('findChangeset', () => {
@@ -268,21 +269,24 @@ describe('generateChangesetFromGit', () => {
   const makeCommitResponse = (commits: Record<string, string>[]) =>
     mockClient.mockResolvedValueOnce({
       repository: Object.fromEntries(
-        commits.map(({ hash, pr, milestone }: Record<string, string>) => [
-          `C${hash}`,
-          {
-            associatedPullRequests: {
-              nodes: pr
-                ? [
-                    {
-                      number: pr,
-                      milestone: milestone ? { number: milestone } : null,
-                    },
-                  ]
-                : [],
+        commits.map(
+          ({ hash, pr, prBody, milestone }: Record<string, string>) => [
+            `C${hash}`,
+            {
+              associatedPullRequests: {
+                nodes: pr
+                  ? [
+                      {
+                        number: pr,
+                        body: prBody || '',
+                        milestone: milestone ? { number: milestone } : null,
+                      },
+                    ]
+                  : [],
+              },
             },
-          },
-        ])
+          ]
+        )
       ),
     });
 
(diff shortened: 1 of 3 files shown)


src/utils/__tests__/changelog.test.ts
```diff
@@ -418,36 +429,42 @@ describe('generateChangesetFromGit', () => {
       `);
     });
   });
-  it('should handle group prs under milestones', async () => {
+  it('should group prs under milestones', async () => {
```
chadwhitacre, line 432: 👀

src/utils/changelog.ts
```diff
@@ -11,6 +11,7 @@ import { getVersion } from './version';
  */
 export const DEFAULT_CHANGELOG_PATH = 'CHANGELOG.md';
 export const DEFAULT_UNRELEASED_TITLE = 'Unreleased';
+export const SKIP_CHANGELOG_MAGIC_WORD = '#skip-changelog';
```
chadwhitacre, line 14: Any considerations around escaping? In #296 we added escaping of #, I don't _think_ this should conflict but we are using the same OCTOTHORPE character here so something to think about.

src/utils/changelog.ts
```diff
@@ -250,9 +253,16 @@ export async function generateChangesetFromGit(
     const hash = gitCommit.hash;
 
     const githubCommit = githubCommits[hash];
+    if (
+      githubCommit?.prBody &&
+      githubCommit.prBody.includes(SKIP_CHANGELOG_MAGIC_WORD)
```
chadwhitacre, line 258: Why would this not work here?

```suggestion
      githubCommit?.prBody?.includes(SKIP_CHANGELOG_MAGIC_WORD)
```

src/utils/changelog.ts
```diff
@@ -365,6 +380,7 @@ async function getPRAndMilestoneFromCommit(
       hash.slice(1),
       {
         pr: commit?.associatedPullRequests.nodes[0]?.number ?? null,
+        prBody: commit?.associatedPullRequests.nodes[0]?.body ?? null,
```
chadwhitacre, line 383: Kinda WET but meh.

chadwhitacre (changes requested):
Few small considerations, overall lgtm.

src/utils/changelog.ts
BYK: No need to escape as this does not end up in markdown.

src/utils/changelog.ts
BYK: It does, it is just needed as we already establish the fact that both `githubCommit` and `githubCommit.prBody` exists. See [optional chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) if you are curious 🙂

src/utils/changelog.ts
BYK: Yeah, I'll see if I can tidy up this and the tests in a future PR. Tests also annoy me with the repetition.

src/utils/changelog.ts
chadwhitacre: Sorry ... maybe wasn't clear, why have the initial `githubCommit?.prBody &&` if we can just use optional chaining?

chadwhitacre (approved):

src/utils/changelog.ts
BYK: Oh, I guess I can do that 😅

BYK: #298

src/utils/changelog.ts
BYK: #298

PR #301, opened 2021-09-29
fix(changelog): Accomodate `null` milestone descriptions

BYK:
While testing this on Sentry, we came by some milestones where their `description` field was `null`. This does not happen with new milestones but we still need to cover for that case to avoid crashes.

### DIFF
--- a/src/utils/__tests__/changelog.test.ts
+++ b/src/utils/__tests__/changelog.test.ts
@@ -297,7 +297,7 @@ describe('generateChangesetFromGit', () => {
 
   interface TestMilestone {
     title: string;
-    description: string;
+    description: string | null;
   }
 
   function setup(
@@ -516,13 +516,20 @@ describe('generateChangesetFromGit', () => {
       ].join('\n'),
     ],
     [
-      'should omit milestone body if it is empty',
+      'should omit milestone body if it is empty or null',
       [
         {
           hash: 'abcdef1234567890',
           title: 'Upgraded the kernel',
           body: '',
           pr: { local: '123', remote: { number: '123', milestone: '1' } },
+        },
+
+        {
+          hash: 'bcdef123456789a',
+          title: 'Upgraded the manifold (#456)',
+          body: '',
+          pr: { local: '456', remote: { number: '456', milestone: '2' } },
         },
       ],
       {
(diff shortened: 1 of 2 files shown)


src/utils/changelog.ts
```diff
@@ -300,7 +300,7 @@ export async function generateChangesetFromGit(
         `${milestone.title}${milestone.state === 'OPEN' ? ' (ongoing)' : ''}`
       )
     );
-    if (milestone.description !== '') {
+    if (milestone.description && milestone.description !== '') {
```
chadwhitacre, line 303: Both null and empty string are false-y, so I think this is enough:

```suggestion
    if (milestone.description) {
```

src/utils/changelog.ts
```diff
@@ -416,18 +421,23 @@ async function getMilestoneInfo(
...
+      title
+      description
+      state
+    }
+  `;
+  logger.trace('Running graphql query:', graphqlQuery);
+  const milestoneInfo = ((await getGitHubClient().graphql(
+    graphqlQuery
+  )) as MilestonesDetailsResult).repository;
+  logger.trace('Query result:', milestoneInfo);
```
chadwhitacre, line 440: The rest of the changes in this file are just adding `logger.trace` calls for GraphQL queries, correct? Anything substantive beyond that?

src/utils/__tests__/changelog.test.ts
```diff
@@ -516,23 +516,43 @@ describe('generateChangesetFromGit', () => {
       ].join('\n'),
     ],
     [
-      'should omit milestone body if it is empty',
+      'should omit milestone body if it is empty or null',
```
chadwhitacre, line 519: Suggest: "if it is falsey" and then throw in an undefined case as well, for good measure.

chadwhitacre (changes requested):
Consider generalizing slightly more to encompass undefined as well.

Also a question about scope of changes, make sure I'm not missing something. :)

src/utils/changelog.ts
BYK: I wanted it to be explicit but you are right.

src/utils/__tests__/changelog.test.ts
BYK: `undefined` can never happen tho, at least given our types and how the API behaves. But I guess it makes sense. It just requires me to update the types again which I don't want to.

src/utils/changelog.ts
BYK: Nothing. These were added to make debugging the issue easier and I then realized these should exist anyway and left them there.

chadwhitacre: Sounds good, just wanted to make sure I wasn't missing something. :)

src/utils/__tests__/changelog.test.ts
chadwhitacre: If you saaaaaaaayyyy sooooooooooo ......... ;)

chadwhitacre (approved):

PR #307, opened 2021-10-05
fix: UPM no longer expects exactly 1 artifact

bitsandfoxes:
The requirements for releasing the Unity SDK changed and with it the artifacts we upload

- Previously:
   Exactly one artifact `package-release.zip` that contains the UPM package.
- Now (with native support):
   Also contains symbols for the Symbols Collector

So we just need to pick that one `package-release.zip` from all the artifacts and we're good to go!


HazAT (approved):

src/targets/upm.ts
```diff
@@ -30,6 +30,8 @@ export class UpmTarget extends BaseTarget {
   public readonly github: Octokit;
   /** Internal GitHub Target */
   private readonly githubTarget: GithubTarget;
+  /**  */
+  public readonly packageName: string = 'package-release.zip';
```
BYK, line 34: Shouldn't this be a constant?

BYK:
GitHub actions are broken so waiting till the builds pass.

BYK (approved):

src/targets/upm.ts
bitsandfoxes: It should! Changed it.

src/targets/upm.ts
```diff
@@ -72,16 +75,17 @@ export class UpmTarget extends BaseTarget {
       reportError('Cannot publish UPM: No release artifact found.');
       return;
     }
-    if (packageFiles.length > 1) {
+
+    const packageFile = packageFiles.find(
+      packageFile => packageFile.filename === ARTIFACT_NAME
```
BYK, line 80:
```suggestion
      ({ filename }) => filename === ARTIFACT_NAME
```

BYK (approved):

Issue #24, opened 2019-06-21
add repository parameters to pypi


dhuckins:
would it be possible to add additional parameters to the pypi upload?
specifically `TWINE_REPOSITORY_URL` for uploading to private/internal repositories

can submit PR

jan-auer:
Definitely, PR appreciated. You can add it in a similar fashion as the credentials are set, or also add it to the target config.

https://github.com/getsentry/craft/blob/f06a0157989134eecf3127272f1e9fd2688b589e/src/targets/pypi.ts#L60-L63

PR #310, opened 2021-10-06
Adding Twine Repository as config

joshfinnie:
This PR handles #24 which allows users to add a private/different PyPi repository url by adding a `TWINE_REPOSITORY_URL` as an environment variable.

Hoping this is all that is needed!


joshfinnie:
@BYK ah, that makes sense... should have tested that before ;-) I'll confirm and close if so.

PR #308, opened 2021-10-05
feat(github): Verify release uploads for GitHub

BYK:
We had an incident where we sent UTF-encoded string
when uploading GitHub release artifacts after an
Octokit upgrade due to misleading types. We haven't
noticed the issue until it was too late as we lack
basic checks such as file size and checksums. This
patch introduces both checks to the GitHub target.

Closes #303.

PR #328, opened 2021-11-16
feat(target-gh): Check asset MD5 hash by downloading the asset

iker-barriocanal:
The GitHub API doesn't provide any endpoints to get the hash of an asset.
GitHub's response contains an ETag header with the MD5 hash of the asset.  In
certain cases, the ETag header is not present. This means we don't have the
hash of the asset, and thus, we cannot verify the asset correctness. To verify
it, we download the file and calculate the hash locally.

Files are downloaded into memory, and their size is not checked. This has a
risk of downloading too big files, which is something we accept at the moment.

The GitHub SDK throws exceptions when server responses are HTTP errors. These
interactions are wrapped and more detailed errors are thrown.

Originally, the hash check was introduced in https://github.com/getsentry/craft/pull/308.
This is a follow-up to the initial fix: https://github.com/getsentry/craft/pull/323.
Releases were failing before this fix, see https://github.com/getsentry/publish/issues/635.

Closes https://github.com/getsentry/craft/issues/322.

### DIFF
--- a/src/targets/github.ts
+++ b/src/targets/github.ts
@@ -378,11 +378,11 @@ export class GithubTarget extends BaseTarget {
         );
       }
 
-      const remoteChecksum = await this.getRemoteChecksum(url);
-      const localChecksum = this.md5FromData(file);
+      const remoteChecksum = await this.checksumFromUrl(url);
+      const localChecksum = this.checksumFromData(file);
       if (localChecksum !== remoteChecksum) {
         throw new Error(
-          `Uploaded asset MD5 checksum does not match local asset checksum for "${name} (${localChecksum} != ${remoteChecksum})`
+          `Uploaded asset checksum does not match local asset checksum for "${name} (${localChecksum} != ${remoteChecksum})`
         );
       }
       uploadSpinner.succeed(`Uploaded asset "${name}".`);
@@ -394,7 +394,7 @@ export class GithubTarget extends BaseTarget {
     }
   }
 
-  private async getRemoteChecksum(url: string): Promise<string> {
+  private async checksumFromUrl(url: string): Promise<string> {
     // XXX: This is a bit hacky as we rely on two things:
     // 1. GitHub issuing a redirect to S3, where they store the artifacts,
     //    or at least pass those request headers unmodified to us
@@ -443,10 +443,10 @@ export class GithubTarget extends BaseTarget {
         `Cannot download asset from GitHub. Status: ${(e as any).status}\n` + e
       );
     }
-    return this.md5FromData(Buffer.from(response.data));
+    return this.checksumFromData(Buffer.from(response.data));
   }
 
-  private md5FromData(data: BinaryLike): string {
+  private checksumFromData(data: BinaryLike): string {
     return createHash('md5').update(data).digest('hex');
   }
 
--- a/src/targets/github.ts
+++ b/src/targets/github.ts
@@ -395,11 +395,13 @@ export class GithubTarget extends BaseTarget {
   }
 
   private async checksumFromUrl(url: string): Promise<string> {
-    // XXX: This is a bit hacky as we rely on two things:
-    // 1. GitHub issuing a redirect to S3, where they store the artifacts,
-    //    or at least pass those request headers unmodified to us
-    // 2. AWS S3 using the MD5 hash of the file for its ETag cache header
-    //    when we issue a HEAD request.
+    // XXX: This is a bit hacky as we rely on various things:
+    // 1. GitHub issuing a redirect to AWS S3.
+    // 2. S3 using the MD5 hash of the file for its ETag cache header.
+    // 3. The file being small enough to fit in memory.
+    //
+    // Note that if assets are large (5GB) assumption 2 is not correct. See
+    // https://github.com/getsentry/craft/issues/322#issuecomment-964303174
     let response;
     try {
       response = await this.github.request(`HEAD ${url}`, {
--- a/src/targets/github.ts
+++ b/src/targets/github.ts
@@ -2,7 +2,7 @@ import { Octokit, RestEndpointMethodTypes } from '@octokit/rest';
 import { RequestError } from '@octokit/request-error';
 import { readFileSync, promises, statSync } from 'fs';
 import { basename } from 'path';
-import { createHash } from 'crypto';
+import { BinaryLike, createHash } from 'crypto';
 
 import { getConfiguration } from '../config';
 import {
@@ -379,7 +379,7 @@ export class GithubTarget extends BaseTarget {
       }
 
       const remoteChecksum = await this.getRemoteChecksum(url);
-      const localChecksum = createHash('md5').update(file).digest('hex');
+      const localChecksum = this.md5FromData(file);
       if (localChecksum !== remoteChecksum) {
         throw new Error(
           `Uploaded asset MD5 checksum does not match local asset checksum for "${name} (${localChecksum} != ${remoteChecksum})`
@@ -443,8 +443,11 @@ export class GithubTarget extends BaseTarget {
         `Cannot download asset from GitHub. Status: ${(e as any).status}\n` + e
       );
     }
+    return this.md5FromData(Buffer.from(response.data));
+  }
 
-    return createHash('md5').update(Buffer.from(response.data)).digest('hex');
+  private md5FromData(data: BinaryLike): string {
+    return createHash('md5').update(data).digest('hex');
   }
 
   private async handleGitHubUpload(
--- a/src/targets/github.ts
+++ b/src/targets/github.ts
@@ -385,6 +385,7 @@ export class GithubTarget extends BaseTarget {
           `Uploaded asset MD5 checksum does not match local asset checksum for "${name} (${localChecksum} != ${remoteChecksum})`
         );
       }
+      uploadSpinner.succeed(`Uploaded asset "${name}".`);
       return url;
     } catch (e) {
       uploadSpinner.fail(`Cannot upload asset "${name}".`);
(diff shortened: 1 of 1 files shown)


src/targets/github.ts
```diff
@@ -378,12 +378,30 @@ export class GithubTarget extends BaseTarget {
...
       }
 
-      // XXX: This is a bit hacky as we rely on two things:
-      // 1. GitHub issuing a redirect to S3, where they store the artifacts,
-      //    or at least pass those request headers unmodified to us
-      // 2. AWS S3 using the MD5 hash of the file for its ETag cache header
-      //    when we issue a HEAD request.
-      const response = await this.github.request(`HEAD ${url}`, {
+      const remoteChecksum = await this.getRemoteChecksum(url);
+      const localChecksum = createHash('md5').update(file).digest('hex');
```
chadwhitacre, line 382: Should we preserve the note added in #323 about this local computation only working for simple cases?

chadwhitacre: Nevermind. Now that we're computing the remote checksum locally (😜) we are guaranteed the same algorithm 👍

chadwhitacre: Though in that case should we factor out a function that we call here and also beneath `md5FromUrl` to reinforce/ensure the same algorithm?

src/targets/github.ts
```diff
@@ -395,32 +413,37 @@ export class GithubTarget extends BaseTarget {
...
+    let response;
+    try {
+      response = await this.github.request(`GET ${url}`, {
+        headers: {
+          Accept: DEFAULT_CONTENT_TYPE,
+        },
+      });
+    } catch (e) {
+      throw new Error(
+        `Cannot download asset from GitHub. Status: ${(e as any).status}\n` + e
```
chadwhitacre, line 442: How hard is it to put a retry around this? Intermittent network issues seem likely enough that I feel like we're asking for randomly failed releases if we aren't a little more graceful here.

chadwhitacre (changes requested):
Looks good overall, a couple questions/suggestions.

src/targets/github.ts
rhcarvalho: If ETag is computed different for large files, as https://github.com/getsentry/craft/issues/322#issuecomment-964303174 suggests, that would still be a problem -- not something we're fixing here.

Do we have any repo with a release asset >> 1 GB?

I'm okay adding one more caveat to the list of assumptions the code makes.

As for "reinforce/ensure the same algorithm" -- do you mean wrapping `createHash('md5').update(BYTES).digest('hex')` as a function, or something else?

chadwhitacre: > If ETag is computed different for large files

It is by S3, but we're not depending on S3's calculation anymore, we're downloading the entire asset and computing a hash ourselves.

> do you mean wrapping

Yeah that's all, a small helper function to guarantee that `remoteChecksum` and `localChecksum` remain identical. Wdyt?

src/targets/github.ts
rhcarvalho: @iker-barriocanal and I talked about the ideal scenario of being able to gather metrics around this and other code paths -- to focus effort on what needs improvement.

We've also talked about retries.

For the cases we've seen recently, retry on this level would not have helped. E.g. in https://github.com/getsentry/publish/runs/4222412947?check_suite_focus=true#step:8:427, retrying on 404 would just take longer to fail the overall run. Retrying by rerunning the target fixed the problem though, because it re-uploads assets.

So seems like a more holistic architectural decision around retries would be what we need, something we didn't intend to own right now.

@chadwhitacre || @iker-barriocanal do you have in mind cases where a retry in this level would help? If you do have concrete cases, then by all means we can add retries.

chadwhitacre: Nope. You've thought it through further than I have. Suggestion withdrawn. :)

src/targets/github.ts
rhcarvalho: > we're downloading the entire asset and computing a hash ourselves.

We only do that for the case where the `ETag` header is not present. If it is present and different (because it was computed in a different way or any other reason) then we'll fail the comparison and the GitHub target will fail (as it does today).

By the time we reach that size concern I wish GitHub will have a proper API to get a checksum so we don't need to upload/download every asset 😄 

BTW we've also noticed #308 added both an asset size check and a checksum check. Most likely, the size check was enough to detect the sort of problem #308 was dealing with. If we run into more problems with ETag/MD5 and if re-downloading assets causes collateral damage, I'd consider removing the checksum altogether.

rhcarvalho: > Yeah that's all, a small helper function to guarantee that `remoteChecksum` and `localChecksum` remain identical. Wdyt?

👍

chadwhitacre: > We only do that for the case where the ETag header is not present.

Ah, good point! In that case I'll bring back my original suggestion:

> Should we preserve the note added in #323 about this local computation only working for simple cases?

☺️

iker-barriocanal:
Let's summarize the conversation so far:

- [x] A small helper function to guarantee that `remoteChecksum` and `localChecksum` remain identical. Done in https://github.com/getsentry/craft/pull/328/commits/e016782914c53a8b0767d9b52812fc4bd3fa796c.
- [ ] Preserve the note about this local computation only working for simple cases.

We also identified a bug in https://github.com/getsentry/craft/pull/328/commits/469925dfa995312c440222ed309df90743301478, where the spinner wasn't being completed. This has been fixed in https://github.com/getsentry/craft/pull/328/commits/5083340e6da6a6209d9ba03e7865a300436f7190.

src/targets/github.ts
iker-barriocanal: This has been resolved, added https://github.com/getsentry/craft/pull/328#issuecomment-971638002 to summarize the state.

rhcarvalho (approved):
LGTM, but since I'm biased as @iker-barriocanal and I worked together on this, let's wait for @chadwhitacre to chime in.

chadwhitacre (approved):
Looks great! :shipit:

src/targets/github.ts
BYK: Late to the game but we already have a retry plugin on Octokit here which retries on 5xx errors automatically. Retrying on 404s doesn't make much sense.

src/targets/github.ts
```diff
@@ -395,32 +416,40 @@ export class GithubTarget extends BaseTarget {
...
+  private async md5FromUrl(url: string): Promise<string> {
+    this.logger.debug('Downloading asset from GitHub to check MD5 hash: ', url);
+    let response;
+    try {
+      response = await this.github.request(`GET ${url}`, {
+        headers: {
+          Accept: DEFAULT_CONTENT_TYPE,
+        },
+      });
+    } catch (e) {
```
BYK, line 443: I'm not sure if catching the original exception and throwing a new one is actually good here as you obscure the original issue. In debug mode, we already log the response status etc.

src/targets/github.ts
```diff
@@ -378,12 +378,33 @@ export class GithubTarget extends BaseTarget {
...
+
+      throw e;
+    }
+  }
+
+  private async checksumFromUrl(url: string): Promise<string> {
+    // XXX: This is a bit hacky as we rely on various things:
+    // 1. GitHub issuing a redirect to AWS S3.
+    // 2. S3 using the MD5 hash of the file for its ETag cache header.
+    // 3. The file being small enough to fit in memory.
```
BYK, line 401: Calculating MD5 hash from a streaming data is quite trivial in Node so I'd encourage you to revisit this approach in a follow up PR as it causes unnecessary memory usage. We may have rather large files.

src/targets/github.ts
rhcarvalho: Thanks @BYK !!

In this case here, we were seeing something like "Failed: Not Found" mixed up in the logs. So the `RequestError` (the `e`) from Octokit didn't provide much context, and not knowing clearly if the problem was in HEAD (the original ETag method) or GET (the new download method added in this PR) was not a great prospect.

The original error is still printed along with a custom message. Do you think we could do better?

src/targets/github.ts
rhcarvalho: Yes, @iker-barriocanal and I talked extensively about this.

Trade-offs:
- Octokit's `request` method seems to buffer the full response payload, so if we wanted to work with streams we'd need to depart from using `request`. I believe we're originally using Octokit in this code path to benefit from the different rate limits given to authenticated requests -- @BYK can you confirm that?
- We don't work with streams when doing the upload. We read (with `readSync`!) full asset files from disk into memory before uploading. We use the same bytes in memory to calculate the "local checksum". So the assumption that files are small enough was implicitly already there.
- If memory usage becomes a concern, then certainly we need to revisit this, but probably then not only the fallback download step.
- We've filed issues here on GitHub and internal tracking tickets for future improvements. Perhaps none specifically about streaming -- now done in https://github.com/getsentry/craft/issues/329

BYK: >Octokit's request method seems to buffer the full response payload, so if we wanted to work with streams we'd need to depart from using request. I believe we're originally using Octokit in this code path to benefit from the different rate limits given to authenticated requests 

Octokit takes care of a bunch of things from authentication to rate limiting to parallelization to automatic retries. So I'd say stick with it as long as it can. If solving this issue requires a departure from Octokit, delay that as long as possible and file an issue with them ahead of time so they might be able to bring this into the library itself.

>We don't work with streams when doing the upload. We read (with readSync!) full asset files from disk into memory before uploading. We use the same bytes in memory to calculate the "local checksum". So the assumption that files are small enough was implicitly already there.

Great catch. This is again due to Octokit limitations tho as it does not support streams or piping, only buffers.

rhcarvalho: References to our future selves (?):

- https://github.com/octokit/octokit.js/issues/916#issuecomment-397742916
- https://github.com/octokit/octokit.js/issues/887#issuecomment-391448184

2018 was a fine year, right? :)

PR #335, opened 2021-12-15
Limit the number of leftovers listed

chadwhitacre:
Nobody's going to read through a list of 1,000s of commits, may as well use git log at that point.

What's more, over-long changesets can lead to [GitHub API failures](https://github.com/getsentry/self-hosted/issues/1193#issuecomment-995011767).

### DIFF
--- a/src/utils/__tests__/changelog.test.ts
+++ b/src/utils/__tests__/changelog.test.ts
@@ -411,10 +411,18 @@ describe('generateChangesetFromGit', () => {
           body: '',
           pr: { remote: { number: '456', author: { login: 'bob' } } },
         },
+        {
+          hash: 'cdef1234567890ad',
+          title: 'Refactored the crankshaft again',
+          body: '',
+          pr: { remote: { number: '458', author: { login: 'bob' } } },
+        },
       ],
       {},
       [
         '### Various fixes & improvements',
+        '',
+        '(Only listing 3 out of 4.)',
         '',
         '- Upgraded the kernel (abcdef12)',
         '- Upgraded the manifold (#123) by @alice',
@@ -706,7 +714,7 @@ describe('generateChangesetFromGit', () => {
       output: string
     ) => {
       setup(commits, milestones);
-      const changes = await generateChangesetFromGit(dummyGit, '1.0.0');
+      const changes = await generateChangesetFromGit(dummyGit, '1.0.0', 3);
       expect(changes).toBe(output);
     }
   );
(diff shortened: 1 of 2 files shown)


src/utils/__tests__/changelog.test.ts
```diff
@@ -411,11 +411,19 @@ describe('generateChangesetFromGit', () => {
...
+          title: 'Refactored the crankshaft again',
+          body: '',
+          pr: { remote: { number: '458', author: { login: 'bob' } } },
+        },
       ],
       {},
       [
         '### Various fixes & improvements',
         '',
+        '(Only listing 3 out of 4.)',
```
billyvg, line 425: Maybe one of these alternatives with no parens and less words?

Listing 3 out of 4
_Listing 3 out of 4_
Listing **3** out of 4
Listing 3 of 4
_Showing 3 of 4_
Showing 3 of 4

src/utils/changelog.ts
```diff
@@ -246,7 +247,8 @@ function formatCommit(commit: Commit): string {
 
 export async function generateChangesetFromGit(
   git: SimpleGit,
-  rev: string
+  rev: string,
+  maxLeftovers: number = MAX_LEFTOVERS
```
billyvg, line 251: This is fine as-is, but I generally prefer having a single object parameter as it helps when reading the code at the callsite of the function since you'll be able to see the object's keys. (i.e. python's args vs kwargs).

billyvg (approved):

src/utils/__tests__/changelog.test.ts
chadwhitacre: Sure thing, done in b303490feab8a05f6af717329ae20f958c26c0e2. :)

BYK (approved):

BYK:
I don't think you are still immune to body overflow as we can go over the length limit via Milestones. Quite hard but possible, so just noting.

BYK:
Also sorry, didn't ever think we'd reach this kind of limit so quickly 😅

chadwhitacre:
Yes, it's possible with Milestones but less likely, GitHub is capping the POST body at 125,000 bytes.

And no need to apologize, easy enough to fix. Thanks for reviewing! :)

src/utils/__tests__/changelog.test.ts
chadwhitacre: Maybe 2ec7c16299b89e192aae735e61592edea254fe13 is even better, @billyvg?

jan-auer:
I'm afraid this didn't work, at least the last release still contained >2000 lines from the changelog. I'll check if we forgot to bump the version in the publish repository (i.e. what's "latest").

On another note, I'm not sure how useful it is to list 24 commits out of a list of 1500-2000. Might as well leave it out completely.

chadwhitacre:
> While publish uses the latest published craft, the release action uses the latest _released_ craft.

https://github.com/getsentry/publish/issues/666

PR #343, opened 2022-01-14
Publish to GitHub more atomically using draft repos

### LINKED ISSUE #336: Tag more carefully in GitHub target
Currently the GitHub target creates a tag fairly early in the process, which means that if something fails after it, the tag must be recreated. On https://github.com/getsentry/self-hosted/issues/1193 we observed a fairly drastic scenario where an overlarge changelog prevented creating a release object, and then a manual tag remove/add cascaded into a broken CI incident. Uploading artifacts to GitHub is another source of intermittent errors that can land us in a hanging tag state.

It seems from my reading of `createAnnotatedTag` that we need a tag kind of early in the process. Can we use a temporary tag and then set the final tag as the last step in the process?

What other ways could we solve this problem?

chadwhitacre:
Closes #336.

Tested with https://github.com/getsentry/testing-craft-chadwhitacre/blob/0.0.9/test.

E.g. https://github.com/getsentry/testing-craft-chadwhitacre/

------------

Screen Shot 2022-01-14 at 2 28 52 PM

------------

Screen Shot 2022-01-14 at 2 29 05 PM

### DIFF
--- a/src/targets/github.ts
+++ b/src/targets/github.ts
@@ -37,8 +37,6 @@ export interface GithubTargetConfig extends GithubGlobalConfig {
   tagPrefix: string;
   /** Mark release as pre-release, if the version looks like a non-public release */
   previewReleases: boolean;
-  /** Use annotated (not lightweight) tag */
-  annotatedTag: boolean;
 }
 
 /**
@@ -107,8 +105,6 @@ export class GithubTarget extends BaseTarget {
       owner,
       repo,
       changelog,
-      annotatedTag:
-        this.config.annotatedTag === undefined || !!this.config.annotatedTag,
       previewReleases:
         this.config.previewReleases === undefined ||
         !!this.config.previewReleases,
@@ -118,55 +114,7 @@ export class GithubTarget extends BaseTarget {
   }
 
   /**
-   * Creates an annotated tag for the given revision.
-   *
-   * Unlike a lightweight tag (basically just a pointer to a commit), to
-   * create an annotated tag we must create a tag object first, and then
-   * create a reference to it manually.
-   *
-   * @param version The version to release
-   * @param revision Git commit SHA to be published
-   * @param tag Tag to create
-   * @returns The newly created release
-   */
-  public async createAnnotatedTag(
-    version: string,
-    revision: string,
-    tag: string
-  ): Promise<void> {
-    this.logger.debug(`Creating a tag object: "${tag}"`);
-    const createTagParams = {
-      message: `Tag for release: ${version}`,
-      object: revision,
-      owner: this.githubConfig.owner,
-      repo: this.githubConfig.repo,
-      tag,
-      type: 'commit' as GithubCreateTagType,
-    };
-    const tagCreatedResponse = await this.github.git.createTag(createTagParams);
-
-    const ref = `refs/tags/${tag}`;
-    const refSha = tagCreatedResponse.data.sha;
-    this.logger.debug(`Creating a reference "${ref}" for object "${refSha}"`);
-    try {
-      await this.github.git.createRef({
-        owner: this.githubConfig.owner,
-        ref,
-        repo: this.githubConfig.repo,
-        sha: refSha,
-      });
-    } catch (e) {
-      if (e instanceof Error && e.message.match(/reference already exists/i)) {
-        this.logger.error(
-          `Reference "${ref}" already exists. Does tag "${tag}" already exist?`
-        );
-      }
-      throw e;
-    }
-  }
-
-  /**
-   * Gets an existing or creates a new release for the given version.
+   * Create a release draft for the given version.
    *
    * The release name and description body is brought in from `changes`
    * respective tag, if present. Otherwise, the release name defaults to the
@@ -177,7 +125,7 @@ export class GithubTarget extends BaseTarget {
    * @param changes The changeset information for this release
    * @returns The newly created release
    */
-  public async getOrCreateRelease(
+  public async createReleaseDraft(
     version: string,
     revision: string,
     changes?: Changeset
@@ -187,56 +135,25 @@ export class GithubTarget extends BaseTarget {
     const isPreview =
       this.githubConfig.previewReleases && isPreviewRelease(version);
 
-    try {
-      const response = await this.github.repos.getReleaseByTag({
-        owner: this.githubConfig.owner,
-        repo: this.githubConfig.repo,
-        tag,
-      });
-      this.logger.warn(`Found the existing release for tag "${tag}"`);
-      return response.data;
-    } catch (e: any) {
-      if (e.status !== 404) {
-        throw e;
-      }
-      this.logger.debug(`Release for tag "${tag}" not found.`);
-    }
-
-    const createReleaseParams = {
-      draft: false,
-      name: tag,
-      owner: this.githubConfig.owner,
-      prerelease: isPreview,
-      repo: this.githubConfig.repo,
-      tag_name: tag,
-      target_commitish: revision,
-      ...changes,
-    };
-
-    this.logger.debug(`Annotated tag: ${this.githubConfig.annotatedTag}`);
-    if (!isDryRun()) {
-      if (this.githubConfig.annotatedTag) {
-        await this.createAnnotatedTag(version, revision, tag);
-        // We've just created the tag, so "target_commitish" will not be used.
-        createReleaseParams.target_commitish = '';
-      }
-
-      this.logger.info(
-        `Creating a new ${
-          isPreview ? '*preview* ' : ''
-        }release for tag "${tag}"`
-      );
-      const created = await this.github.repos.createRelease(
-        createReleaseParams
-      );
-      return created.data;
-    } else {
+    if (isDryRun()) {
       this.logger.info(`[dry-run] Not creating the release`);
       return {
         id: 0,
         tag_name: tag,
         upload_url: '',
       };
+    } else {
+      const created = await this.github.repos.createRelease({
+        draft: true,
+        name: tag,
+        owner: this.githubConfig.owner,
+        prerelease: isPreview,
+        repo: this.githubConfig.repo,
+        tag_name: tag,
+        target_commitish: revision,
+        ...changes,
+      });
+      return created.data;
     }
   }
 
@@ -292,7 +209,7 @@ export class GithubTarget extends BaseTarget {
    *
    * @param release Release to fetch assets from
    */
-  public async getAssetsForRelease(
+  public async getAssetsForReleaseDraft(
     release_id: number
   ): Promise<ReposListAssetsForReleaseResponseItem[]> {
     const assetsResponse = await this.github.repos.listReleaseAssets({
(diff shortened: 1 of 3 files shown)


chadwhitacre:
cc: @BYK :)

chadwhitacre:
As implemented, we always start a new draft with every `craft publish`, hence there is no more need to delete assets from a possibly already-existing release object. In the event of a failure the draft release will be left over and can be deleted manually.

src/targets/github.ts
```diff
@@ -187,56 +130,25 @@ export class GithubTarget extends BaseTarget {
...
+        name: tag,
+        owner: this.githubConfig.owner,
+        prerelease: isPreview,
+        repo: this.githubConfig.repo,
+        tag_name: tag,
+        target_commitish: revision,
+        ...changes,
+      });
+      return created.data;
     }
```
kamilogorek, line 152:
```suggestion

    const created = await this.github.repos.createRelease({
      draft: true,
      name: tag,
      owner: this.githubConfig.owner,
      prerelease: isPreview,
      repo: this.githubConfig.repo,
      tag_name: tag,
      target_commitish: revision,
      ...changes,
    });
    return created.data;
```

kamilogorek: Nitpick: I always prefer early returns for non-happy-path branches, where `else` is redundant. But it's just a visual difference, so I don't mind.

src/targets/github.ts
```diff
@@ -422,6 +334,24 @@ export class GithubTarget extends BaseTarget {
...
   }
 
+  /**
+   * Publishes the release draft.
+   *
+   * @param release_id Release object ID
+   */
+  public async publishReleaseDraft(release_id: number) {
+    if (isDryRun()) {
```
kamilogorek, line 343: Same here, but your call. We already have this style in the `deleteAsset` function above.

src/targets/github.ts
```diff
@@ -436,31 +366,12 @@ export class GithubTarget extends BaseTarget {
...
-        if (failed.length === 0) {
-          this.logger.debug(`Deleted ${assets.length} assets`);
-        } else {
-          this.logger.debug(
-            'Failed to delete some assets:',
-            failed.map(({ reason }) => reason)
-          );
-        }
-      }
+      throw new Error('Existing assets found for the release draft! Why?');
```
kamilogorek, line 374: Should we have a way to force the deletion/upload despite assets being somehow there in case of an incident?

kamilogorek (changes requested):
So much code removed 🥰

src/targets/github.ts
```diff
@@ -422,6 +334,24 @@ export class GithubTarget extends BaseTarget {
...
+  /**
+   * Publishes the release draft.
+   *
+   * @param release_id Release object ID
+   */
+  public async publishReleaseDraft(release_id: number) {
+    if (isDryRun()) {
+      this.logger.info(`[dry-run] Not publishing the release draft`);
+    } else {
+      await this.github.repos.updateRelease({
```
iker-barriocanal, line 346: Is having some protection of this method (like wrapping it in a `try..catch`) overengineering it? [Docs](https://octokit.github.io/rest.js/v18#repos-update-release) don't failure use cases when (e.g. release doesn't exist, an issue with GitHub...), but other methods [throw exceptions](https://github.com/getsentry/craft/pull/328/files#diff-5c75e39457a6946369f245a4d5672bd24e7bb641f645f1ae5412fe4f368cf145R437-R443).

src/targets/github.ts
chadwhitacre: Done in d7886d943388383dddec2db271e371d29488d4f8.

src/targets/github.ts
chadwhitacre: Done in d7886d943388383dddec2db271e371d29488d4f8.

src/targets/github.ts
chadwhitacre: This is really only here because we _definitely_ needed this check for the previous logic where we were creating a true release object each time. In the draft case, each call to `createRelease` starts a new draft, so this is really more of a sanity check than an expected case. Maybe we should remove it? 🤔

src/targets/github.ts
```diff
@@ -187,56 +130,25 @@ export class GithubTarget extends BaseTarget {
...
-    } else {
+    if (isDryRun()) {
       this.logger.info(`[dry-run] Not creating the release`);
       return {
         id: 0,
         tag_name: tag,
         upload_url: '',
       };
+    } else {
+      const created = await this.github.repos.createRelease({
```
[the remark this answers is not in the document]
chadwhitacre: A new draft release is created each time.

I tested with this patch against d7886d943388383dddec2db271e371d29488d4f8:

```diff
diff --git a/src/targets/github.ts b/src/targets/github.ts
index 65c277e..5617461 100644
--- a/src/targets/github.ts
+++ b/src/targets/github.ts
@@ -139,6 +139,8 @@ export class GithubTarget extends BaseTarget {
       };
     }
 
+    this.logger.warn('CALLING CREATE TWICE');
+
     const created = await this.github.repos.createRelease({
       draft: true,
       name: tag,
@@ -149,6 +151,22 @@ export class GithubTarget extends BaseTarget {
       target_commitish: revision,
       ...changes,
     });
+
+    const second = await this.github.repos.createRelease({
+      draft: true,
+      name: tag,
+      owner: this.githubConfig.owner,
+      prerelease: isPreview,
+      repo: this.githubConfig.repo,
+      tag_name: tag,
+      target_commitish: revision,
+      ...changes,
+    });
<img width="490" alt="Screen Shot 2022-02-08 at 12 39 11 PM" src="https://user-images.githubusercontent.com/134455/153044816-371b2d35-a464-4385-a45a-bf527abf2657.png">

+
+    this.logger.warn('CALLED CREATE TWICE');
+    this.logger.warn(created);
+    this.logger.warn(second);
+
     return created.data;
   }
 
```

Screen Shot 2022-02-08 at 12 39 11 PM

src/targets/github.ts
chadwhitacre: If the release doesn't exist we want to bail and find out about it, normal error handling seems fine.

If it's a transient network error I guess we would want to retry, but I'd rather not solve that as a one-off here, we shoudl address that across the board if we want.

src/targets/github.ts
```diff
@@ -187,57 +130,26 @@ export class GithubTarget extends BaseTarget {
...
+        id: 0,
+        tag_name: tag,
+        upload_url: '',
+      };
     }
 
-    const createReleaseParams = {
-      draft: false,
+    const created = await this.github.repos.createRelease({
+      draft: true,
```
BYK, line 143:
```suggestion
    const { data } = await this.github.repos.createRelease({
      draft: true,
```

BYK: Or even:
```suggestion
    return (await this.github.repos.createRelease({
      draft: true,
      ...
    })).data;
```

src/targets/github.ts
```diff
@@ -271,7 +183,7 @@ export class GithubTarget extends BaseTarget {
   ): Promise<boolean> {
     this.logger.debug(`Deleting asset: "${asset.name}"...`);
     if (isDryRun()) {
-      this.logger.info(`[dry-run] Not uploading deleting "${asset.name}"`);
```
BYK, line 274: lol

src/targets/github.ts
```diff
@@ -187,57 +130,26 @@ export class GithubTarget extends BaseTarget {
...
-      });
-      this.logger.warn(`Found the existing release for tag "${tag}"`);
-      return response.data;
-    } catch (e: any) {
-      if (e.status !== 404) {
-        throw e;
-      }
-      this.logger.debug(`Release for tag "${tag}" not found.`);
+    if (isDryRun()) {
+      this.logger.info(`[dry-run] Not creating the release`);
```
BYK, line 134: Since this is a draft, I think it might be actually good to create the draft even in dry mode. It doesn't do anything destructive.

src/targets/github.ts
```diff
@@ -422,6 +334,25 @@ export class GithubTarget extends BaseTarget {
     }
   }
 
+  /**
+   * Publishes the release draft.
+   *
+   * @param release_id Release object ID
+   */
+  public async publishReleaseDraft(release_id: number) {
+    if (isDryRun()) {
```
BYK, line 343: Feels like this pattern can use something like a [decorator](https://www.typescriptlang.org/docs/handbook/decorators.html):
```js
@noopInDryRun
```

src/targets/github.ts
```diff
@@ -422,6 +334,25 @@ export class GithubTarget extends BaseTarget {
...
+      this.logger.info(`[dry-run] Not publishing the release draft`);
+      return;
+    }
+
+    await this.github.repos.updateRelease({
+      owner: this.githubConfig.owner,
+      repo: this.githubConfig.repo,
```
BYK, line 350:
```suggestion
      ...this.githubConfig,
```

src/targets/github.ts
```diff
@@ -436,31 +367,12 @@ export class GithubTarget extends BaseTarget {
     if (config.changelogPolicy !== ChangelogPolicy.None) {
       changelog = await this.getChangelog(version);
     }
-    const release = await this.getOrCreateRelease(version, revision, changelog);
 
-    const assets = await this.getAssetsForRelease(release.id);
+    const draft = await this.createReleaseDraft(version, revision, changelog);
```
BYK, line 371:
```suggestion
    const draftRelease = await this.createReleaseDraft(version, revision, changelog);
```

src/targets/github.ts
```diff
@@ -473,8 +385,10 @@ export class GithubTarget extends BaseTarget {
 
     await Promise.all(
       localArtifacts.map(({ path, mimeType }) =>
-        this.uploadAsset(release, path, mimeType)
+        this.uploadAsset(draft, path, mimeType)
       )
     );
+
+    await this.publishReleaseDraft(draft.id);
```
BYK, line 392: I think this method should accept a `release` object like `uploadAssets` for consistency. (or vice versa)

src/targets/upm.ts
```diff
@@ -145,11 +145,12 @@ export class UpmTarget extends BaseTarget {
         } else {
           await git.push(['origin', 'main']);
           const changes = await this.githubTarget.getChangelog(version);
-          await this.githubTarget.getOrCreateRelease(
+          const draft = await this.githubTarget.createReleaseDraft(
```
BYK, line 148:
```suggestion
          const draftRelease = await this.githubTarget.createReleaseDraft(
```

BYK (approved):
Looks pretty great to me. Does this magically pass all existing tests? Is that how great(!) our test coverage is? 😀

src/targets/github.ts
chadwhitacre: I don't expect a dry run to have any side effects whatsoever, it should just log. 🐭

src/targets/github.ts
chadwhitacre: Done in 7a968f67abeba6dadb462224bba5b1ed4f6fa1dc. 👍

src/targets/github.ts
chadwhitacre: Done in 7a968f67abeba6dadb462224bba5b1ed4f6fa1dc. 👍

src/targets/github.ts
chadwhitacre: Done in 7a968f67abeba6dadb462224bba5b1ed4f6fa1dc. 👍

chadwhitacre:
Hrm, where did CI go? 🤔

chadwhitacre:
@kamilogorek Could I get a re-review here? 🙏

kamilogorek (approved):
Just a minor style change, but otherwise looks good!

Issue #306, opened 2021-10-05
Non-attended mode for the Maven target


marandaneto:
https://github.com/getsentry/craft/releases/tag/0.25.0 introduced the maven target used by the https://github.com/getsentry/sentry-java and https://github.com/getsentry/sentry-android-gradle-plugin repos, because of GPG signing, this is still a user attended mode since you've to generate a GPG key locally and type user/pass when releasing a new version via Craft.

this is a blocker to add these 2 repos into the new release flow via https://github.com/getsentry/publish

marandaneto:
Related to https://github.com/getsentry/sentry-java/issues/1490

Craft should be able to use the Sentry GPG key from the docker image or something when executing `mvn` commands, likely here:
https://github.com/getsentry/craft/blob/76fd10de9681e03cab59e30bd181ace14a6b8d85/src/targets/maven.ts#L349-L383
Passing the commands with the GPG data, so the release machine won't pop up to ask user and password.

PR #346, opened 2022-02-10
feat: Unattended GPG signing for Maven target

kamilogorek:


kamilogorek:
Closes https://github.com/getsentry/craft/issues/306

marandaneto (approved):
Thanks @kamilogorek 
LGTM.

src/utils/__tests__/gpg.test.ts
```diff
@@ -0,0 +1,40 @@
...
+jest.mock('fs', () => ({
+  ...jest.requireActual('fs'),
+  promises: {
+    writeFile: jest.fn(() => Promise.resolve()),
+    unlink: jest.fn(),
+  },
+}));
+
+describe('importGPGKey', () => {
+  const KEY = 'very_private_key_like_for_real_really_private';
```
iker-barriocanal, line 16: lol

src/targets/maven.ts
```diff
@@ -73,6 +78,10 @@ export class MavenTarget extends BaseTarget {
     super(config, artifactProvider);
     this.mavenConfig = this.getMavenConfig();
     this.checkRequiredSoftware();
+
+    if (process.env.GPG_PRIVATE_KEY) {
+      importGPGKey(process.env.GPG_PRIVATE_KEY);
+    }
```
iker-barriocanal, line 84: This check is already done in `getTargetSecrets`, line 102. At this point, if the env var exists it's available in the target config, so you should be able to just import it.

On the other hand, what's the error message if the `importGPGKey` fails? Is it descriptive enough?

iker-barriocanal (commented):
How does a Craft release behave when there aren't any GPG keys in the machine? Can we expect they are always imported correctly?

src/targets/maven.ts
kamilogorek: `getTargetSecrets` is making them required, where this is optional.

bruno-garcia:
Correction: `GPG_SECRET_KEY` is not the right env var. `GPG_PRIVATE_KEY` is

bruno-garcia:
https://github.com/getsentry/publish/pull/810#issuecomment-1035166333

PR #349, opened 2022-02-11
fix: Make sure that gradle directory exists before writing to it

kamilogorek:
There is no `$HOME/.gradle` directory by default in our docker instance, thus writing to `$HOME/.gradle/gradle.properties` thows an error.

### DIFF
--- a/src/targets/__tests__/maven.test.ts
+++ b/src/targets/__tests__/maven.test.ts
@@ -1,3 +1,4 @@
+import { promises as fsPromises } from 'fs';
 import { homedir } from 'os';
 import { join } from 'path';
 import { NoneArtifactProvider } from '../../artifact_providers/none';
@@ -19,6 +20,7 @@ jest.mock('fs', () => ({
   promises: {
     writeFile: jest.fn(() => Promise.resolve()),
     readFile: jest.fn((file: string) => file),
+    mkdir: jest.fn(() => Promise.resolve()),
     readdir: async () => Promise.resolve([]), // empty dir
     unlink: jest.fn(),
   },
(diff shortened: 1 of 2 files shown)


marandaneto (approved):

src/targets/maven.ts
```diff
@@ -232,7 +235,7 @@ export class MavenTarget extends BaseTarget {
   /**
    * Deletes the user's `gradle.properties` file.
    */
-  public deleteUserGradlePropsFile(): Promise<void> {
+  public async deleteUserGradlePropsFile(): Promise<void> {
```
iker-barriocanal, line 238: This `async` isn't required, is it?

src/targets/__tests__/maven.test.ts
```diff
@@ -351,3 +353,24 @@ describe('get gradle home directory', () => {
...
+    delete process.env[gradleHomeEnvVar];
+  });
+
+  test('should make sure that directory exists before writing gradle props file', async () => {
+    const randomTmpDir = '/random/depth/of/directories';
+    process.env[gradleHomeEnvVar] = randomTmpDir;
+    await createMavenTarget().createUserGradlePropsFile();
+    expect(fsPromises.mkdir).toHaveBeenCalledWith(randomTmpDir, {
+      recursive: true,
+    });
```
iker-barriocanal, line 370: Considering there's a helper [`withTempDir`](https://github.com/getsentry/craft/blob/master/src/utils/files.ts#L76) function, I think it'd be better to not be that implementation-specific and abstract how the method is behaving. With an actual temporary directory, test whether the gradle props file exists after the method has been executed (and two use cases here: when the gradle directory exists, and when it doesn't). This approach doesn't rely on implementation details.

src/targets/__tests__/maven.test.ts
```diff
@@ -351,3 +353,24 @@ describe('get gradle home directory', () => {
...
+    expect(fsPromises.writeFile).toHaveBeenCalledWith(
+      `${randomTmpDir}/gradle.properties`,
+      `mavenCentralUsername=my_default_value\nmavenCentralPassword=my_default_value`
```
iker-barriocanal, line 373: The values of both the username and password are coming from `DEFAULT_OPTION_VALUE`. Can we use that, instead of hardcoding the values? Using a snapshot is another option.

src/targets/maven.ts
kamilogorek: Its not only due to linter setting, but I prefer having explicit `async` when returning a promise. Easier to read imo.

src/targets/__tests__/maven.test.ts
kamilogorek: Agree, although its not that trivial to do with all the mocking done in this file. I can improve on those once we get this thing working in the first place. Appreciate the after-closing review!

PR #351, opened 2022-02-11
feat: Publish maven packages without use of Gradle

kamilogorek:
This PR utilizes direct API calls to close and publish packages to Nexus, instead of going through a 3rd party Gradle plugin.

Tests and verification are still TBD, but it's already EOD, and wanted to get this PR out for an initial review.

This change will also make https://github.com/getsentry/craft/pull/278 obsolete.

### DIFF
--- a/src/targets/__tests__/maven.test.ts
+++ b/src/targets/__tests__/maven.test.ts
@@ -1,6 +1,3 @@
-import { promises as fsPromises } from 'fs';
-import { homedir } from 'os';
-import { join } from 'path';
 import { NoneArtifactProvider } from '../../artifact_providers/none';
 import {
   MavenTarget,
@@ -53,7 +50,6 @@ function getFullTargetConfig(): any {
     GPG_PASSPHRASE: DEFAULT_OPTION_VALUE,
     OSSRH_USERNAME: DEFAULT_OPTION_VALUE,
     OSSRH_PASSWORD: DEFAULT_OPTION_VALUE,
-    gradleCliPath: DEFAULT_OPTION_VALUE,
     mavenCliPath: DEFAULT_OPTION_VALUE,
     mavenSettingsPath: DEFAULT_OPTION_VALUE,
     mavenRepoId: DEFAULT_OPTION_VALUE,
@@ -71,7 +67,6 @@ function getRequiredTargetConfig(): any {
     GPG_PASSPHRASE: DEFAULT_OPTION_VALUE,
     OSSRH_USERNAME: DEFAULT_OPTION_VALUE,
     OSSRH_PASSWORD: DEFAULT_OPTION_VALUE,
-    gradleCliPath: DEFAULT_OPTION_VALUE,
     mavenCliPath: DEFAULT_OPTION_VALUE,
     mavenSettingsPath: DEFAULT_OPTION_VALUE,
     mavenRepoId: DEFAULT_OPTION_VALUE,
@@ -105,7 +100,7 @@ describe('Maven target configuration', () => {
 
   test('env vars without options', () => {
     expect(() => createMavenTarget({})).toThrowErrorMatchingInlineSnapshot(
-      `"Required configuration gradleCliPath not found in configuration file. See the documentation for more details."`
+      `"Required configuration mavenCliPath not found in configuration file. See the documentation for more details."`
     );
   });
 
@@ -203,38 +198,16 @@ describe('publish', () => {
   test('main flow', async () => {
     const callOrder: string[] = [];
     const mvnTarget = createMavenTarget();
-    const createGradlePropsMock = jest.fn(
-      async () => void callOrder.push('createGradleProps')
+    mvnTarget.upload = jest.fn(async () => void callOrder.push('upload'));
+    mvnTarget.closeAndReleaseRepository = jest.fn(
+      async () => void callOrder.push('closeAndReleaseRepository')
     );
-    mvnTarget.createUserGradlePropsFile = createGradlePropsMock;
-    const deleteGradlePropsMock = jest.fn(
-      async () => void callOrder.push('deleteGradleProps')
-    );
-    mvnTarget.deleteUserGradlePropsFile = deleteGradlePropsMock;
-    const uploadMock = jest.fn(async () => void callOrder.push('upload'));
-    mvnTarget.upload = uploadMock;
-    (retrySpawnProcess as jest.MockedFunction<
-      typeof retrySpawnProcess
-    >).mockImplementationOnce(
-      async () => void callOrder.push('closeAndRelease')
-    );
-
     const revision = 'r3v1s10n';
     await mvnTarget.publish('1.0.0', revision);
-    expect(createGradlePropsMock).toHaveBeenCalledTimes(1);
-    expect(uploadMock).toHaveBeenCalledTimes(1);
-    expect(uploadMock).toHaveBeenLastCalledWith(revision);
-    expect(deleteGradlePropsMock).toHaveBeenCalledTimes(1);
-    expect(retrySpawnProcess).toHaveBeenCalledTimes(1);
-    expect(retrySpawnProcess).toHaveBeenCalledWith(DEFAULT_OPTION_VALUE, [
-      'closeAndReleaseRepository',
-    ]);
-    expect(callOrder).toStrictEqual([
-      'createGradleProps',
-      'upload',
-      'closeAndRelease',
-      'deleteGradleProps',
-    ]);
+    expect(mvnTarget.upload).toHaveBeenCalledTimes(1);
+    expect(mvnTarget.upload).toHaveBeenLastCalledWith(revision);
+    expect(mvnTarget.closeAndReleaseRepository).toHaveBeenCalledTimes(1);
+    expect(callOrder).toStrictEqual(['upload', 'closeAndReleaseRepository']);
   });
 
   test('upload POM', async () => {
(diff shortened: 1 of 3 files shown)


src/targets/maven.ts
```diff
@@ -4,28 +4,31 @@ import {
...
- * https://docs.gradle.org/current/userguide/build_environment.html#sec:gradle_environment_variables
- */
-const DEFAULT_GRADLE_USER_HOME = join(homedir(), '.gradle');
 export const POM_DEFAULT_FILENAME = 'pom-default.xml';
 const POM_FILE_EXT = '.xml'; // Must include the leading `.`
 const BOM_FILE_KEY_REGEXP = new RegExp('<packaging>pom</packaging>');
 
+const NEXUS_API_BASE_URL = 'https://oss.sonatype.org/service/local/staging';
+const NEXUS_RETRY_DELAY = 10 * 1000; // 10s
+const NEXUS_RETRY_DEADLINE = 15 * 60 * 1000; // 15min
```
bruno-garcia, line 24: Nexus can be pretty flaky so might want to bump that to 30

bruno-garcia (approved):

src/targets/maven.ts
```diff
@@ -444,4 +391,105 @@ export class MavenTarget extends BaseTarget {
...
+      );
+    }
+
+    const poolingStartTime = Date.now();
+
+    while (true) {
+      if (Date.now() - poolingStartTime > NEXUS_RETRY_DEADLINE) {
+        throw new Error('Deadline for Nexus repository status change reached.');
+      }
+
```
kamilogorek, line 468: Note to self: add debug log message here

src/targets/maven.ts
```diff
@@ -444,4 +391,105 @@ export class MavenTarget extends BaseTarget {
 
     return `${moduleName}.jar`;
   }
+
+  // Maven central does not indicate when it completes the action, so we need to
+  // retry every so often and query it for the new state of repository.
+  // Based on: https://github.com/vanniktech/gradle-maven-publish-plugin/ implementation.
+  public async closeAndReleaseRepository(): Promise<void> {
+    const repository = await this.getRepository();
+    const { repositoryId, type } = repository;
```
iker-barriocanal, line 400:
```suggestion
    const { repositoryId, type } = await this.getRepository();
```

src/targets/maven.ts
```diff
@@ -444,4 +391,105 @@ export class MavenTarget extends BaseTarget {
...
+
+    await this.closeRepository(repositoryId);
+    await this.releaseRepository(repositoryId);
+  }
+
+  private getNexusRequestHeaders(): Record<string, string> {
+    return {
+      Accept: 'application/json',
+      Authorization: `Basic ${Buffer.from(
+        `${process.env.OSSRH_USERNAME}:${process.env.OSSRH_USERNAME}`
```
iker-barriocanal, line 416: Environment variables may change. We should pick these from the target settings (`this.mavenConfig.OOSH_USERNAME`) to avoid such cases.

src/targets/maven.ts
```diff
@@ -444,4 +391,105 @@ export class MavenTarget extends BaseTarget {
...
+      ).toString(`base64`)}`,
```
iker-barriocanal, line 417: Why parsing to base64?

src/targets/maven.ts
```diff
@@ -444,4 +391,105 @@ export class MavenTarget extends BaseTarget {
...
+      throw new Error(
+        `Unable to fetch repository: ${response.status}, ${response.statusText}`
+      );
+    }
+
+    const body = await response.json();
+    const repositories = body.data;
+
+    if (repositories.length === 0) {
+      throw new Error(`No available repositories. Nothing to publish.`);
```
iker-barriocanal, line 435: nit:
```suggestion
      throw new Error(`No available repositories. Nothing to publish to.`);
```
?

src/targets/maven.ts
```diff
@@ -444,4 +391,105 @@ export class MavenTarget extends BaseTarget {
...
+
+    if (!response.ok) {
+      throw new Error(
+        `Unable to close repository: ${response.status}, ${response.statusText}`
+      );
+    }
+
+    const poolingStartTime = Date.now();
+
+    while (true) {
```
iker-barriocanal, line 464: Can we benefit from [`withRetry`](https://github.com/getsentry/craft/blob/2a0eda0b9921dca762090b3fa357e03b56c2f66b/src/utils/async.ts#L155) in this whole block?

src/targets/maven.ts
```diff
@@ -444,4 +391,105 @@ export class MavenTarget extends BaseTarget {
...
+      if (Date.now() - poolingStartTime > NEXUS_RETRY_DEADLINE) {
+        throw new Error('Deadline for Nexus repository status change reached.');
+      }
+
+      await sleep(NEXUS_RETRY_DELAY);
+
+      const repository = await this.getRepository();
+      const { type, transitioning } = repository;
```
iker-barriocanal, line 472:
```suggestion
      const { type, transitioning } = await this.getRepository();
```

src/targets/maven.ts
```diff
@@ -444,4 +391,105 @@ export class MavenTarget extends BaseTarget {
...
+      headers: this.getNexusRequestHeaders(),
+      method: 'POST',
+      body: JSON.stringify({
+        data: { stagedRepositoryIds: [repositoryId] },
+      }),
+    });
+
+    if (!response.ok) {
+      throw new Error(
+        `Unable to release repository: ${response.status}, ${response.statusText}`
```
iker-barriocanal, line 491: Should we include the repository ID? Does it help when manual action is required? Is there any security reason to not expose it?

src/targets/maven.ts
kamilogorek: We publish "a thing" (artifacts already present at this repository), not "to something", so I believe this is correct?

src/targets/maven.ts
kamilogorek: We could if those servers would be slightly more reliable. `withRetry` is based on a number of attempts, whereas here, we are using time-based deadline. With attempts-based and this servers, it could take between 10 and 100min for the same number of attempts 🥲

src/targets/maven.ts
kamilogorek: There can be only a single repository during this process, so it's not possible when observed instantly, however it can be useful for reference when you want to do something after a weekend or something. Added.

src/targets/maven.ts
kamilogorek: Missed it, will patch in the next commit.

src/targets/maven.ts
kamilogorek: Basic Auth protocol is using base64 - https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#basic_authentication_scheme

src/targets/__tests__/maven.test.ts
```diff
@@ -457,7 +457,9 @@ describe('closeRepository', () => {
 
     const mvnTarget = createMavenTarget();
     await expect(mvnTarget.closeRepository(repositoryId)).rejects.toThrow(
-      new Error('Unable to close repository: 500, Internal Server Error')
+      new Error(
+        'Unable to close repository sentry-java: 500, Internal Server Error'
```
iker-barriocanal, line 461: The repository name must be taken from the config. `sentry-java` is not the only repo using this target (see [`sentry-android-gradle-plugin`](https://github.com/getsentry/sentry-android-gradle-plugin/blob/464f228a42116ee995bda14e6fdae05398ec3725/.craft.yml#L4)), and there may be more in the future.

src/targets/__tests__/maven.test.ts
```diff
@@ -563,7 +565,9 @@ describe('releaseRepository', () => {
 
     const mvnTarget = createMavenTarget();
     await expect(mvnTarget.releaseRepository(repositoryId)).rejects.toThrow(
-      new Error('Unable to release repository: 500, Internal Server Error')
+      new Error(
+        'Unable to release repository sentry-java: 500, Internal Server Error'
```
iker-barriocanal, line 569: Same here for the repo name.

src/targets/__tests__/maven.test.ts
```diff
@@ -303,3 +325,245 @@ describe('publish', () => {
...
+      Promise.resolve(getRepositoryInfo('open', false))
+    );
+    mvnTarget.closeRepository = jest.fn(() => Promise.reject());
+    mvnTarget.releaseRepository = jest.fn(() => Promise.resolve(true));
+
+    try {
+      await mvnTarget.closeAndReleaseRepository();
+    } catch (e) {
+      // no-empty
+    }
```
iker-barriocanal, line 372: Why not something like `expect(...).not.toThrow()`?

src/targets/__tests__/maven.test.ts
kamilogorek: It's take from the registry itself, those are just tests

src/targets/__tests__/maven.test.ts
kamilogorek: Because we actually want to throw here

src/targets/maven.ts
iker-barriocanal: Yes, we publish artifacts; but, we throw the error when there's no repository to publish the artifacts to. The error is not about the artifacts, but about where we publish them. This is clear in the first sentence, and thus a bike-shedding discussion, so your call.

src/targets/__tests__/maven.test.ts
kamilogorek: Because imo assertions should match the name of the test.

> should not release repostiory if it was not closed properly

And assertion is that `releaseRepository` was never called. `closeAndReleaseRepository` is only a setup trigger.

BYK: >Because imo assertions should match the name of the test

Yeah, that makes sense.

>And assertion is that releaseRepository was never called. closeAndReleaseRepository is only a setup trigger.

I think a comment here instead of `// on-empty` would go a long way to communicate this to other developers such as myself and @iker-barriocanal 🙂

kamilogorek: Fair point, will keep that in mind :)

PR #352, opened 2022-02-14
fix: Correctly set auth header and content-type for Nexus requests

kamilogorek:
Such a mistake 🥲

### DIFF
--- a/src/targets/maven.ts
+++ b/src/targets/maven.ts
@@ -464,13 +464,14 @@ export class MavenTarget extends BaseTarget {
       const { type, transitioning } = await this.getRepository();
 
       if (type === 'closed' && !transitioning) {
+        this.logger.info(`Nexus repository close correctly.`);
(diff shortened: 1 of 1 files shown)


src/targets/maven.ts
```diff
@@ -490,14 +491,16 @@ export class MavenTarget extends BaseTarget {
       );
     }
 
+    this.logger.info(`Nexus repository closed correctly.`);
     return true;
   }
 
   private getNexusRequestHeaders(): Record<string, string> {
     return {
+      'Content-Type': 'application/json',
```
kamilogorek, line 500: Nexus API is using `Accept` is for get requests, `Content-Type` for post requests, and it needs both.

iker-barriocanal: I'd add this as a comment.

iker-barriocanal (approved):

Issue #327, opened 2021-11-16
Last step of publish is selecting the wrong merge target


chadwhitacre:
After a release we merge the release branch back to the base branch:

https://github.com/getsentry/craft/blob/78d2ca9c4bae075ce45ca888b94c86e176d9848f/src/commands/publish.ts#L331-L363

Unfortunately it looks like our algorithm for determining which branch to merge into if not explicitly provided is wrong. The past two `sentry` releases have merged back into the wrong branch:

https://github.com/getsentry/publish/runs/4218472116?check_suite_focus=true#step:8:123 &rarr; https://github.com/getsentry/sentry/commit/9ae170c4b3b2fd608caabb37e9ce08281d5e4747

https://github.com/getsentry/publish/runs/3909141316?check_suite_focus=true#step:8:666 &rarr;
https://github.com/getsentry/sentry/commit/b15920f0c741c35f70dab85b3b2b76ffa0317602

The result is that (at least) the CHANGES file in the sentry repo has not been updated since 21.9.0 ... we're missing all of the new changelog-y goodness!

https://github.com/getsentry/sentry/blob/master/CHANGES

chadwhitacre:
I'm pretty sure this is because `master` moves so quickly.

<img width="978" alt="Screen Shot 2021-12-15 at 10 35 43 AM" src="https://user-images.githubusercontent.com/134455/146245296-3fb2a078-e818-4a13-8224-417d040badd6.png">

chadwhitacre:
There is a CLI option. Can we use that for `sentry`? 🧐

https://github.com/getsentry/craft/blob/95677da7d2f8c15df19cf1bd6369bdeb77209b3b/src/commands/publish.ts#L75-L80

chadwhitacre:
Hrm, [not easily](https://github.com/getsentry/publish/blob/b229c878e97fe7529e809b1646fe5f57bab0d9aa/.github/workflows/publish.yml#L57). 🤔

BYK:
@chadwhitacre you can probably make "Uses the default branch from GitHub as a fallback" case the default and move the current logic into `-m auto`?

chadwhitacre:
@BYK That sounds more reasonable to me. What was the use-case for the `auto` strategy in the first place? Are there other craft users/projects that depend on that? 🤔

BYK:
>@BYK That sounds more reasonable to me. What was the use-case for the auto strategy in the first place? Are there other craft users/projects that depend on that? 🤔

[...]

PR #355, opened 2022-02-17
ref: Use default branch as merge target instead of parent detection

kamilogorek:
This is WiP to fix https://github.com/getsentry/craft/issues/327

We remove automatic branch detection, use the default as the main source of truth, but allow for overriding it through the GH Action input (this needs to be added per repository, inside `.github/workflows/release.yml` in `with` clause).

Related required issues:
https://github.com/getsentry/action-prepare-release/pull/23
https://github.com/getsentry/publish/pull/838

### DIFF
--- a/src/commands/publish.ts
+++ b/src/commands/publish.ts
@@ -35,7 +35,7 @@ import { isValidVersion } from '../utils/version';
 import { BaseStatusProvider } from '../status_providers/base';
 import { BaseArtifactProvider } from '../artifact_providers/base';
 import { SimpleGit } from 'simple-git';
-import { getGitClient, getDefaultBranch, stripRemoteName } from '../utils/git';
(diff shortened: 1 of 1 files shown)


kamilogorek:
cc @BYK

BYK:
This will require reopening #239. I think we can still consider using `git merge-target` and then find the branch name from there?

kamilogorek:
https://github.com/getsentry/publish/pull/873

lobsterkatie:
> this needs to be added per repository, inside .github/workflows/release.yml in with clause

Specifically what needs to be added? Could you link to an example here for individual repo maintainers to follow?

kamilogorek:
@lobsterkatie I updated the readme yesterday here - https://github.com/getsentry/publish#merge-target

PR #353, opened 2022-02-15
feat: Adds pub.dev target for Dart and Flutter

kamilogorek:
Closes https://github.com/getsentry/craft/issues/264

PR #357, opened 2022-02-21
ci: Change machineType for Google Cloud Build to E2_HIGHCPU_8

kamilogorek:
This gives our machine more memory, which prevents `"gcr.io/kaniko-project/executor:v1.5.1" failed: step exited with non-zero status: 137` errors after adding too many new targets.


kamilogorek:
Merged in https://github.com/getsentry/craft/pull/353 🙄

BYK:
@kamilogorek interesting, we may wanna replace these ones with E2-class instances too then: https://github.com/getsentry/sentry/blob/037ce428a4f60f39b891f410be0904187b05f2b8/docker/cloudbuild.yaml#L97 (they seem equivalent but E2 is cheaper).

PR #375, opened 2022-04-01
Use a custom action

chadwhitacre:
Easier to iterate this way

### DIFF
--- a/.github/workflows/enforce-license-compliance.yml
+++ b/.github/workflows/enforce-license-compliance.yml
@@ -8,31 +8,7 @@ on:
 
 jobs:
   enforce-license-compliance:
-    if: github.repository_owner == 'getsentry'
     runs-on: ubuntu-latest
     steps:
-      - name: 'Fetch FOSSA_API_KEY'
-        id: fetch_key
-        run: |
-
-          # We can't use GitHub Secrets for this key because we want it to be
-          # available in forks. This is a push-only key for a low-privilege
-          # account, so it is safe (enough) to expose publicly. This is a hack
-          # to set it in one place in case we ever do need to roll it. 🤷
-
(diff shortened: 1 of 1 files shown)


.github/workflows/enforce-license-compliance.yml
```diff
@@ -8,31 +8,7 @@ on:
...
-          # to set it in one place in case we ever do need to roll it. 🤷
-
-          curl -O https://raw.githubusercontent.com/getsentry/.github/main/.github/workflows/FOSSA_API_KEY
-          echo "::set-output name=key::$(cat FOSSA_API_KEY | grep -v '#')"
-
-      - name: 'Checkout Code'
-        uses: actions/checkout@v2
-
-      - name: 'Run FOSSA Scan'
-        uses: fossas/fossa-action@v1.1.0
```
billyvg, line 30: You should pin these to a commit hash

kamilogorek (approved):

billyvg (approved):

.github/workflows/enforce-license-compliance.yml
kamilogorek: Then it's not making it any easier to iterate tbh :P

chadwhitacre: Will do that over in the action.

chadwhitacre: Done in https://github.com/getsentry/action-enforce-license-compliance/commit/b4fde989274b441d2b747d9b19067fe329bdf432.

> Then it's not making it any easier to iterate

This is referring to the upstream `fossas/fossa-action` that we don't control. References to `getsentry/action-enforce-license-compliance` can use a movable pointer such as `main` on this PR (I'm thinking for other repos to point to a `latest` branch/tag).

PR #379, opened 2022-04-07
feat(changelog): Limit changes to current project folder

BYK:
This change will limit the generated changelog entries to the current working directory, supporting monorepos with multiple different Craft configurations in them. It should have no effect on existing multi-repo use cases.


BYK:
@kamilogorek turns out that lint thing did not get fixed :D I think I have enough info to fix it tho so stay tuned.

kamilogorek (approved):

Issue #384, opened 2022-04-22
Craft gets stuck while uploading GH artifacts


jan-auer:
The publish action on sentry-native frequently gets stuck while uploading assets to GitHub. The last logs read (terminating with cancellation or timeout):

```
[info] [[target/github]] Git tag: "0.4.17"
[debug] POST /repos/getsentry/sentry-native/releases - 201 in 346ms
- Uploading asset "sentry-native.zip" to getsentry/sentry-native:0.4.17
[debug] POST /repos/getsentry/sentry-native/releases/65043191/assets?name=sentry-native.zip& - 500 in 231ms
Error: The operation was canceled.
```

Note the status code `500` response from the upload. Assuming we can trust the logs from GitHub Actions, there's something that keeps craft running in this state rather than retrying/failing.

asottile-sentry:
I believe this was resolved by #398

PR #385, opened 2022-04-22
ref: Remove spinners from github target artifacts upload

kamilogorek:
Due to `ora` not being able to handle multiple spinners, there's a suspicion that it interrupts our main `Promise.all` handler in some way, or with the retry process. This makes some releases somehow stuck. Funny enough, in 99% of cases they work on the first try during retry.

Before swapping `ora` for something else, I believe it's way easier to verify that this is the culprit by simply removing it entirely and leaving just the text itself. In the CI, where we look at those logs, there are no spinners anyway, and everything is logged one by one.

I removed `Verifying asset` text, as it looks redundant to me, due to a very clear error message in case something goes wrong. With the happy path, it provides no value.

Also used `process.stderr.write` instead of `console.error` (which is an alias for `process.stderr.write` with a new-line added) to make it more obvious that it's not a left out by an accident `console` call. It also mimics `ora` behavior 1:1.

A possible fix for https://github.com/getsentry/craft/issues/384

cc @BYK

### DIFF
--- a/src/targets/github.ts
+++ b/src/targets/github.ts
@@ -20,7 +20,6 @@ import { isPreviewRelease, versionToTag } from '../utils/version';
 import { BaseTarget } from './base';
 import { BaseArtifactProvider } from '../artifact_providers/base';
 import { logger } from '../logger';
-import ora from 'ora';
 
 /**
  * Default content type for GitHub release assets.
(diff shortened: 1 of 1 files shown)


BYK (approved):
>Before swapping ora for something else, I believe it's way easier to verify that this is the culprit by simply removing it entirely and leaving just the text itself. In the CI, where we look at those logs, there are no spinners anyway, and everything is logged one by one.

💯 

Let's also remove `ora` from `package.json` too?

kamilogorek:
It's still used for manual status provider pooling, but there we always have only one await, so it was safe to leave.

Swatinem (approved):

PR #386, opened 2022-04-26
ref: Remove @octokit/plugin-retry and rely on our own logic

kamilogorek:
We do already have our own logic implemented here - https://github.com/getsentry/craft/blob/2d8ba63ab9670d9b20c8c9e1a082ab60175c2594/src/targets/github.ts#L297-L333

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,6 @@
   "devDependencies": {
     "@aws-sdk/client-lambda": "^3.2.0",
     "@google-cloud/storage": "^5.7.0",
-    "@octokit/plugin-retry": "^3.0.9",
     "@octokit/request-error": "^2.1.0",
     "@octokit/rest": "^18.10.0",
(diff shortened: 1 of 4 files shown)


BYK (changes requested):
Don't think removing octokit-retry altogether is wise as we rely on that in other places.

The additional logic we have only handles `422` errors specifically which are not handled by `octokit-retry`: https://github.com/getsentry/craft/blob/master/src/targets/github.ts#L297-L333

kamilogorek:
Reached out to GitHub support, let’s put this PR aside for now.

mattgauntseo-sentry:
@kamilogorek is this PR still relevant / worth exploring further?

kamilogorek:
Nope, resolved by https://github.com/getsentry/craft/pull/398
Closed and branch deleted.

PR #387, opened 2022-04-26
GitHub is sunsetting git.io

chadwhitacre:
https://github.blog/changelog/2022-04-25-git-io-deprecation/

### DIFF
--- a/src/targets/github.ts
+++ b/src/targets/github.ts
@@ -310,7 +310,7 @@ export class GitHubTarget extends BaseTarget {
       // This usually happens when the upload gets interrupted somehow with a
       // 5xx error. See the docs here:
       // https://docs.github.com/en/rest/releases/assets#upload-a-release-asset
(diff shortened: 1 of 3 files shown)


src/targets/github.ts
```diff
@@ -308,8 +308,9 @@ export class GitHubTarget extends BaseTarget {
       }
 
       // This usually happens when the upload gets interrupted somehow with a
-      // 5xx error. See the docs here: https://git.io/JKZot
-      const isAssetExistsError =
+      // 5xx error. See the docs here:
+      // https://docs.github.com/en/rest/releases/assets#upload-a-release-asset
+      cons t isAssetExistsError =
```
kamilogorek, line 313:
```suggestion
      const isAssetExistsError =
```

kamilogorek (changes requested):

kamilogorek (approved):

PR #389, opened 2022-04-28
ref: Upload GitHub artifacts in series

kamilogorek:
GH Support suggested that our issues might be caused by too many requests in quick succession.
This change would verify this claim, but I'm still discussing this possibility with them.

### DIFF
--- a/src/targets/github.ts
+++ b/src/targets/github.ts
@@ -379,11 +379,12 @@ export class GitHubTarget extends BaseTarget {
       revision,
       changelog
     );
-    await Promise.all(
-      localArtifacts.map(({ path, mimeType }) =>
-        this.uploadAsset(draftRelease, path, mimeType)
-      )
-    );
+
+    // We don't want to do this in parallel but in serial, because the GitHub
+    // artifacts endpoint has a tendency to randomly timeout on multiple requests.
+    for (const { path, mimeType } of localArtifacts) {
+      await this.uploadAsset(draftRelease, path, mimeType);
+    }
 
     await this.publishRelease(draftRelease);
   }


BYK:
@kamilogorek `octokit-retry` uses a library called [`bottleneck`](https://github.com/SGrondin/bottleneck). If this turns out to be the case, maybe we can use it here to limit things instead of going serial all the way?

kamilogorek:
We sure can, but I've already seen a failed release with just 2 files being uploaded, sentry-cocoa for example:

failed: https://github.com/getsentry/publish/runs/6178545661?check_suite_focus=true
succeeded: https://github.com/getsentry/publish/runs/6193773150?check_suite_focus=true

If that'll be the case of rapid requests, then it'd mean that we'll have to go full in-series mode :(

src/targets/github.ts
```diff
@@ -379,11 +379,12 @@ export class GitHubTarget extends BaseTarget {
...
-    await Promise.all(
-      localArtifacts.map(({ path, mimeType }) =>
-        this.uploadAsset(draftRelease, path, mimeType)
-      )
-    );
+
+    // We don't want to do this in parallel but in serial, because the GitHub
+    // artifacts endpoint has a tendency to randomly timeout on multiple requests.
+    for (const { path, mimeType } of localArtifacts) {
+      await this.uploadAsset(draftRelease, path, mimeType);
```
AbhiPrasad, line 386: Do we know how much this will potentially slow down stuff by?

AbhiPrasad (approved):

src/targets/github.ts
kamilogorek: I don't have any specific number. However, we don't publish any repositories that contain more than 10-20 artifacts to GitHub afaik.

PR #390, opened 2022-04-29
ref: Allow for all logger levels in Octokit

kamilogorek:
This will allow us to see debug output of requests going out to GitHub.

### DIFF
--- a/src/utils/githubApi.ts
+++ b/src/utils/githubApi.ts
@@ -1,6 +1,6 @@
 import { Octokit } from '@octokit/rest';
 
-import { LogLevel, logger } from '../logger';
+import { logger } from '../logger';
 
 import { ConfigurationError } from './errors';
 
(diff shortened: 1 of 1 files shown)


src/utils/githubApi.ts
```diff
@@ -103,14 +103,9 @@ export function getGitHubClient(token = ''): Octokit {
   if (!_GitHubClientCache[githubApiToken]) {
     const attrs: any = {
       auth: `token ${githubApiToken}`,
+      log: logger,
```
BYK, line 106: Wow, have you tested this and it is _this_ easy? 😅

BYK (approved):

src/utils/githubApi.ts
kamilogorek: Yup, that's it! :D `consola` provides all the APIs that `octokit` requires - https://octokit.github.io/rest.js/v18#logging

AbhiPrasad (approved):

PR #392, opened 2022-05-10
fix(crates): Resume workspace publish

jan-auer:
The crates target is not able to resume a workspace publish because already published crates cannot be republished. This means a partially successful publish cannot be retried.

To work around this, the `crates` target can skip publishing a crate if it already exists in this version. The downside of this is that we would silence errors when publishing the wrong version.

See https://github.com/getsentry/publish/issues/1070#issuecomment-1122077509

### DIFF
--- a/src/targets/crates.ts
+++ b/src/targets/crates.ts
@@ -26,6 +26,13 @@ const CARGO_BIN = process.env.CARGO_BIN || DEFAULT_CARGO_BIN;
 const VERSION_ERROR = 'failed to select a version for the requirement';
 
 /**
+ * A message fragment emitted by cargo when publishing fails because the crate
+ * had already been published in this exact version. This happens especially
+ * when rerunning a workspace publish after it has failed in the middle.
+ */
+const REPUBLISH_ERROR = 'is already uploaded';
+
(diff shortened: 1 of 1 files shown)


Swatinem (approved):

kamilogorek (approved):

BYK:
I'd argue a better way to handle this is to have multiple `crates` targets per package.

BYK:
Or at least not make this behavior the default and hide behind a flag.

jan-auer:
Having multiple `crates` targets is pretty much not an option here. In workspaces, you frequently have dozens of crates which you would then have to order topological by their dependency graph.

We could also add this as a flag, @chadwhitacre is there an easy way to pass a "this is a retry" flag to craft already?

BYK:
>Having multiple crates targets is pretty much not an option here. In workspaces, you frequently have dozens of crates which you would then have to order topological by their dependency graph. That's too much manual work every time, unfortunately.

I never said "manual" 😁 I think we can expand the publish state file format to have this extra metadata.

kamilogorek (approved):
👀

PR #397, opened 2022-05-24
Redo retries

kamilogorek:
This PR introduces 3 changes, 2 of which are reversals of previous attempts to debug the GH upload issue https://github.com/getsentry/craft/pull/389 and https://github.com/getsentry/craft/pull/390

```
- ref: Rework how GH artifacts upload retrying works
- ref: Disable Octokit debug logging as it's not useful
- ref: Upload GitHub artifacts in parallel
```

The main change is how GH artifact upload requests are handled:
- skip `@octokit/retry` as it cannot perform any intermediate actions (deletion attempt)
- use our own internal retry counter instead
- set request timeout to 10min, which should be more than enough for any request to come through, and will give us enough time to perform 3 attempts (as our publish workflow timeout is currently 60min)
- always attempt to delete an artifact before retrying, and warn instead of erroring, if the artifact is not there


src/targets/github.ts
```diff
@@ -228,11 +209,12 @@ export class GitHubTarget extends BaseTarget {
...
     const assetToDelete = assets.find(({ name }) => name === assetName);
     if (!assetToDelete) {
-      throw new Error(
-        `No such asset with the name "${assetToDelete}". We have these instead: ${assets.map(
+      logger.warn(
+        `No such asset with the name "${assetToDelete}", moving on. We have these instead: ${assets.map(
           ({ name }) => name
         )}`
       );
+      return false;
```
BYK, line 217: I'd move listing of assets to a `logger.trace` as it can be very noisy as a warning.

src/utils/githubApi.ts
```diff
@@ -103,9 +103,17 @@ export function getGitHubClient(token = ''): Octokit {
   if (!_GitHubClientCache[githubApiToken]) {
     const attrs: any = {
       auth: `token ${githubApiToken}`,
-      log: logger,
     };
 
+    // Silence debug logs, as they do not provide any useful information
+    // about the requests, yet they are very noisy and make it difficult
```
BYK, line 109: 🤣 So I guess this is why it was done this way in the past but now with the comment, it is much much clear and hopefully nobody will reinvent that wheel.

asottile-sentry (approved):

src/targets/github.ts
kamilogorek: I left it as `warn` to see if we stumble upon this branch now. We can move it to `trace` once this issue is resolved.

PR #398, opened 2022-05-24
fix: re-open asset when retrying asset upload

asottile-sentry:
when we were reusing the `data` param it would always hang indefinitely upon retries due to reusing the same file descriptor.

to fix, I moved all of the file processing into our manually-retried portion (and comment on why we're disabling octokit retries)

___

I validated this with a simpler reproduction locally:

```python
import flask

app = flask.Flask(__name__)

i = 0

@app.route('/<path:s>', methods=['POST'])
def upload(s):
    global i
    i += 1

    print('upload!', flush=True)

    print(f'content-length: {flask.request.headers["content-length"]}')
    print(f'len: {len(flask.request.stream.read())}', flush=True)
    if i % 2 == 0:
        return 'wat', 500
    else:
        return '', 204


def main() -> int:
    app.run(port=9001, threaded=False)
    return 0

if __name__ == '__main__':
    raise SystemExit(main())
```

```js
const fetch = require('node-fetch');
const fs = require('fs');

async function main() {
    const name = 'wat';

    const stats = fs.statSync(name);

    const file = fs.createReadStream(name);

    const url = 'http://localhost:9001/wat'
    const opts = {
      body: file,
      method: 'POST',
      headers: {
        'Content-Type': 'application/octet-stream',
        'Content-Length': stats.size,
      }
    };

    try {
        const res = await fetch(url, opts);
        if (res.status != 204) {
            throw 'wat';
        }
    } catch {
        console.log('error');
        await fetch(url, opts);
    }
}

main();
```

___

`node-fetch` has a mechanism for this (`fileFrom`) but it's not available until 3.x (which octokit is pinned to 2.x at the moment) so we may be able to revisit this when that upgrades


src/targets/github.ts
```diff
@@ -301,7 +325,7 @@ export class GitHubTarget extends BaseTarget {
         'Got an error when trying to upload an asset, deleting and retrying...'
       );
       await this.deleteAssetByName(params.release_id, params.name);
-      return this.handleGitHubUpload(params, --retries);
+      return this.handleGitHubUpload(release, path, contentType, --retries);
     }
```
BYK, line 329: We should close the file (this can probably be moved into `catch` as `autoClose` defaults to true so on success, the file handler should be freed.

See https://nodejs.org/api/stream.html#readabledestroyerror

```suggestion
      return this.handleGitHubUpload(release, path, contentType, --retries);
    } finally {
      file.destroy();
    }
```

BYK (changes requested):
Great find!

Added one important comment and then I think this is ready to fly.

src/targets/github.ts
```diff
@@ -277,19 +253,67 @@ export class GitHubTarget extends BaseTarget {
...
+      },
+    };
+
+    this.logger.trace('Upload parameters:', params);
+
     try {
-      return (
+      const ret = (
         await this.github.repos.uploadReleaseAsset({
-          ...params,
```
kamilogorek, line 286: `await this.github.repos.uploadReleaseAsset(params)`

src/targets/github.ts
```diff
@@ -277,19 +253,67 @@ export class GitHubTarget extends BaseTarget {
...
+        timeout: 10 * 1000,
+      },
+    };
+
+    this.logger.trace('Upload parameters:', params);
+
     try {
-      return (
+      const ret = (
         await this.github.repos.uploadReleaseAsset({
```
kamilogorek, line 291: `const ret = await this.github.repos.uploadReleaseAsset(params).data;`

kamilogorek (changes requested):

src/targets/github.ts
```diff
@@ -277,19 +253,67 @@ export class GitHubTarget extends BaseTarget {
...
+      },
+      release_id: release.id,
+      name,
+      // XXX: Octokit types this out as string, but in fact it also
+      // accepts a `Buffer` here. In fact passing a string is not what we
+      // want as we upload binary data.
+      data: file as any,
+      request: {
+        // we are handling retries -- octokit-retries will resuse our fd and
+        // hang forever
```
armenzg, line 281: Is there an upstream issue we can make a reference to?

armenzg:
Hi @BYK ! Good to see you!

kamilogorek:
Also, one thing to note is that we can actually bump it to `node-fetch@3` and use `fileFrom` directly, as `Octokit` allows for overriding fetch implementation

```js
import fetch from 'node-fetch';

new Octokit({
  request: {
    fetch
  }
});
```

src/targets/github.ts
asottile-sentry: the closest I can find is https://github.com/node-fetch/node-fetch/issues/368

PR #411, opened 2022-08-31
Add additional context to error when version has a 'v' prefix

mattgauntseo-sentry:
This PR also adds tests to the `checkVersionOrPart` function, this is largely for my own piece of mind that I haven't broken anything, let me know if these should be changed or removed (I acknowledge they aren't the most useful tests in the world).

### DIFF
--- a/src/commands/__tests__/prepare.test.ts
+++ b/src/commands/__tests__/prepare.test.ts
@@ -1,6 +1,6 @@
 import { join as pathJoin } from 'path';
 import { spawnProcess } from '../../utils/system';
-import { runPreReleaseCommand } from '../prepare';
+import { runPreReleaseCommand, checkVersionOrPart } from '../prepare';
 
 jest.mock('../../utils/system');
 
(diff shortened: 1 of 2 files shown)


src/commands/__tests__/prepare.test.ts
```diff
@@ -52,4 +52,36 @@ describe('runPreReleaseCommand', () => {
       }
     );
   });
+
+  test('return true for valid version', () => {
```
asottile-sentry, line 56: I would probably put these in a separate and/or nested describe block -- that way the test reads `checkVersionOrPart - returns true for valid version` rather than `runPreReleaseCommand - returns true for valid version`

src/commands/prepare.ts
```diff
@@ -112,14 +112,18 @@ const SLEEP_BEFORE_PUBLISH_SECONDS = 30;
...
+export function checkVersionOrPart(argv: Arguments<any>, _opt: any): boolean {
   const version = argv.newVersion;
   if (['major', 'minor', 'patch'].indexOf(version) > -1) {
     throw Error('Version part is not supported yet');
   } else if (isValidVersion(version)) {
     return true;
   } else {
-    throw Error(`Invalid version or version part specified: "${version}"`);
+    let errMsg = `Invalid version or version part specified: "${version}"`;
+    if (version.indexOf('v') === 0) {
```
asottile-sentry, line 123: perhaps `.startsWith('v')` ?

asottile-sentry (commented):
just a few small things !

overall looks good

src/commands/__tests__/prepare.test.ts
mattgauntseo-sentry: Good catch

src/commands/prepare.ts
```diff
@@ -277,7 +281,7 @@ function checkGitStatus(repoStatus: StatusResult, rev: string) {
   ) {
     reportError(
       'Your repository is in a dirty state. ' +
-        'Please stash or commit the pending changes.',
+      'Please stash or commit the pending changes.',
```
asottile-sentry, line 284: this looks like formatter noise? I would probably rebase this out and/or we should set up linters that make this consistent

asottile-sentry (approved):

bukzor-sentryio (commented):
Looks good to me.

PR #412, opened 2022-09-01
feat: maven kotlin multiplatform support

buenaflor:
This introduces some additions to be able to support a Kotlin Multiplatform publish. 

Additional artifacts include: `*.klib, *-all.jar, *-metadata.jar, *.module`.

Example artifacts of [Kermit](https://github.com/touchlab/Kermit):
 - ios artifacts: https://repo1.maven.org/maven2/co/touchlab/kermit-sentry-iosarm64/1.2.0-M2/
 - root artifacts: https://repo1.maven.org/maven2/co/touchlab/kermit-sentry/1.2.0-M2/

For context: for a KMP publish all platforms  need to be published and the root artifact contains metadata, see: https://github.com/getsentry/sentry-kotlin-multiplatform/pull/37#issuecomment-1219123937

The generated gpg command example:
root artifact:
```
"-Dfile=/var/folders/j2/r3dsjfkx4pj5n0v8jjlb6lv80000gn/T/craft-mSNdRu/sentry-kotlin-multiplatform-0.0.1/sentry-kotlin-multiplatform-0.0.1.jar" 
"-Dfiles=/var/folders/j2/r3dsjfkx4pj5n0v8jjlb6lv80000gn/T/craft-mSNdRu/sentry-kotlin-multiplatform-0.0.1/sentry-kotlin-multiplatform-0.0.1-javadoc.jar,/var/folders/j2/r3dsjfkx4pj5n0v8jjlb6lv80000gn/T/craft-mSNdRu/sentry-kotlin-multiplatform-0.0.1/sentry-kotlin-multiplatform-0.0.1-sources.jar,/var/folders/j2/r3dsjfkx4pj5n0v8jjlb6lv80000gn/T/craft-mSNdRu/sentry-kotlin-multiplatform-0.0.1/sentry-kotlin-multiplatform-0.0.1-all.jar,/var/folders/j2/r3dsjfkx4pj5n0v8jjlb6lv80000gn/T/craft-mSNdRu/sentry-kotlin-multiplatform-0.0.1/sentry-kotlin-multiplatform-0.0.1.module" 
"-Dclassifiers=javadoc,sources,all,module" 
"-Dtypes=jar,jar,jar,module" 
...
``` 

ios artifact:
```
"-Dfile=/var/folders/j2/r3dsjfkx4pj5n0v8jjlb6lv80000gn/T/craft-QEa8yL/sentry-kotlin-multiplatform-iosArm64-0.0.1/sentry-kotlin-multiplatform-iosArm64-0.0.1.klib" 
"-Dfiles=/var/folders/j2/r3dsjfkx4pj5n0v8jjlb6lv80000gn/T/craft-QEa8yL/sentry-kotlin-multiplatform-iosArm64-0.0.1/sentry-kotlin-multiplatform-iosArm64-0.0.1-javadoc.jar,/var/folders/j2/r3dsjfkx4pj5n0v8jjlb6lv80000gn/T/craft-QEa8yL/sentry-kotlin-multiplatform-iosArm64-0.0.1/sentry-kotlin-multiplatform-iosArm64-0.0.1-sources.jar/var/folders/j2/r3dsjfkx4pj5n0v8jjlb6lv80000gn/T/craft-QEa8yL/sentry-kotlin-multiplatform-iosArm64-0.0.1/sentry-kotlin-multiplatform-iosarm64-0.0.1-cinterop-Sentry.NSException.klib,/var/folders/j2/r3dsjfkx4pj5n0v8jjlb6lv80000gn/T/craft-QEa8yL/sentry-kotlin-multiplatform-iosArm64-0.0.1/sentry-kotlin-multiplatform-iosarm64-0.0.1-cinterop-Sentry.Scope.klib,/var/folders/j2/r3dsjfkx4pj5n0v8jjlb6lv80000gn/T/craft-QEa8yL/sentry-kotlin-multiplatform-iosArm64-0.0.1/sentry-kotlin-multiplatform-iosarm64-0.0.1-cinterop-Sentry.klib,/var/folders/j2/r3dsjfkx4pj5n0v8jjlb6lv80000gn/T/craft-QEa8yL/sentry-kotlin-multiplatform-iosArm64-0.0.1/sentry-kotlin-multiplatform-iosArm64-0.0.1-metadata.jar,/var/folders/j2/r3dsjfkx4pj5n0v8jjlb6lv80000gn/T/craft-QEa8yL/sentry-kotlin-multiplatform-iosArm64-0.0.1/sentry-kotlin-multiplatform-iosArm64-0.0.1.module" 
"-Dclassifiers=javadoc,sources,cinterop,cinterop,cinterop,metadata,module"
"-Dtypes=jar,jar,klib,klib,klib,jar,module"
...
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1016,14 +1016,16 @@ Note: in order to see the output of the commands, set the [logging level](#loggi
 
 **Configuration**
 
-| Option              | Description                                                          |
-| ------------------- | -------------------------------------------------------------------- |
-| `mavenCliPath`      | Path to the Maven CLI. It must be executable by the calling process. |
-| `mavenSettingsPath` | Path to the Maven `settings.xml` file.                               |
-| `mavenRepoId`       | ID of the Maven server in the `settings.xml`.                        |
-| `mavenRepoUrl`      | URL of the Maven repository.                                         |
-| `android`           | Android configuration, see below.                                    |
+| Option                | Description                                                          |
+| --------------------- | -------------------------------------------------------------------- |
+| `mavenCliPath`        | Path to the Maven CLI. It must be executable by the calling process. |
+| `mavenSettingsPath`   | Path to the Maven `settings.xml` file.                               |
+| `mavenRepoId`         | ID of the Maven server in the `settings.xml`.                        |
+| `mavenRepoUrl`        | URL of the Maven repository.                                         |
+| `android`             | Android configuration, see below.                                    |
+| `kotlinMultiplatform` | Kotlin Multiplatform configuration, see below.                       |
 
+The Kotlin Multiplatform configuration is optional and `false` by default.
 If your project isn't related to Android, you don't need this configuration and
 can set the option to `false`. If not, set the following nested elements:
 
@@ -1056,6 +1058,23 @@ targets:
       distDirRegex: /^sentry-android-.*$/
       fileReplaceeRegex: /\d\.\d\.\d(-SNAPSHOT)?/
       fileReplacerStr: release.aar
+```
+
+**Example (with Kotlin Multiplatform config)**
+```yaml
+targets:
+  - name: maven
+    mavenCliPath: scripts/mvnw.cmd
+    mavenSettingsPath: scripts/settings.xml
+    mavenRepoId: ossrh
+    mavenRepoUrl: https://oss.sonatype.org/service/local/staging/deploy/maven2/
+    android:
+      distDirRegex: /^sentry-android-.*$/
+      fileReplaceeRegex: /\d\.\d\.\d(-SNAPSHOT)?/
+      fileReplacerStr: release.aar
+    kotlinMultiplatform:
+      rootDistDirRegex: /sentry-kotlin-multiplatform-[0-9]+.*$/
+      appleDistDirRegex: /sentry-kotlin-multiplatform-(macos|ios|tvos|watchos).*/
 ```
 
 ### Symbol Collector (`symbol-collector`)
--- a/README.md
+++ b/README.md
@@ -1052,7 +1052,7 @@ Note: in order to see the output of the commands, set the [logging level](#loggi
 | `mavenRepoId`         | ID of the Maven server in the `settings.xml`.                        |
 | `mavenRepoUrl`        | URL of the Maven repository.                                         |
 | `android`             | Android configuration, see below.                                    |
-| `kotlinMultiplatform` | Kotlin Multiplatform configuration, see below.                       |
+| `kmp` | Kotlin Multiplatform configuration, see below.                       |
 
 The Kotlin Multiplatform configuration is optional and `false` by default.
 If your project isn't related to Android, you don't need this configuration and
@@ -1101,7 +1101,7 @@ targets:
       distDirRegex: /^sentry-android-.*$/
       fileReplaceeRegex: /\d\.\d\.\d(-SNAPSHOT)?/
       fileReplacerStr: release.aar
-    kotlinMultiplatform:
+    kmp:
       rootDistDirRegex: /sentry-kotlin-multiplatform-[0-9]+.*$/
       appleDistDirRegex: /sentry-kotlin-multiplatform-(macos|ios|tvos|watchos).*/
 ```
--- a/src/targets/maven.ts
+++ b/src/targets/maven.ts
@@ -153,17 +153,10 @@ export class MavenTarget extends BaseTarget {
   }
 
   private getKotlinMultiplatformSettings(): KotlinMultiplatformFields {
-    if (this.config.kotlinMultiplatform === false) {
+    if (this.config.kotlinMultiplatform === false || !this.config.kotlinMultiplatform) {
       return {
         kotlinMultiplatform: false,
       };
-    }
-
-    if (!this.config.kotlinMultiplatform) {
-      throw new ConfigurationError(
-        'Required Kotlin Multiplatform configuration was not found in the configuration file. ' +
-          'See the documentation for more details'
-      );
     }
 
     if (
@@ -181,7 +174,6 @@ export class MavenTarget extends BaseTarget {
         'Required apple configuration for Kotlin Multiplatform is incorrect. See the documentation for more details.'
       );
     }
-
 
     return {
       kotlinMultiplatform: {
--- a/src/targets/maven.ts
+++ b/src/targets/maven.ts
@@ -14,6 +14,7 @@ import { ConfigurationError } from '../utils/errors';
 import { stringToRegexp } from '../utils/filters';
 import { checkEnvForPrerequisite } from '../utils/env';
 import { importGPGKey } from '../utils/gpg';
+import { readdirSync } from 'fs'
 
 export const POM_DEFAULT_FILENAME = 'pom-default.xml';
 const POM_FILE_EXT = '.xml'; // Must include the leading `.`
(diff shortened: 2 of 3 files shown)


src/targets/maven.ts
```diff
@@ -393,7 +448,10 @@ export class MavenTarget extends BaseTarget {
         );
       }
     }
-
+    // If klib target file name exists, use that instead of .jar
+    if (existsSync(join(distDir, `${moduleName.toLowerCase()}.klib`))) {
+      return `${moduleName}.klib`;
+    }
```
iker-barriocanal, line 454: If the file isn't accessible (e.g. it doesn't exist), the `.jar` filename will be returned. I don't know if the `.jar` file exists for kotlin, or what the consequences are of uploading it instead of the `.klib`. If a required file for a release doesn't exist, we should fail the release.

Another thing to consider is how easy it'd be to debug a failing release. A message like "hey, I can't find xxx file" is more helpful than "hey, I'm uploading the files and can't read yyy file, which isn't related to the release at all".

An alternative approach is to update the target's config to keep track of what we are releasing (android? kotlin?). If we know that, we could return directly the `.klib` filename, and let the upload fail if it can't read the file (which automatically produces a more helpful failure message). This is what currently happens with the `.jar` file (no disk operations happen in this method).

src/targets/maven.ts
```diff
@@ -355,14 +391,33 @@ export class MavenTarget extends BaseTarget {
...
       pomFile: join(distDir, 'pom-default.xml'),
     };
+    if (existsSync(join(distDir, `${moduleName.toLowerCase()}-all.jar`))) {
+      Object.assign(files, { allFile: join(distDir, `${moduleName}-all.jar`)})
+    }
+    if (existsSync(join(distDir, `${moduleName.toLowerCase()}.module`))) {
+      Object.assign(files, { moduleFile: join(distDir, `${moduleName}.module`) })
+    }
+    if (existsSync(join(distDir, `${moduleName.toLowerCase()}-metadata.jar`))) {
+      Object.assign(files, { metadataFile: join(distDir, `${moduleName}-metadata.jar`) })
```
iker-barriocanal, line 409: Instead of using `Object.assign`, can we use the following more-readable syntax?

```js
files['allFile'] = join(...);
```

src/targets/maven.ts
```diff
@@ -355,14 +391,33 @@ export class MavenTarget extends BaseTarget {
...
+    }
+    if (existsSync(join(distDir, `${moduleName.toLowerCase()}.klib`))) {
+      const cinteropFiles: string[] = []
+      readdirSync(distDir).forEach(file => {
```
iker-barriocanal, line 413: This code is potentially running for every artifact. Can we read the directory async? You can do this with `fsPromises.readdir`, and see an [example on line 281](https://github.com/getsentry/craft/pull/412/files#diff-d138225e2b0f12b719ddf3deee29e05257b82ef40f9adf1c7aa880696e7237f1R281).

iker-barriocanal, line 413: I suggest not to check for the existence of individual files if we are going to read the whole directory just after that. You could do something like the following, using [the `includes` function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes):
```
files = await readFilesAsync()
if (files.includes(fileXXX)) {
  // do sth
}
```

iker-barriocanal: You also may want to consider what the output should look like if someone does a kotlin release and a file is missing.

src/targets/maven.ts
```diff
@@ -328,17 +329,52 @@ export class MavenTarget extends BaseTarget {
...
       pomFile,
     } = this.getFilesForMavenPomDist(distDir);
 
+    // Default values
+    let sideArtifacts = `${javadocFile},${sourcesFile}`
+    let classifiers = 'javadoc,sources';
+    let types = 'jar,jar';
+
+    if (klibFiles) {
+      sideArtifacts += klibFiles
```
iker-barriocanal, line 345: Is the format correct here? These variables have a specific format, and all modifications below consider it.

iker-barriocanal (commented):
The craft codebase uses semicolons, can we follow the same style?

src/targets/maven.ts
buenaflor: These format should be fine as I manually tested them and also have some sample output in the pr summary, but I'll also write tests to ensure correctness

src/targets/maven.ts
buenaflor: I guess what I can do is create a target config for `kotlin-multiplatform` or something like that to keep the `uploadPomDistribution` separated and only add those files if it's a `kotlin-multiplatform` artifact. that way it doesn't interfere with the way it has been before I added the changes

iker-barriocanal (commented):
Looks much better!

Also -- don't forget to add documentation for this new config in the `README`.

README.md
```diff
@@ -1058,6 +1060,23 @@ targets:
...
+    mavenSettingsPath: scripts/settings.xml
+    mavenRepoId: ossrh
+    mavenRepoUrl: https://oss.sonatype.org/service/local/staging/deploy/maven2/
+    android:
+      distDirRegex: /^sentry-android-.*$/
+      fileReplaceeRegex: /\d\.\d\.\d(-SNAPSHOT)?/
+      fileReplacerStr: release.aar
+    kotlinMultiplatform:
+      rootDistDirRegex: /sentry-kotlin-multiplatform-[0-9]+.*$/
+      appleDistDirRegex: /sentry-kotlin-multiplatform-(macos|ios|tvos|watchos).*/
```
iker-barriocanal, line 1077: Do we need the kotlin config _and_ android config? Shouldn't the android config be `false` if we are making a kotlin release?

buenaflor: We need to upload multiple artifacts including Android (it's a target platform that we want to support within Kotlin Multiplatform)

This is everything that needs to be uploaded: Screenshot 2022-09-06 at 02 07 37

marandaneto:
@buenaflor did you try to call `gpg:sign-and-deploy-file` manually with all those params and check the temporary `~/.m2/repository/io/sentry` folder if it's correct?

iker-barriocanal (commented):
Tests currently only cover config loading. Put simply, what this PR introduces is spawning a process with a list of arguments executed under a certain condition (when we make a kotlin multiplatform release, run `gpg blahblahblah`). Can we verify that the process is spawned, when it should and with the expected arguments?

romtsn:
@iker-barriocanal what do you think is still missing for this PR? just tests from your latest comment or some other comments need to be addressed?

iker-barriocanal:
@romtsn I think only tests are missing (verifying the temporary folder as Manoel suggests may be one of them). That said, I'm not blocking the PR. I don't know if @asottile-sentry has other thoughts too.

marandaneto:
Making it a draft since it's not fully tested yet.

README.md
```diff
@@ -1045,14 +1045,16 @@ Note: in order to see the output of the commands, set the [logging level](#loggi
...
-| `android`           | Android configuration, see below.                                    |
-
+| Option                | Description                                                          |
+| --------------------- | -------------------------------------------------------------------- |
+| `mavenCliPath`        | Path to the Maven CLI. It must be executable by the calling process. |
+| `mavenSettingsPath`   | Path to the Maven `settings.xml` file.                               |
+| `mavenRepoId`         | ID of the Maven server in the `settings.xml`.                        |
+| `mavenRepoUrl`        | URL of the Maven repository.                                         |
+| `android`             | Android configuration, see below.                                    |
+| `kotlinMultiplatform` | Kotlin Multiplatform configuration, see below.                       |
```
marandaneto, line 1055:
```suggestion
| `kmp` | Kotlin Multiplatform configuration, see below.                       |
```

Maybe just shorter since there's a description?

marandaneto:
LGTM from my side but I'd let the owners of this repo approve it.

asottile-sentry (approved):

README.md
buenaflor: yup sounds good

PR #414, opened 2022-09-08
Adding delay between retries for getting artifacts.

mattgauntseo-sentry:
This change will add a delay between retries to look for artifacts.

This should help to address the issues raised in https://github.om/getsentry/publish/issues/1209 which appears to be caused by a race condition between artifacts being uploaded and GitHub's API returning those artifacts.

NOTE: The additional tests ensure the logic for handing pagination and revision date work as expected.

### DIFF
--- a/src/artifact_providers/github.ts
+++ b/src/artifact_providers/github.ts
@@ -16,8 +16,11 @@ import {
   withTempDir,
 } from '../utils/files';
 import { extractZipArchive } from '../utils/system';
+import { sleep } from '../utils/async';
 
 const MAX_TRIES = 3;
+const MILLISECONDS = 1000;
+const ARTIFACTS_POLLING_INTERVAL = 10 * MILLISECONDS;
 
 export type ArtifactItem = RestEndpointMethodTypes['actions']['listArtifactsForRepo']['response']['data']['artifacts'][0];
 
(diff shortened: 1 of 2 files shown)


src/artifact_providers/github.ts
```diff
@@ -49,48 +52,52 @@ export class GitHubArtifactProvider extends BaseArtifactProvider {
...
+      if (total_count <= per_page * (page + 1)) {
+        this.logger.debug(`No more pages remaining`);
+        break;
+      }
 
-    let checkNextPage = false;
-    if (artifactResponse.total_count > per_page * (page + 1)) {
       if (revisionDate === undefined) {
         revisionDate = (
           await this.github.git.getCommit({
```
mattgauntseo-sentry, line 103: **One subtle change with this refactor**

Previously the revision date would be passed to the recursive call to `getRevisionArtifact`. With this change, the `revisionDate` value is retrieved once per call to `searchForRevisionArtifact`.

IF we want to reduce the `getCommit` API calls we could do the following:

1. Getting the revision date once before calling `searchForRevisionArtifact` and pass it in as a value. The down side of this is I'd expect the majority of artifacts to be found in the first page of results.
1. Return the `revisionDate` from `searchForRevisionArtifact` and reuse it. The down side here is we have to handle multiple return values which just muddles the code a little.

Curious if anyone has strong opinions about this. I'm fine with how it is right now, it keeps the retry logic consistent, otherwise I'd probably opt for the first option, get the revision date and pass it in to each search call.

asottile-sentry: another idea might be:

```js
class LazyDate {
    #date = null;
    constructor(github) {
        this.github = github
    }
    get date() {
        return this.#date || (this.#date = await this.github.whatever(...));
    }
}
```

(idk what I'm doing, I've never written a class in js)

mattgauntseo-sentry: Very good point, so we could do:

```
export class GitHubArtifactProvider extends BaseArtifactProvider {
  public readonly github: Octokit;
  private revisionDates: Map<string, string>;

  public constructor(config: ArtifactProviderConfig) {
    super(config);
    this.github = getGitHubClient();
    this.revisionDates = new Map<string, string>();
  }
  
  private async getRevisionDate(revision: string): Promise<string> {
    const { repoName: repo, repoOwner: owner } = this.config;
    return this.revisionDates.get(revision) || (
      await this.github.git.getCommit({
        owner,
        repo,
        commit_sha: revision,
      })
    ).data.committer.date;
  }
```

mattgauntseo-sentry: This was a great idea. 

I didn't go with a class because we should be able to rely on Javascript scopes for this. I used generics so can keep type-safety and it removes the need for `LazyDate`, `LazyCommit`, `Lazy...` if we ever want this for something else.

asottile-sentry (approved):

src/artifact_providers/github.ts
```diff
@@ -110,25 +109,51 @@ export class GitHubArtifactProvider extends BaseArtifactProvider {
         lastArtifact.created_at >= revisionDate;
     }
 
-    if (checkNextPage) {
-      page += 1;
-    } else {
-      // If we are retrying, reset page index and start over
-      page = 0;
-      tries += 1;
-    }
```
BYK, line 119: If you are removing this logic, then you should probably remove the `checkNextPage` logic above too, right? Or am I missing something?

src/artifact_providers/github.ts
```diff
@@ -216,3 +235,17 @@ export class GitHubArtifactProvider extends BaseArtifactProvider {
...
+
+export function lazyRequest<T>(cb: lazyRequestCallback<T>): lazyRequestCallback<T> {
+  let data: T;
+  return async () => {
+    if (!data) {
+      data = await cb();
+    }
+    return data;
+  }
+}
```
BYK, line 247: This looks a lot like Promise caching?

mattgauntseo-sentry: Yeah that's what it's doing, let me know if there is a cleaner approach and / or naming suggestions.

src/artifact_providers/github.ts
BYK: Ah, sorry I glanced over the for loop completely missing the use of `checkNextPage` there :)

src/artifact_providers/github.ts
BYK: You can `await` a promise multiple times so I don't really understand the use case here.

You can do something like:
```js
const commitPromise = 
const getRevisionDate = async () => {
  return (await commitPromise).data.committer.date;
}
```

I also don't see the value of delaying the request until we are in the for loop as the for loop _will_ happen at least once, so why not just fetch the revision date once, before the loop and use that after on, eliminating another async step in the for loop (which should be _slightly_ more efficient):

```js
const revisionDate = await this.github.git.getCommit({
  owner,
  repo,
  commit_sha: revision,
}).data.committer.date;
```

src/artifact_providers/github.ts
```diff
@@ -110,25 +109,51 @@ export class GitHubArtifactProvider extends BaseArtifactProvider {
...
+
+      artifact = await this.searchForRevisionArtifact(revision, getRevisionDate);
+      if (artifact) {
+        return artifact;
+      }
+
+      // There may be a race condition between artifacts being uploaded
+      // and the GitHub API having the info to return.
+      // Wait before retries to give GitHub a chance to propagate changes.
+      if (tries + 1 < MAX_TRIES) {
```
BYK, line 148: Also, now that I'm thinking more clearly, I don't think there's any point going through all the pages in retries:

If we assume the response is order by descending date and we couldn't find the artifact in our first pagination, it is impossible for it to show up on any page except for the first page, right?

asottile-sentry: in a repository with lots of activity and lots of artifacts it may appear paginated (since this api is the entire repository)

src/artifact_providers/github.ts
asottile-sentry: the request for the commit date is delayed because most of the time it isn't needed (the first page will _always_ be requested)

src/artifact_providers/github.ts
BYK: For that to happen you need 100+ artifacts showing up in a matter of seconds. Don't think that's very realistic but yeah, valid point.

src/artifact_providers/github.ts
BYK: Correct. That said even in that case I'm not sure if this generalized `lazyRequest` interface that is only being used in a single place is still needed.

Since `searchForRevisionArtifact` is already passed `revision` and accesses the API for other things, it can encapsulate the delayed request and then caching its response in itself, making both testing and following code easier.

asottile-sentry: sure sure -- was mostly addressing this bit of your comment:

> why not just fetch the revision date once, before the loop and use that after on

mattgauntseo-sentry: `searchForRevisionArtifact` may be called multiple times `getRevisionArtifact`.

The options for managing getting the commit is:

1. Get the commit in each `searchForRevision` call (can result in three calls vs one) and is what I had in the first commit for this PR.
2. Get the commit once in `getRevisionArtifact` and pass it to `searchForRevision` which may result in an unnecessary API call where the artifact is on the first page and the revision commit is never queried. (This is a small issue but was the original behavior)
3. The lazy request approach where the API call is made when needed and data is cached.

With respect to this vs the promise caching, I think that'll run into option 2 of making a request when we may no need it.

BYK: Yeah, I see the reasoning. Was just arguing that the lazy fetch approach is adding some complexity into the code that is also not used anywhere to just save a single API request, that's all.

asottile-sentry: that complexity was already existing -- it's now just preserved over retries

PR #432, opened 2022-11-30
feat: Support initial manifest.json files for new SDK releases

kamilogorek:
With this feature, we are able to update https://github.com/getsentry/sentry-release-registry readme and allow people to add `manifest.json` file with `0.0.0` version instead of a specific "existing" release.
It will perform a correct release flow for the initial release and clean up the `manifest.json` file afterwards.


kamilogorek:
Actually lets do this so we don't need to perform any actions on the registry side.

PR #433, opened 2022-11-30
feat: Allow for initial registry release with manifest templates

kamilogorek:
I'm not fully convinced it's a good idea, but wanted to get the code out for discussion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -814,14 +814,16 @@ _none_
 
 **Configuration**
 
-| Option                         | Description                                                                                                                                                                                                                                |
-| ------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
-| `apps`                         | List of `app` configs as a dict, keyed by their canonical names (example: `app:craft`)                                                                                                                                                     |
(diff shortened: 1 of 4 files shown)


tonyo (commented):
(has been a while since I last looked at Craft things, so apologize if things were already discussed/agreed on elsewhere)

Not saying we shouldn't do this (we can totally start with this, assuming everyone is on the same page about this new behaviour), just a few thoughts for consideration:

* I feel like failing the `publish` step when something is wrong with the registry manifest is a bit too late: it still has a high potential to leave the release in half-published state (some targets succeeded, but registry and some others didn't). So ideally, developers should catch this on the `prepare` stage somehow. 
  * Maybe we could introduce optional "pre-flight checks" for all targets, that are executed at the `prepare` stage? There we could do all sorts of validations and let devs fix the things before it's too late.
* The main goal of not automating this at the very beginning of Craft was to avoid accidentally pushing some non-validated garbage to the release registry (first releases are sometimes messy), but it doesn't sound like a good enough reason anymore, so I generally agree that we should do this.
  * Continuing the pre-flight check idea, maybe

PR #435, opened 2022-12-06
Bump symbol collector

bruno-garcia:
The [release 1.8.0](https://github.com/getsentry/symbol-collector/releases/tag/1.8.0) contains some fixes
Including to this issue that's still coming from `craft` runs: 

https://sentry.io/organizations/sentry/issues/3061145532/?project=5953213

Does CI smoke test the commands? I tested the Android and macOS builds of 1.8.0 but not on Linux.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,7 +41,7 @@ RUN apt-get -qq update \
   && cargo install cargo-hack \
   && gem install cocoapods \
   # Install https://github.com/getsentry/symbol-collector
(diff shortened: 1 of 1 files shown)


asottile-sentry:
here's my usual procedure for debugging a failed docker build: https://youtu.be/hd1AKYGiWNk

bruno-garcia:
> I think it's failing to build, exit code 137 is usually SIGKILL -- often OOM or otherwise (it could also be a build timeout?)

Original (compressed) was 28 MB, new one is 29 MB. Bad luck if that 1 MB pushed it over a limit. Let me do a cheap test here and try re-run things with close/reopen

asottile-sentry (approved):

PR #437, opened 2022-12-16
Emit dotnet and nuget version info

mattjohnsonpint:
Adds `dotnet --version` and `dotnet nuget --version` info to the publish output, and (hopefully) works around the bug that is failing this workflow:

https://github.com/getsentry/publish/actions/runs/3716669257/jobs/6303253135#step:10:176


asottile-sentry (approved):
lol, sad but also still valuable either way!

PR #444, opened 2023-01-05
document the sentry-pypi target

asottile-sentry:


README.md
```diff
@@ -603,6 +604,34 @@ targets:
...
+| Option             | Description                                                        |
+| ------------------ | ------------------------------------ |
+| `internalPypiRepo` | GitHub repo containing pypi metadata |
+
+**Example**
+
+```yaml
+targets:
+  - name: pypi
+  - name: sentry-pypi
```
mattgauntseo-sentry, line 631: NIT: Would it make sense to change the name of this to `internal-pypi` or `alternate-pypi` instead of `sentry-pypi` if it's meant for general use outside of sentry?

mattgauntseo-sentry (approved):

README.md
BYK: But it looks like a flow that can be generally used so I kind of agree with @mattgauntseo-sentry here.

PR #451, opened 2023-02-01
Skip already published pub.dev package for the same package and version

marandaneto:
Context https://github.com/getsentry/publish/actions/runs/4066915710/jobs/7003618966

> STDERR:dart: Version 7.0.0-beta.2 of package sentry already exists.

If the very same package and version already has been published, skip it and go to the next one.

### DIFF
--- a/src/targets/__tests__/pubDev.test.ts
+++ b/src/targets/__tests__/pubDev.test.ts
@@ -128,6 +128,7 @@ describe('PubDev target configuration', () => {
 describe('publish', () => {
   test('single package', async () => {
     const revision = 'r3v1s10n';
+    const version = '1.0.0';
     const callOrder: string[] = [];
     const target = createPubDevTarget();
     target.createCredentialsFile = jest.fn(
@@ -140,7 +141,7 @@ describe('publish', () => {
       async () => void callOrder.push('publishPackage')
     );
 
-    await target.publish('1.0.0', revision);
+    await target.publish(version, revision);
 
     expect(target.createCredentialsFile).toHaveBeenCalled();
     expect(target.cloneRepository).toHaveBeenCalledWith(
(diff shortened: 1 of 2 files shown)


BYK:
The old behavior was intentional to catch versioning issues. That means this PR would introduce a breaking change, hence the new behavior should be triggered with a new flag.

marandaneto:
> The old behavior was intentional to catch versioning issues. That means this PR would introduce a breaking change, hence the new behavior should be triggered with a new flag.

Agreed to be a breaking change, 3 observations.

1. If we are under [semver](https://semver.org/#spec-item-4) and 0.x, that's not super important, does it matter?
2. Is there anyone besides us using this? I'm not aware of that.
3. [#392](https://github.com/getsentry/craft/pull/392) introduced similar behavior with a patch version, with no flags, but I see you also commented there, I don't think there was a follow-up.

marandaneto:
> > If we are under [semver](https://semver.org/#spec-item-4) and 0.x, that's not super important, does it matter?
> 
> I think we just have been lazy as this has been being used for prod-level for multiple years now. It might be better to make this explicit and release a 1.0 version and be more mindful of these.
> 
> > Is there anyone besides us using this? I'm not aware of that.
> 
> If _this_ refers to this specific target, I don't know. If _this_ refers to `craft`, the answer is yes. At least [formsort/oss](https://github.com/formsort/oss) (which I implemented myself) and possibly more.
> 
> > #392 introduced similar behavior with a patch version, with no flags, but I see you also commented there, I don't think there was a follow-up.
> 
> I think if this is the way to move forward, we may want to make it consistent across all targets. I don't have the formal veto power for changes that can be breaking nor I think blocking them for unrealized potential harms at the expense of potentially urgent business needs is a good idea. I just highlight "the more compatible" way forward based on my opinion. It is up to you to take the advice or not as hopefully I won't be the one bearing the consequences and if I do, I can always fork the project at that point (or make a PR and try my luck first). 😅

And I appreciate the advice, very valuable, Agree with everything.
The changes are pub.dev specifically (so it's less of a problem I guess).
#392 Also made the changes specifically to Crates.
Let me hear @kamilogorek and @asottile-sentry opinions here as well, I'm blocked by this right now, have a broken release, and would like to solve it asap, but I do agree that we should consider your advice, by either this is a must and I should address it now with a flag, or it's ok to open a new issue/PR eventually.

Issue #452, opened 2023-02-03
Flutter users should run `flutter pub get` instead of `dart pub get`.


marandaneto:
### Environment

I'm not sure what's configured in the image of the [publish](https://github.com/getsentry/publish) action.

### Steps to Reproduce

Try to publish the pub.dev target with a Flutter package, [example](https://github.com/getsentry/publish/issues/1821).
This is blocking us to release new versions of our SDKs.

### Expected Result

Publish with no errors

### Actual Result

https://github.com/getsentry/publish/actions/runs/4084948587/jobs/7042272509

> [info] dart: Resolving dependencies...
> [info] dart: 
> Error:  Process "dart" errored with code 1
> STDOUT: dart: Resolving dependencies...
> dart: 
> STDERR:dart: Warning: pubspec.yaml has overrides from pubspec_overrides.yaml
> dart: Because sentry_flutter requires the Flutter SDK, version solving failed.
> dart: 
> dart: Flutter users should run `flutter pub get` instead of `dart pub get`.
> dart: 
>   
>   STDOUT: dart: Resolving dependencies...
>   dart:
>   
>   
>   STDERR:dart: Warning: pubspec.yaml has overrides from pubspec_overrides.yaml
>   dart: Because sentry_flutter requires the Flutter SDK, version solving failed.
>   dart:
>   dart: Flutter users should run `flutter pub get` instead of `dart pub get`.
>   dart:

I'm not sure what has changed, maybe we need to upgrade the `dart` version in the image?
Maybe we need to access the raw logs and check if there's more info?
Maybe there's a breaking change and we cannot use `dart pub publish` anymore for Flutter packages?

I could not reproduce this locally, I actually got a different error, by trying `dart pub publish --dry-run` but it's different than the output in the error.

> Package validation found the following potential issue:
> * Your dependency on "sentry" should allow more than one version. For example:
> 
>   dependencies:
>     sentry: ^6.20.0
> 
>   Constraints that are too tight will make it difficult for people to use your package
>   along with other packages that also depend on "sentry".

BYK:
There is no "publish image". It uses the docker image built here: https://github.com/getsentry/craft/blob/master/Dockerfile

BYK:
It is defined here: https://github.com/getsentry/craft/blob/0b738d51ca7504a883bd12b40b9ac6cc12434715/src/targets/pubDev.ts#L67

BYK:
[...]

PR #453, opened 2023-02-03
Make Flutter available in the docker image

marandaneto:
Closes https://github.com/getsentry/craft/issues/452

Install Flutter in the docker image.
Removing Dart because Flutter ships Dart already and they should be compatible.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ ENV DEBIAN_FRONTEND=noninteractive \
   COCOAPODS_ALLOW_ROOT=1 \
   CARGO_HOME=/root/.cargo \
   RUSTUP_HOME=/root/.rustup \
-  PATH=${PATH}:/root/.cargo/bin:/usr/lib/dart/bin
+  PATH=${PATH}:/root/.cargo/bin:/tmp/flutter/bin
 
 RUN apt-get -qq update \
   && apt-get install -y --no-install-recommends \
@@ -28,13 +28,13 @@ RUN apt-get -qq update \
   && rm /tmp/packages-microsoft-prod.deb \
   && curl -fsSL https://download.docker.com/linux/debian/gpg | apt-key add - \
   && echo 'deb [arch=amd64] https://download.docker.com/linux/debian buster stable' >> /etc/apt/sources.list \
-  && curl -fsSL https://dl-ssl.google.com/linux/linux_signing_key.pub | gpg --dearmor -o /usr/share/keyrings/dart.gpg \
-  && echo 'deb [signed-by=/usr/share/keyrings/dart.gpg arch=amd64] https://storage.googleapis.com/download.dartlang.org/linux/debian stable main' | tee /etc/apt/sources.list.d/dart_stable.list \
+  && curl -fsSL https://storage.googleapis.com/flutter_infra_release/releases/stable/linux/flutter_linux_3.7.1-stable.tar.xz -o /tmp/flutter_linux_3.7.1-stable.tar.xz \
+  && tar xf /tmp/flutter_linux_3.7.1-stable.tar.xz
+  &&
   && apt-get update -qq \
   && apt-get install -y --no-install-recommends \
     dotnet-sdk-7.0 \
     docker-ce-cli \
-    dart \
   && apt-get clean \
   && rm -rf /var/lib/apt/lists/* \
   && curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s --  --profile minimal -y \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -28,13 +28,13 @@ RUN apt-get -qq update \
...
-  && curl -fsSL https://dl-ssl.google.com/linux/linux_signing_key.pub | gpg --dearmor -o /usr/share/keyrings/dart.gpg \
-  && echo 'deb [signed-by=/usr/share/keyrings/dart.gpg arch=amd64] https://storage.googleapis.com/download.dartlang.org/linux/debian stable main' | tee /etc/apt/sources.list.d/dart_stable.list \
+  && curl -fsSL https://storage.googleapis.com/flutter_infra_release/releases/stable/linux/flutter_linux_3.7.1-stable.tar.xz -o /tmp/flutter_linux_3.7.1-stable.tar.xz \
+  && tar xf /tmp/flutter_linux_3.7.1-stable.tar.xz
+  &&
   && apt-get update -qq \
   && apt-get install -y --no-install-recommends \
     dotnet-sdk-7.0 \
     docker-ce-cli \
-    dart \
```
marandaneto, line 37: I'm wondering if this should be indeed removed, no experience ith `--no-install-recommends`.

Dockerfile
```diff
@@ -28,13 +28,12 @@ RUN apt-get -qq update \
   && rm /tmp/packages-microsoft-prod.deb \
   && curl -fsSL https://download.docker.com/linux/debian/gpg | apt-key add - \
   && echo 'deb [arch=amd64] https://download.docker.com/linux/debian buster stable' >> /etc/apt/sources.list \
-  && curl -fsSL https://dl-ssl.google.com/linux/linux_signing_key.pub | gpg --dearmor -o /usr/share/keyrings/dart.gpg \
-  && echo 'deb [signed-by=/usr/share/keyrings/dart.gpg arch=amd64] https://storage.googleapis.com/download.dartlang.org/linux/debian stable main' | tee /etc/apt/sources.list.d/dart_stable.list \
+  && curl -fsSL https://storage.googleapis.com/flutter_infra_release/releases/stable/linux/flutter_linux_3.7.1-stable.tar.xz -o /tmp/flutter_linux_3.7.1-stable.tar.xz \
```
marandaneto, line 31: We cannot use `snapd`.
I've not found a `/latest` endpoint, so it's the [latest pinned version](https://docs.flutter.dev/get-started/install/linux) as of now.

Dockerfile
```diff
@@ -6,7 +6,7 @@ ENV DEBIAN_FRONTEND=noninteractive \
   COCOAPODS_ALLOW_ROOT=1 \
   CARGO_HOME=/root/.cargo \
   RUSTUP_HOME=/root/.rustup \
-  PATH=${PATH}:/root/.cargo/bin:/usr/lib/dart/bin
+  PATH=${PATH}:/root/.cargo/bin:/tmp/flutter/bin
```
marandaneto, line 9: The `bin` folder contains `flutter` and `dart` already.

marandaneto:
cc @BYK

Dockerfile
BYK: Nice

Dockerfile
BYK: You should not use a dynamic 'latest' version to make image building as idempotent as possible anyway 😉

BYK: Not sure if this applies to files in tmp as well but ideally you should remove this tar file after extracting.

Dockerfile
BYK: `--no-install-recommends` is to tell `apt` to not automatically install any recommended packages automatically to avoid bulk in the Docker image.

If you remove dart, it may cause failures for the old use cases of pub.dev target as it still assumes there to be a dart binary (this should be fine if the flutter package provides it now, I just don't know)

BYK (approved):
Looks good overall but you should definitely test this image locally to ensure you can run both `dart` and `flutter` commands.

Dockerfile
```diff
@@ -49,6 +46,10 @@ RUN apt-get -qq update \
   && rm /tmp/sym-collector.zip \
   && chmod +x /usr/local/bin/SymbolCollector.Console
 
+RUN curl -fsSL https://storage.googleapis.com/flutter_infra_release/releases/stable/linux/flutter_linux_3.7.1-stable.tar.xz -o /tmp/flutter_linux_3.7.1-stable.tar.xz \
+  && tar xf /tmp/flutter_linux_3.7.1-stable.tar.xz \
+  && rm /tmp/flutter_linux_3.7.1-stable.tar.xz
```
marandaneto, line 51: Using another layer due to OOM

BYK: That should not be necessary or a solution, at least toy knowledge. Is there a source saying another layer would avoid an OOM?

asottile-sentry: cloud build was OOMing while finalizing the layer

asottile-sentry (approved):

Dockerfile
marandaneto: Should be fine then since `dart` is still available, thanks.

Dockerfile
marandaneto: I also prefer pinned versions, this PR is only needed because the new dart version has breaking changes.

marandaneto:
Thank you @asottile-sentry and @BYK :)

Dockerfile
BYK: Would be great if this is added as a comment in the file as it is definitely not obvious.

PR #456, opened 2023-02-06
Skip duplicate nuget packages

mattjohnsonpint:
Since `Sentry.Android.AssemblyReader` will rarely need to be modified, we've elected to [version it independently](https://github.com/getsentry/sentry-dotnet/blob/089d7020c2a220afbb87a0357c5abaaf99f37b3e/src/Sentry.Android.AssemblyReader/Sentry.Android.AssemblyReader.csproj#L6-L7) from the other packages that typically get published with each release of sentry-dotnet.  

Craft will attempt to publish all nuget packages, and currently it will fail if any of them already exist.  We can pass a `--skip-duplicate` flag to `dotnet nuget push` so that we continue without failure.

This is probably a good idea regardless of the versioning issue, as presently if we get *any* error after the first package is published, we won't be able to succeed the publishing step on retry, without bumping the version.

### DIFF
--- a/src/targets/nuget.ts
+++ b/src/targets/nuget.ts
@@ -73,6 +73,7 @@ export class NugetTarget extends BaseTarget {
       '${NUGET_API_TOKEN}',
       '--source',
       this.nugetConfig.serverUrl,
+      '--skip-duplicate'
     ]);
   }
 


mattjohnsonpint:
Example failure:

https://github.com/getsentry/publish/actions/runs/4087468571/jobs/7048017788#step:10:269

```
Error:  Process "dotnet" errored with code 1

STDOUT: dotnet: Pushing Sentry.Android.AssemblyReader.1.0.0.nupkg to 'https://www.nuget.org/api/v2/package'...
dotnet:   PUT https://www.nuget.org/api/v2/package/
dotnet:   Conflict https://www.nuget.org/api/v2/package/ 1101ms
dotnet: To skip already published packages, use the option --skip-duplicate
dotnet: error: Response status code does not indicate success: 409 (A package with ID 'Sentry.Android.AssemblyReader' and version '1.0.0' already exists and cannot be modified.).
```

asottile-sentry:
a similar change was rejected here: #451

the reasoning given that it would hide versioning issues

mattjohnsonpint:
Ok.  We'll just use the same version number then.  Thanks.

PR #457, opened 2023-02-06
Fix extracting flutter file to tmp folder instead of current folder

marandaneto:
Fix https://github.com/getsentry/publish/issues/1828

> Error:  Error creating target instance for pub-dev!
> Error:  Executable "dart" not found. Is it installed?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,7 +47,7 @@ RUN apt-get -qq update \
   && chmod +x /usr/local/bin/SymbolCollector.Console
 
 RUN curl -fsSL https://storage.googleapis.com/flutter_infra_release/releases/stable/linux/flutter_linux_3.7.1-stable.tar.xz -o /tmp/flutter_linux_3.7.1-stable.tar.xz \
-  && tar xf /tmp/flutter_linux_3.7.1-stable.tar.xz \
+  && tar xf /tmp/flutter_linux_3.7.1-stable.tar.xz -C /tmp \
   && rm /tmp/flutter_linux_3.7.1-stable.tar.xz
 
 # craft does `git` things against mounted directories as root
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ ENV DEBIAN_FRONTEND=noninteractive \
   COCOAPODS_ALLOW_ROOT=1 \
   CARGO_HOME=/root/.cargo \
   RUSTUP_HOME=/root/.rustup \
-  PATH=${PATH}:/root/.cargo/bin:/tmp/flutter/bin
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -47,7 +47,7 @@ RUN apt-get -qq update \
   && chmod +x /usr/local/bin/SymbolCollector.Console
 
 RUN curl -fsSL https://storage.googleapis.com/flutter_infra_release/releases/stable/linux/flutter_linux_3.7.1-stable.tar.xz -o /tmp/flutter_linux_3.7.1-stable.tar.xz \
-  && tar xf /tmp/flutter_linux_3.7.1-stable.tar.xz \
+  && tar xf /tmp/flutter_linux_3.7.1-stable.tar.xz -C /tmp \
```
marandaneto, line 50: `PATH` is set to `/tmp/flutter/bin`, so this should fix.

chadwhitacre (approved):

Dockerfile
asottile-sentry, line 50: it would be more natural to put this in `/opt/flutter` instead of `/tmp`

asottile-sentry (changes requested):

Dockerfile
marandaneto: @asottile-sentry I'd rather do it in a separate PR and unblock myself asap, this is just missing for now, other things use `tmp` as well.
is this fine?

Dockerfile
```diff
@@ -46,9 +46,9 @@ RUN apt-get -qq update \
   && rm /tmp/sym-collector.zip \
   && chmod +x /usr/local/bin/SymbolCollector.Console
 
-RUN curl -fsSL https://storage.googleapis.com/flutter_infra_release/releases/stable/linux/flutter_linux_3.7.1-stable.tar.xz -o /tmp/flutter_linux_3.7.1-stable.tar.xz \
-  && tar xf /tmp/flutter_linux_3.7.1-stable.tar.xz \
-  && rm /tmp/flutter_linux_3.7.1-stable.tar.xz
+RUN curl -fsSL https://storage.googleapis.com/flutter_infra_release/releases/stable/linux/flutter_linux_3.7.1-stable.tar.xz -o /opt/flutter_linux_3.7.1-stable.tar.xz \
+  && tar xf /opt/flutter_linux_3.7.1-stable.tar.xz -C /opt \
+  && rm /opt/flutter_linux_3.7.1-stable.tar.xz
```
asottile-sentry, line 51: it doesn't matter but the tgz should sit in `/tmp` but the unpack should be in `/opt`

asottile-sentry (approved):

PR #458, opened 2023-02-08
Support nuget snupkg artifacts

mattjohnsonpint:
Adds support for publishing [Nuget Symbol Package files](https://learn.microsoft.com/nuget/create-packages/symbol-packages-snupkg) (`.snupkg` extension).

We'll need this to publish the .NET SDK repo, now that we start using them with https://github.com/getsentry/sentry-dotnet/pull/2166.

### DIFF
--- a/src/targets/nuget.ts
+++ b/src/targets/nuget.ts
@@ -15,7 +15,6 @@ export const DEFAULT_NUGET_SERVER_URL = 'https://api.nuget.org/v3/index.json';
 
 /** A regular expression used to find the package tarball */
 const DEFAULT_NUGET_REGEX = /^.*\d\.\d.*\.nupkg$/;
-const SYMBOLS_NUGET_REGEX = /^.*\d\.\d.*\.snupkg$/;
 
 /** Nuget target configuration options */
 export interface NugetTargetOptions {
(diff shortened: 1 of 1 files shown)


src/targets/nuget.ts
```diff
@@ -106,8 +110,19 @@ export class NugetTarget extends BaseTarget {
     await Promise.all(
       packageFiles.map(async (file: RemoteArtifact) => {
         const path = await this.artifactProvider.downloadArtifact(file);
+
+        // If an artifact containing a .snupkg file exists with the same base
+        // name as the .nupkg file, then download it to the same location.
+        // It will be picked up automatically when pushing the .nupkg.
+        const symbolFileName = file.filename.replace('.nupkg', '.snupkg');
```
bruno-garcia, line 117: would be nice to add a test that these two extensions work and some third e.g `*.zip` doesn't

mattjohnsonpint: Well, we don't have _any_ tests for nuget, (though [we do for some other targets](https://github.com/getsentry/craft/tree/master/src/targets/__tests__)) so we'd have to start there first I think.

As for picking up other files, that would be more of a test on `getArtifactsForRevision`.  All I'm doing here is matching `*.nupkg` with `*.snupkg` if it exists, and downloading it so it's in the same path before publishing.

`dotnet nuget push` automatically publishes a `.snupkg` if it is side-by-side the `.nupkg` in the same folder.

mattjohnsonpint: Actually, nevermind.  I overcomplicated this.  Replacing with a simpler approach.

mattjohnsonpint:
The first approach was overcomplicated.  This will work as well - it just publishes both `.nupkg` and `.snupkg` files via separate invocations of `dotnet nuget push`, which should work just fine.

PR #461, opened 2023-03-22
meta: fix typo in debug statement

billyvg:


billyvg:
Arg, was using the dumb github.dev editor and it autoformatted

PR #468, opened 2023-06-27
ref: update jest

asottile-sentry:
this should silence at least one of the security notices


tsconfig.build.json
```diff
@@ -12,7 +12,8 @@
     "noImplicitThis": false,
     "esModuleInterop": true,
     "importHelpers": true,
-    "noEmitHelpers": false
+    "noEmitHelpers": false,
+    "useUnknownInCatchVariables": false
```
asottile-sentry, line 16: there's a bunch of (admittedly correct) errors about not narrowing `error` in `catch` blocks -- going to leave those for another day

mattgauntseo-sentry (approved):

PR #475, opened 2023-07-18
Revert changes since 1.2.3

mydea:
It seems publishing with 1.3.0 is broken - see https://getsentry.atlassian.net/browse/INC-456. 

Reverting all changes since 1.2.3 for now, so we can re-release this as 1.4.0 and then figure this out with more time.


cleptric (approved):

asottile-sentry:
fwiw #474 didn't need reverting even on the old dockerfile -- but that's ok!

Issue #405, opened 2022-07-12
Craft fails the release-registry target if version already exists


bruno-garcia:
When we add a new package to release-registry, it's intuition for folks to add the version we're about to release.
The one that actually has that package.

Craft can't handle that, and crashes if the version already exists:

![image](https://user-images.githubusercontent.com/1633368/178547460-0844a693-3c48-41ba-9c86-33a7acca4e4f.png)

If the package isn't there, craft also fails. So we're required to add it manually first.

Request: Please ignore when the package already exists and carry on the release.

asottile-sentry:
hmmm -- if you create the release manually should you then check the registry box in the publish issue so it skips the (already done) one?

PR #476, opened 2023-07-18
feat: Handle existing NPM packages

mydea:
When publishing a JS monorepo, if something goes wrong in the publish process, it can happen that only _some_ packages are released, but others aren't. So you can e.g. have a state for the monorepo packages:

* package-1@1.0.2
* package-2@1.0.2 --> after this publish, something fails
* package-3@1.0.1

Now there is no way to fix the 1.0.2 release, because any rerun of this will fail when trying to publish package-1 because 1.0.2 already exists for it. See for example: https://github.com/getsentry/publish/actions/runs/5584823217/jobs/10206737755

This PR adds some special casing for this specific error, in which case we simply ignore and skip this. I don't have a great way of testing this directly, if there are any ideas let me know.

### DIFF
--- a/src/targets/npm.ts
+++ b/src/targets/npm.ts
@@ -209,6 +209,16 @@ export class NpmTarget extends BaseTarget {
       // Disable output buffering because NPM/Yarn can ask us for one-time passwords
       return spawnProcess(bin, args, spawnOptions, {
         showStdout: true,
+      }).catch(error => {
+        // When publishing a list of packages fails for only one package,
+        // you can never fix this by re-running publish because it will then _always_ fail due to existing versions.
+        // Instead, in this case we want to simply ignore the already existing package and continue.
+        if (error.message?.includes('You cannot publish over the previously published versions:')) {
+          console.warn(error);
+          return Promise.resolve();
+        }
+
+        return Promise.reject(error);
       });
     });
   }


BYK:
I really like the idea behind the PR however I think we should add more guardrails:

1. This behavior is only okay on a monorepo release with a retry so the behavior should be limited to that. We may consider adding a CLI flag "--retry" to indicate this is a retry. We should also check of this is part of multi-release (not necessarily a monorepo).
2. Instead of checking for an error string, it would be safer to check for a specific error code as the string can change without notice.

BYK:
Also related: https://github.com/getsentry/craft/issues/405

mydea:
Yeah, I'm also not too happy about the string matching, but the error I see is:

```
npm: npm ERR! code E403
  npm: npm ERR! 403 403 Forbidden - PUT https://registry.npmjs.org/@sentry-internal%2feslint-plugin-sdk - You cannot publish over the previously published versions: 7.59.1.
  npm: npm ERR! 403 In most cases, you or one of your dependencies are requesting
  npm: npm ERR! 403 a package version that is forbidden by your security policy.
```

So not too sure how to better detect this, because just checking 403 is probably also not a good idea 😅 

Any pointers on how to detect a multi-release? 🤔 but maybe it's good enough if we simply add a new options to be passed to opt-into this behavior?

`retry` is maybe not an ideal name for an option, as it may be ambiguous (could also mean retry to publish x times). What about `skipWhenPackageVersionExists` or something like this?

Lms24 (approved):
I believe we should get this in to avoid future situation where this blocks us or it unnecessarily prolongs an incident.

src/targets/npm.ts
```diff
@@ -209,6 +209,16 @@ export class NpmTarget extends BaseTarget {
       // Disable output buffering because NPM/Yarn can ask us for one-time passwords
       return spawnProcess(bin, args, spawnOptions, {
         showStdout: true,
+      }).catch(error => {
+        // When publishing a list of packages fails for only one package,
+        // you can never fix this by re-running publish because it will then _always_ fail due to existing versions.
```
asottile-sentry, line 214: an admin can check the box in publish that this target has completed -- this will skip the failing check

I'd prefer the existing approach rather than changing how craft behaves here

BYK (changes requested):
See the issue I linked please. This has been discussed earlier extensively and is a breaking change due to a fundamental behavior change. It should at least happen with an explicit flag.

src/targets/npm.ts
mydea: But if a single step publishes multiple packages, and one of them already exists, this will not work (it does not work right now) - you can then only skip `npm` overall.

BYK: It doesn't matter as that is the right (safest) approach to have by default. You may have out of sync packages published etc. and Craft cannot know that unless you explicitly tell it.

asottile-sentry: @mydea in such a case I'd recommend publishing an entirely new version and yanking the partially released one.  version numbers are ~free

PR #479, opened 2023-07-25
feat(elixir): Use asdf to install erlang and elixir

sl0thentr0py:


tonyo (approved):

asottile-sentry:
hi -- can we please wait a little bit so the repository owners can review things before making major changes like introducing a whole plugin ecosystem?  asdf is quite heavy handed for this and there's more direct solutions that'll be easier to maintain in the long run

sl0thentr0py:
I already tried the other route which caused the npm target to break. Bottom line is I want to unblock a contractor who's been waiting on this of upwards of 3 weeks now and I do not have time to go through all the dependencies / targets and see what breaks in that environment (nor is there any easy way of testing all the targets against the docker image). 
All of this requires better tooling but I believe neither the SDK team nor dev infra has time for all that, so this is the quickest way forward.

I can wait in the future for reviews, sure but fwiw craft feels like it has no clear owner for larger feature requests etc.

asottile-sentry:
craft _is_ owned by dev-infra.  this patch increases the docker image by 20% which is pretty significant for one language.

sl0thentr0py:
do you have a concrete viable alternative that unblocks me?

sl0thentr0py:
you need erlang/otp too which is too old on this debian

asottile-sentry:
the elixir page also links to downloads for erlang as well [here](https://www.erlang-solutions.com/downloads/)

PR #481, opened 2023-07-25
feat(elixir): Use precompiled packages from erlang solutions for elixir

sl0thentr0py:


cloudbuild.yaml
```diff
@@ -9,9 +9,10 @@ steps:
       - '-f'
       - 'builder.dockerfile'
   - name: 'us.gcr.io/$PROJECT_ID/craft:builder-$COMMIT_SHA'
-  - name: 'gcr.io/kaniko-project/executor:v1.5.1'
+  - name: 'gcr.io/kaniko-project/executor:v1.12.1'
     args:
       - '--cache=true'
+      - '--compressed-caching=false'
```
sl0thentr0py, line 15: had to add this otherwise build goes OOM with error code 137
see https://github.com/GoogleContainerTools/kaniko/issues/1669#issuecomment-1260772687

asottile-sentry: I think the problem is layer 2 is too big -- it'd probably be fine if that were split into several layers

Dockerfile
```diff
@@ -22,17 +22,21 @@ RUN apt-get -qq update \
     jq \
     unzip \
     openjdk-11-jdk \
-    maven \
-    elixir \
-  && curl -fsSL https://packages.microsoft.com/config/debian/10/packages-microsoft-prod.deb -o /tmp/packages-microsoft-prod.deb \
+    maven
```
asottile-sentry, line 25: nice!

this part needs to be at the end of any layer that does `apt` stuff:

```dockerfile
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/* \
```

asottile-sentry (approved):

PR #483, opened 2023-08-03
Do not update `{major}.json` if an older version was published

cleptric:
This should resolve an issue where craft wrongly updates the `{major}.json` if an older version is released. See https://github.com/getsentry/sentry-release-registry/commit/6db6af4aea76cc2b08940a25883c46e0645921d1

Update (@Lms24): This PR not only resolves the issue mentioned above but it rewrites the linking cases to take previously existing symlinks into account. This way, we can more safely bump the respective symlinks. Added tests to cover the most important + edge case paths.


src/utils/symlink.ts
```diff
@@ -42,24 +42,25 @@ export function createSymlinks(
...
     logger.warn(
-      `Not updating the latest version file: current version is "${oldVersion}", new version is "${newVersion}"`
+      `Not updating the latest version file: current version is "${oldVersion}", new version is "${newVersion}"\n`,
+      `Not updating the major version file: current version is "${oldVersion}", new version is "${newVersion}"`
     );
   } else {
     logger.debug(
-      `Changing symlink for "latest.json" from version "${oldVersion}" to "${newVersion}"`
+      `Changing symlink for "latest.json" from version "${oldVersion}" to "${newVersion}"\n`,
+      `Changing symlink for "{major}.json" from version "${oldVersion}" to "${newVersion}"`
```
BYK, line 57: Same as above. Also please move this new log line below, right before the actual command call.

tonyo (approved):

cleptric:
We currently do not have any tests in place. How would you envision these tests to look like?

asottile-sentry:
there's tests for other targets here -- can follow those as an example https://github.com/getsentry/craft/tree/master/src/targets/__tests__

cleptric:
Could you please provide a bit more specific guidance? How would you like these tests to be structured? Should we mock the symlinks in any way, etc.

asottile-sentry:
I don't particularly care beyond that there are tests for the functionality -- use your judgement to pick what makes sense for you

asottile-sentry:
I don't think this is quite right

consider this scenario:

- release 1.0
- release 2.0
- release 1.0.1

I expect the symlink structure to be:

```
latest -> 2.0
2 -> 2.0
1 -> 1.0.1
```

before this patch it would do that I believe but after this patch the 1 symlink will be untouched because it was "older"

I think the actual fix needs to understand the current set of symlinks and make the symlink based on that

tonyo:
@asottile-sentry yep, you're right 👍 
We had a chat with @Lms24 , we'll cover that case.

Lms24:
Talked with @tonyo (writing down for future reference): The missing check is if `new.major < old.major`, in which case we also need to update the old_major.json file.

asottile-sentry:
I'm not sure that's sufficient either -- the same problem will exist for minor revisions:

- 1.1.0
- 1.2.0
- 1.0.1

src/utils/symlink.ts
```diff
@@ -39,29 +45,80 @@ export function createSymlinks(
...
+    const majorVersionLink = `${parsedNewVersion.major}.json`;
+    logger.debug(
+      `${
+        newVersionMajorSymlinkedVersion ? 'Updating' : 'Adding'
+      } symlink for "${majorVersionLink}" ${
+        newVersionMajorSymlinkedVersion
+          ? `from version "${semVerToString(newVersionMajorSymlinkedVersion)}" `
+          : ''
+      }to "${newVersion}"`
+    );
```
asottile-sentry, line 90: there's way too much going on in this line -- if it's just for a debug line I'd just put the raw data in here instead of trying to make it end-user readable.  should eliminate all the logic here as well since this seems error prone / more likely to break than the actual logic

src/utils/symlink.ts
```diff
@@ -39,29 +45,80 @@ export function createSymlinks(
...
-  // link minor
-  const minorVersionLink = `${parsedNewVersion.major}.${parsedNewVersion.minor}.json`;
-  forceSymlink(baseVersionName, path.join(packageDir, minorVersionLink));
+function getExistingSymlinkedVersion(symlinkPath: string): SemVer | null {
+  try {
+    // using lstat instead of exists because broken symlinks return false for exists
+    fs.lstatSync(symlinkPath);
+    const linkedFile = fs.readlinkSync(symlinkPath);
+    return parseVersion(path.basename(linkedFile));
+  } catch {
```
asottile-sentry, line 120: this try catch is overly broad -- each of these 3 statements could raise errors but I think (?) you only intend to catch errors from the first one?  this could hide a programming error especially after future maintenance

src/utils/symlink.ts
```diff
@@ -39,29 +45,80 @@ export function createSymlinks(
...
+      }to "${newVersion}"`
     );
     forceSymlink(baseVersionName, path.join(packageDir, 'latest.json'));
   }
 
-  // link major
-  const majorVersionLink = `${parsedNewVersion.major}.json`;
-  forceSymlink(baseVersionName, path.join(packageDir, majorVersionLink));
+  // Read possibly existing symlinks for major and minor versions of the new version
+  const newVersionMajorSymlinkedVersion = getExistingSymlinkedVersion(
```
asottile-sentry, line 65: this variable name seems overly verbose and also incorrect -- maybe `existingMajorSymlink` instead ? (same for the other one right below this)

src/utils/__tests__/symlink.test.ts
```diff
@@ -0,0 +1,235 @@
...
+
+        // Make sure it points to the right file
+        const symlinkDestination = await fsPromises.readlink(symlinkFilePath);
+        expect(symlinkDestination).toBe(versionFile);
+      }
+    }, true)
+  );
+
+  describe('correctly handles already existing symlinks', () => {
+    it('when updating an old major version', async () =>
```
asottile-sentry, line 72: the `describe` / `it`s read backwards here -- I think they're supposed to be the other way around so the code "reads"?

asottile-sentry (commented):
new tests look useful!  thanks for those!

src/utils/symlink.ts
asottile-sentry: I think `logger.debug('symlinking', {before, after})` would probably be sufficient

src/utils/symlink.ts
Lms24: good catch!

src/utils/symlink.ts
Lms24: updated it again

asottile-sentry (approved):

PR #484, opened 2023-08-03
fix(github): Increase artifact polling interval and max retries

### LINKED ISSUE #482: GH Artifact lookup times out
It happens repeatedly for us that publishes time out when looking for artifacts uploaded to github. One example: https://github.com/getsentry/publish/actions/runs/5715907847/job/15486244767

This happens basically on every sentry-javascript release if we approve it while the checks are still running, so it seems to me something is not right with the waiting. Maybe we need to extend the `ARTIFACTS_POLLING_INTERVAL` (currently 10s), or extend the `MAX_TRIES` (currently 3)?

Lms24:
As outlined in [#482 by @BYK ](https://github.com/getsentry/craft/issues/482#issuecomment-1664067774) we should increase the time we wait for the build artifacts so that the publishing process doesn't fail if GH didn't upload artifacts within 30s.

closes #482

### DIFF
--- a/src/artifact_providers/__tests__/github.test.ts
+++ b/src/artifact_providers/__tests__/github.test.ts
@@ -1,21 +1,13 @@
 jest.mock('../../utils/githubApi.ts');
-import { getGitHubClient } from '../../utils/githubApi';
-import {
-  GitHubArtifactProvider,
-  ArtifactItem,
-  lazyRequest,
-  lazyRequestCallback,
-} from '../github';
+import { getGitHubClient} from '../../utils/githubApi';
+import { GitHubArtifactProvider, ArtifactItem, lazyRequest, lazyRequestCallback} from '../github';
 import { sleep } from '../../utils/async';
 
 class TestGitHubArtifactProvider extends GitHubArtifactProvider {
   public testGetRevisionArtifact(revision: string): Promise<ArtifactItem> {
     return this.getRevisionArtifact(revision);
   }
-  public testSearchForRevisionArtifact(
-    revision: string,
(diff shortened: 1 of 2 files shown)


BYK:
Looks like you need to update the test snapshots now:

```shell
yarn test -u
```

lforst (approved):

Lms24:
Hmm I don't know why the pr-check job is failing. Anyone have an idea why/how to fix?

src/artifact_providers/__tests__/github.test.ts
```diff
@@ -1,13 +1,21 @@
 jest.mock('../../utils/githubApi.ts');
-import { getGitHubClient} from '../../utils/githubApi';
-import { GitHubArtifactProvider, ArtifactItem, lazyRequest, lazyRequestCallback} from '../github';
+import { getGitHubClient } from '../../utils/githubApi';
+import {
+  GitHubArtifactProvider,
+  ArtifactItem,
+  lazyRequest,
+  lazyRequestCallback,
+} from '../github';
```
asottile-sentry, line 8: the reformat makes it really difficult to see what changed here -- can you discard those changes?

can probably set up formatting separately

asottile-sentry (approved):

Lms24:
I'd love to 🚢 but I can't merge as long as the `pr-check` job is failing. I have 0 context around google cloud build and the logs don't really help me. Any ideas what could be going wrong?

asottile-sentry:
I killed cloudbuild so this should be better now

Lms24:
Closing due to https://github.com/getsentry/craft/issues/482#issuecomment-1670951174

PR #488, opened 2023-08-28
fix(brew): Replace version in artifact names with '__VERSION__' to access checksums from mustache

romtsn:
This PR is a prerequisite for publishing sentry-wizard to brew (see https://github.com/getsentry/sentry-wizard/pull/406)

When artifact names contain versions (e.g. [sentry-wizard-v3.10.0.tgz](https://github.com/getsentry/sentry-wizard/releases/tag/v3.10.0)), it's not possible to access their checksums when publishing to brew, because mustache does not support nested variables. The template for brew formula would have to look something like this, but I couldn't find how to make it work for mustache:

```
sha256 "{{checksums.sentry-wizard-v{{version}}.tgz}}"
```

So this PR is simply replacing the version with `__VERSION__`, if it appears to be in the name of the artifact, when building the checksums dictionary.

Also updated readme to clarify the restrictions on the keys for checksums for the brew target.

Part of https://github.com/getsentry/sentry-wizard/issues/346

### DIFF
--- a/src/targets/brew.ts
+++ b/src/targets/brew.ts
@@ -163,7 +163,7 @@ export class BrewTarget extends BaseTarget {
     const checksums: any = {};
 
     await mapLimit(filesList, MAX_DOWNLOAD_CONCURRENCY, async file => {
-      const key = file.filename.replace(version, 'latest');
+      const key = file.filename.replace(version, '__VERSION__');
       checksums[key] = await this.artifactProvider.getChecksum(
         file,
         HashAlgorithm.SHA256,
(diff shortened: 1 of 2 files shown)


src/targets/brew.ts
```diff
@@ -163,7 +163,8 @@ export class BrewTarget extends BaseTarget {
     const checksums: any = {};
 
     await mapLimit(filesList, MAX_DOWNLOAD_CONCURRENCY, async file => {
-      checksums[file.filename] = await this.artifactProvider.getChecksum(
+      const key = file.filename.replace(version, 'latest');
```
tonyo, line 166: To make it more visually noticeable in the resulting key, wdyt about replacing the version with "__VERSION__"?
"latest" might be a bit misleading (it's somewhat subjective, and might potentially refer to the very latest version of the package, which changes over time).

romtsn: makes sense! I'll change it

tonyo (approved):

PR #491, opened 2023-09-12
Bump symbol collector 1.12.0

bruno-garcia:
https://github.com/getsentry/symbol-collector/compare/1.8.0...1.12.0


asottile-sentry (approved):
the install for this is a little weird -- I'd expect it to use a fixed url (that gets updated) and a sha256 sum to validate it -- but that's out of scope for this PR

PR #492, opened 2023-10-17
feat: Add `commit-on-git-repository` target

lforst:
Adds a `commit-on-git-repository` target which adds a commit to a git repository by consuming a tarball archive.

This is gonna be used for publishing deno modules: https://docs.deno.com/runtime/manual/advanced/publishing


src/targets/gitRepository.ts
```diff
@@ -0,0 +1,150 @@
+import simpleGit from 'simple-git';
```
asottile-sentry, line 1: please add tests for this new target

asottile-sentry: one other thing -- I think this might be too generically named (it's for a very specific extract-code-to-repository rather than something generically for managing a git repo)

README.md
```diff
@@ -1192,6 +1194,32 @@ targets:
   - name: hex
 ```
 
+### Git Repository (`commit-on-git-repository`)
```
mydea, line 1197:
```suggestion
### Commit on Git Repository (`commit-on-git-repository`)
```

README.md
```diff
@@ -1192,6 +1194,32 @@ targets:
...
+
+Takes a tarball and pushes the contents to a git repository.
```
mydea, line 1199:
```suggestion
Takes a tarball and pushes the unpacked contents to a git repository.
```

Not 100% sure, but maybe this is more explicit...?

lforst: Makes sense to me!

src/targets/__tests__/commitOnGitRepository.test.ts
```diff
@@ -0,0 +1,51 @@
...
+
+const mockClone = jest.fn();
+const mockCheckout = jest.fn();
+const mockRaw = jest.fn();
+const mockCommit = jest.fn();
+const mockAddTag = jest.fn();
+const mockPushTags = jest.fn();
+
+// eslint-disable-next-line no-var
+var mockExecSync: jest.Mock;
```
mydea, line 11: hmmm does it not work with the same approach here? (so `const mockExecSync = jest.fn()` and then use this in the `jest.mock` call?) this is sometimes weird, but just to be sure - if that does not work maybe add a comment here why this is that way?

mydea (approved):
LGTM! 🚀

src/targets/__tests__/commitOnGitRepository.test.ts
lforst: I think jest hoists the .mock() calls and they end up above the binding and if it's a `const` it will throw. `var`s get hoisted so they're fine. I sometimes hate jest. I'll play around with it a bit more and if I can't find a clean solution I'll just add a comment.

src/targets/__tests__/commitOnGitRepository.test.ts
```diff
@@ -0,0 +1,51 @@
...
+  expect(execSyncSpy).toHaveBeenCalledWith(
+    'tar -zxvf /tmp/my-archive.tgz --strip-components 1',
+    expect.objectContaining({ cwd: expect.any(String) })
+  );
+  expect(mockRaw).toHaveBeenCalledWith('add', '--all');
+  expect(mockCommit).toHaveBeenCalledWith('release: 1.2.3');
+  expect(mockAddTag).toHaveBeenCalledWith('1.2.3');
+  expect(mockRaw).toHaveBeenCalledWith('push', '--force');
+  expect(mockPushTags).toHaveBeenCalled();
+});
```
asottile-sentry, line 51: this isn't a good validation -- it doesn't actually test anything (it's a "copy paste mock" sort of test where you've just copied the implementation to the test).  it's a fragile test that doesn't actually demonstrate the interesting behaviour of the functionality

please write a test which actually performs an extraction on a real tar and inspects the resulting git repository (you can skip the download / upload / push parts at least since those are the side-effects you can't test)

lforst: I'd end up testing the simple-git library/tar which I don't see a case for. Use-Effort wise in my situation it doesn't make sense to add a testing setup for downloading release artifacts. I'm just gonna follow the convention in this repo which seems to be rather shallow testing.

This target is likely gonna be a one-off for a long time and the SDK team is happy to own it. If there are any issues we'll fix them and we will treat it as an incident if a wider scope is impacted.

asottile-sentry: we're all trying to make craft better -- please follow up with an improved test

PR #495, opened 2023-10-23
feat(commit-on-git-repository): Allow for authentication in github

lforst:
- Limit valid repository URLs to only https. This is technically breaking but impact is marginal and it was likely never working in the first place.
- Allow for authentication when cloning and pushing via a GitHub PAT token via the `GITHUB_API_TOKEN` env variable.

### DIFF
--- a/src/targets/commitOnGitRepository.ts
+++ b/src/targets/commitOnGitRepository.ts
@@ -136,8 +136,15 @@ export async function pushArchiveToGitRepository({
 
       logger?.info(`Cloning ${repositoryUrl} into ${directory}...`);
 
-      // We can assume that the provided url is a valid URL (otherwise parsing will crash)
-      const parsedUrl = new URL(repositoryUrl);
+      let parsedUrl;
+      try {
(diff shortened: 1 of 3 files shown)


mydea (approved):

Lms24 (approved):

src/targets/commitOnGitRepository.ts
```diff
@@ -134,7 +135,18 @@ export async function pushArchiveToGitRepository({
       const git = simpleGit(directory);
 
       logger?.info(`Cloning ${repositoryUrl} into ${directory}...`);
-      await git.clone(repositoryUrl, directory);
+
+      // We can assume that the provided url is a valid URL (otherwise parsing will crash)
+      const parsedUrl = new URL(repositoryUrl);
```
Lms24, line 140: l: I'm fine with letting things crash but let's maybe add an error message saying that only https urls will work, wdyt?

lforst: Makes sense! b3becf94f1eeb977956aedf517167af8086b853e

PR #496, opened 2023-10-23
ref: Pin cocoapods version 

brustolin:
Pinning `cocoapods` and `activesupport` versions. 
This is a temporary solution for [this problem ](https://github.com/CocoaPods/CocoaPods/issues/12081).
We should remove the `activesupport` pinning after cocoapods is fixed.


Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM node:14-buster-slim as builder
 
 WORKDIR /usr/local/lib
 
-COPY package.json yarn.lock ./
+COPY package.json yarn.lock Gemfile Gemfile.lock ./
```
asottile-sentry, line 5: this should not get copied here -- this invalidates every layer when the gemfile changes but it is not used until a much much later layer

Gemfile
```diff
@@ -0,0 +1,4 @@
+source "https://rubygems.org"
+
+gem "activesupport", "= 7.0.8"
+gem "cocoapods", "= 1.13.0"
```
philipphofmann: `m`: @brustolin, please add a comment explaining why we need to pin these versions, and why we can't use `gem install cocoapods --no-document -v 1.21.11.

Dockerfile
```diff
@@ -61,7 +61,10 @@ RUN curl -fsSL https://packages.microsoft.com/config/debian/10/packages-microsof
   && curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s --  --profile minimal -y \
   && cargo --version \
   && cargo install cargo-hack \
-  && gem install cocoapods --no-document \
+  && cargo install cocoapods -v 1.13.0 \
+  # temporary workaround for cocoapod crash
```
philipphofmann, line 65: `m`: Please add a link to the issue and a short explanation.

philipphofmann (approved):
If it works, LGTM.

Dockerfile
```diff
@@ -61,7 +61,10 @@ RUN curl -fsSL https://packages.microsoft.com/config/debian/10/packages-microsof
   && curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s --  --profile minimal -y \
   && cargo --version \
   && cargo install cargo-hack \
-  && gem install cocoapods --no-document \
+  && gem install cocoapods -v 1.13.0 \
```
asottile-sentry, line 64: this should have a fully pinned gemfile instead of pinning a handful of things -- this is going to be fragile as written

asottile-sentry (changes requested):

Dockerfile
```diff
@@ -61,7 +63,7 @@ RUN curl -fsSL https://packages.microsoft.com/config/debian/10/packages-microsof
   && curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s --  --profile minimal -y \
   && cargo --version \
   && cargo install cargo-hack \
-  && gem install cocoapods --no-document \
+  && gem install -g -no-document\
```
asottile-sentry, line 66: this command is incorrect (see the original command) and the space should be preserved

Gemfile
```diff
@@ -0,0 +1,5 @@
+# Gemfile
+# activesupport 7.1.1 is broken, and cocoapods dont have a pinned
```
asottile-sentry, line 2: dont -> doesn't

.dockerignore
```diff
@@ -7,3 +7,5 @@
 !/scripts
 !/src
 !/yarn.lock
+!/Gemfile
+!/Gemfile.lock
```
asottile-sentry, line 11: these were sorted -- can you fix that?

asottile-sentry (approved):

philipphofmann (commented):
Thanks @brustolin.

PR #500, opened 2023-11-09
chore: Bump .NET SDK 8 RC2

bitsandfoxes:
The .NET SDK would like to release an alpha for the upcoming .NET 8 release and requires the .NET 8 RC2 to be available for that.

Once the actual release goes live we can drop .NET 7 in favor of 8.


vaind:
Oh, we've got a little race condition here :) https://github.com/getsentry/craft/pull/501

bitsandfoxes:
Superseded by #501

Issue #498, opened 2023-11-07
Add config for latest tagging branches


mydea:
## Problem

When publishing backported versions, they are auto-set as latest in various targets (npm, github, possibly others). This means that you either need to take manual steps _or_ have to follow up with a release of the "actual" latest, which is far from ideal. 

For npm, you can circumenvent this by putting something like this on the backported branchs package.json:

```json
"publishConfig": {
  "tag": "v1"
}
```

For github, there is no way to do that (and e.g. Homebrew depends on this)

## Proposed solution

We propose to add a new top level config to allow to opt-in to define a list of target branches that should get the latest tag:

```yaml
# .craft.yml
minVersion: '0.23.1'
changelogPolicy: simple
latestTargetBranches: 
  - master
```

If this is configured, craft targets should only add the `latest` tag (or similar behavior, depending on the target) when merging into one of the defined branches.

With this, projects can opt-in to only release as `latest` when merging into e.g. master/main branch, and when merging into e.g. a `1.x` branch we can ensure to not mark this release as latest.

BYK:
Craft already knows about the "default" branch so it feels more appropriate to make this the default behavior?

Fir backwards compatibility, you can internally enable this only for config files setting a minVersion above the change date.

mydea:
Sounds good to me too!

So basically you mean:

* Only tag with latest if merging into the "default" branch
* Only do this if minVersion > version we introduced this

mydea:
Actually, we just thought about this and this will not necessarily work - as the default branch is not necessarily the branch where you run releases from.

Concretely, in sentry-javascript the default branch is develop, but we release from/to the master branch (we use gitflow and still merge into develop while a release is pending etc). So for cases like these you'd still need to configure this.

We can still do both I guess, do this by default for the default branch, but allow to configure/overwrite this if needed?

BYK:
>Concretely, in sentry-javascript the default branch is develop, but we release from/to the master branch (we use gitflow and still merge into develop while a release is pending etc). So for cases like these you'd still need to configure this.

Ah, that's quite unfortunate (and unexpected for me) 😅 

>We can still do both I guess, do this by default for the default branch, but allow to configure/overwrite this if needed?

I'd say yes but I feel like it may complicate things unnecessarily if you were looking to get this out of the door quickly. Use your own judgement then :) For the given scenario, your initial proposal makes sense to me and easy and harmless enough to start with.

PR #502, opened 2023-11-15
feat: Allow to configure `latestTargetBranch`

mydea:
This PR allows to add a new top-level config `latestTargetBranch`, which can be set to a branch name.

Only when merging into this branch name, will targets try to set the given release as "latest". 

Going forward, if your minVersion > 1.8.0, the default behavior will be to use the default branch as default value for this. For older minVersion, the current behavior will remain (=every release will be tagged as latest).

I have implemented this for:

* npm
* github

A few targets try to do this automatically already (e.g. registry or awsLambdaLayer update `latest.json` based on the previous version number). We can also add this for further targets that do this later (I don't know how all of them work) - but this is very much additive so we're safe to ignore this were it's not supported yet (=behavior will just remain the same).

Closes https://github.com/getsentry/craft/issues/498

### DIFF
--- a/src/targets/npm.ts
+++ b/src/targets/npm.ts
@@ -54,6 +54,8 @@ interface NpmPublishOptions {
   otp?: string;
   /** New version to publish */
   version: string;
+  /** If defined, set this tag instead of "latest" */
+  tag?: string;
 }
 
 /**
@@ -178,14 +180,8 @@ export class NpmTarget extends BaseTarget {
       args.push(`--access=${this.npmConfig.access}`);
     }
 
-    // In case we have a prerelease, there should never be a reason to publish
-    // it with the latest tag in npm.
-    if (isPreviewRelease(options.version)) {
-      this.logger.warn('Detected pre-release version for npm package!');
-      this.logger.warn(
-        'Adding tag "next" to not make it "latest" in registry.'
-      );
-      args.push('--tag=next');
+    if (options.tag) {
+      args.push(`--tag=${options.tag}`);
     }
 
     return withTempFile(filePath => {
@@ -219,7 +215,11 @@ export class NpmTarget extends BaseTarget {
    * @param version New version to be released
    * @param revision Git commit SHA to be published
    */
-  public async publish(version: string, revision: string): Promise<any> {
+  public async publish(
+    version: string,
+    revision: string,
+    isLatest: boolean
+  ): Promise<any> {
     this.logger.debug('Fetching artifact list...');
     const packageFiles = await this.getArtifactsForRevision(revision, {
       includeNames: DEFAULT_PACKAGE_REGEX,
@@ -233,6 +233,18 @@ export class NpmTarget extends BaseTarget {
     const publishOptions: NpmPublishOptions = { version };
     if (!isDryRun() && this.npmConfig.useOtp) {
       publishOptions.otp = await this.requestOtp();
+    }
+
+    // In case we have a prerelease, there should never be a reason to publish
+    // it with the latest tag in npm.
+    if (isPreviewRelease(version)) {
+      this.logger.warn('Detected pre-release version for npm package!');
+      this.logger.warn(
+        'Adding tag "next" to not make it "latest" in registry.'
+      );
+      publishOptions.tag = 'next';
+    } else if (!isLatest) {
+      publishOptions.tag = 'old';
     }
 
     await Promise.all(
--- a/README.md
+++ b/README.md
@@ -322,6 +322,19 @@ for a release is `{releaseBranchPrefix}/{version}`. The prefix defaults to
 releaseBranchPrefix: publish
 ```
 
+### Latest Branch Name
+
+If configured, only tag releases as "latest" when merging into this branch.
+This can be useful to make sure that releases from e.g. a `v1` branch do not get tagged as "latest".
+Note that not all targets may use this.
+
+```yaml
+latestTargetBranch: main
+```
+
+If you specific `minVersion: '1.8.0'` or above, this will default to use the default branch.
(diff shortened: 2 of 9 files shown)


asottile-sentry (changes requested):
why can't this just be the default branch?  that seems the most logical target and wouldn't need configuration

if we're going to go forward with this: this new feature needs to be implemented in all the relevant targets before merging

mydea:
As discussed in the linked issue: https://github.com/getsentry/craft/issues/498 this does not work for e.g. sentry-javascript, where we use a GitFlow process with a develop & master branch. `develop` is the default branch but we release off of `master`.

I have implemented this for all targets that I know off that do that. If you or anybody else knows of other targets that set a `latest` tag or similar, let me know (or just feel free to implement them yourself). IMHO this is already very valuable for npm & github only, we've had concrete issues opened about this and it is a massive footgun to publish a backported version and be surprised by `latest` being incorrectly updated.

asottile-sentry:
what benefit is githlow getting you -- it sounds like it's most of the problem and isn't used elsewhere at the company

mydea:
> what benefit is gitflow getting you -- it sounds like it's most of the problem and isn't used elsewhere at the company

The JS monorepo is rather different from most other SDKs because we simply merge a lot. Since doing a release can take quite some time (for tests to pass etc. it can be 30min or more, and with feedback cycles we can have a pending release for hours, sometimes), we don't want to block any merging while a release is in process. This is also hard to enforce because so many different people contribute to the repo at the same time, and not all of them are in the core SDK team that are on top of publishing etc. 

but also, I don't really see how this is relevant to this - craft should not enforce a git branching process, I'd say. 

Also, this PR is specifically built so that this will use the default branch by default, it's just an additonal config that _can_ be added _if_ you do not release off of your default branch. So for the "most common" scenario you do not need to configure anything.

asottile-sentry:
I'm not convinced we need this at all -- the `registry` target seems to handle this correctly without the complexity (it knows when the release is non-latest and does not update latest)

src/targets/npm.ts
```diff
@@ -235,6 +235,18 @@ export class NpmTarget extends BaseTarget {
...
+    // In case we have a prerelease, there should never be a reason to publish
+    // it with the latest tag in npm.
+    if (isPreviewRelease(version)) {
+      this.logger.warn('Detected pre-release version for npm package!');
+      this.logger.warn(
+        'Adding tag "next" to not make it "latest" in registry.'
+      );
+      publishOptions.tag = 'next';
+    } else if (!isLatest) {
+      publishOptions.tag = 'old';
```
Lms24, line 247: q: I guess we need to set  _some_ tag so that NPM doesn't default to `latest`?

Lms24 (approved):
Looks good to me, thanks for taking care of this!

lforst:
@asottile-sentry Before we switched our release process to gitflow, our releases sometimes took upwards of 3 hours because accidental pushes to the main branch invalidated them because craft requires the release branch to be up to date. Asking people not to merge because a release is pending did not work for organizational reasons. We have multiple teams merging into `sentry-javascript`. _We switched to gitflow out of necessity because craft screwed us._

Frankly, I am genuinely interested in the concrete reason why you're blocking this. I hope it's not because you're worried that you'll be pinged about issues with craft - cause more than you or anybody else, we're ourselves affected by it. We are merely trying to solve issues here that have been plaguing us for years. Just today, I had to do a double release again because craft doesn't support releasing old versions.

asottile-sentry:
my main concern is bolting on more complexity without tests to a tool which is under-tested and under-supported.  there are simpler ways to accomplish what you need to solve your problems with the existing interfaces and I'd strongly prefer fixing those rather than adding yet-another-untested option.

lforst:
> there are simpler ways to accomplish what you need to solve your problems with the existing interfaces

Can you point these out?

asottile-sentry:
> > there are simpler ways to accomplish what you need to solve your problems with the existing interfaces
> 
> Can you point these out?

`registry` manages to detect when a version is non-latest and avoids updating symlinks -- we can do the same for npm / github

mydea:
> `registry` manages to detect when a version is non-latest and avoids updating symlinks -- we can do the same for npm / github

this is not that easy because our current flow for npm simply does not know what package name we are even publishing. We only know the version & the path to a tarball. We'd need to unpack all tarballs, read the package.json from the file, then make a request to npm via e.g. `npm view {pkgName} version` in order to be able to compare the versions. If you think this is less complexity/easier to do, I'm also happy to implement it this way. 

For Github, we need to fetch the latest release, which should hopefully be possible via octokit (*but haven't checked yet).

asottile-sentry:
> > `registry` manages to detect when a version is non-latest and avoids updating symlinks -- we can do the same for npm / github
> 
> this is not that easy because our current flow for npm simply does not know what package name we are even publishing. We only know the version & the path to a tarball. We'd need to unpack all tarballs, read the package.json from the file, then make a request to npm via e.g. `npm view {pkgName} version` in order to be able to compare the versions. If you think this is less complexity/easier to do, I'm also happy to implement it this way.
> 
> For Github, we need to fetch the latest release, which should hopefully be possible via octokit (*but haven't checked yet).

ah I see -- that is a little more complicated than I thought it was.  registry has the advantage that it can see the entire version tree all on its own.

let's pursue the current approach then -- but please write some tests which demonstrate this feature

mydea:
Based on your feedback, I ended up splitting this up in two PRs and instead tried to implement the latest tagging in the github & npm targets specifically:

* https://github.com/getsentry/craft/pull/504
* https://github.com/getsentry/craft/pull/503

For github, that should work easily enough. For NPM I introduced a new config option for npm specifically, as that is the main thing that is tricky to infer, sadly.

PR #503, opened 2023-11-16
feat(github): Only set `latest` when new version > old version

mydea:
Automatically set `latest` for github releases only when the new version > the current latest version.
If we can't find a current latest, or something else goes wrong, we default to setting as latest.

Supersedes https://github.com/getsentry/craft/pull/502

### DIFF
--- a/src/targets/github.ts
+++ b/src/targets/github.ts
@@ -15,7 +15,12 @@ import {
 } from '../utils/changelog';
 import { getGitHubClient } from '../utils/githubApi';
 import { isDryRun } from '../utils/helpers';
-import { isPreviewRelease, versionToTag } from '../utils/version';
+import {
+  isPreviewRelease,
+  parseVersion,
+  versionGreaterOrEqualThan,
+  versionToTag,
+} from '../utils/version';
 import { BaseTarget } from './base';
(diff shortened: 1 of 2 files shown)


src/targets/github.ts
```diff
@@ -122,13 +127,23 @@ export class GitHubTarget extends BaseTarget {
       };
     }
 
+    const { data: latestRelease } = await this.github.repos.getLatestRelease({
```
lforst, line 130: Stupid question. What happens here?:

- actual latest release is v3.0
- previous release published was v1.1
- you're now publishing v2.2

mydea: 2.2 will _not_ be marked as latest. It fetches the release that is currently marked as "latest" on Github (3.0), and if that has a version > than the version we are trying to publish, it does not mark it as latest!

Lms24 (approved):

asottile-sentry (approved):

src/targets/github.ts
asottile-sentry, line 130: this crashes when there are no releases in a repository -- can you fix this regression?

mydea: I will look into it!

PR #504, opened 2023-11-16
feat(npm): Allow to configure `checkPackageName` for npm target

mydea:
This updates the `npm` target with the following new behavior:

* If the new config for the npm target `checkPackageName` is defined, we check the current version of this package on npm
* If this version is newer than the to-be-published version, we tag it as "old" instead of "latest".

This supersedes https://github.com/getsentry/craft/pull/502.

### DIFF
--- a/src/targets/__tests__/npm.test.ts
+++ b/src/targets/__tests__/npm.test.ts
@@ -1,0 +1,129 @@
+import { parseVersion } from '../../utils/version';
+import { getPublishTag, getLatestVersion } from '../npm';
+
+function setNpmToken() {
+  process.env.NPM_TOKEN = 'test-token';
+}
+
+const defaultNpmConfig = {
+  useYarn: false,
+  token: 'xxx',
+};
+
+describe('getLatestVersion', () => {
+  beforeEach(() => {
+    setNpmToken();
+  });
+
+  it('returns undefined for unexisting checkPackageName', async () => {
+    const actual = await getLatestVersion(
+      'sentry-xx-this-does-not-exist',
+      defaultNpmConfig
+    );
+    expect(actual).toEqual(undefined);
+  });
+
+  it('returns version for valid id', async () => {
+    const actual = await getLatestVersion('@sentry/browser', defaultNpmConfig);
+    expect(actual).toBeDefined();
+
+    // Returns a valid version
+    const parsed = parseVersion(actual as string);
+    expect(parsed).toBeDefined();
+    expect(parsed?.major).toBeGreaterThanOrEqual(7);
+  });
+});
+
+describe('getPublishTag', () => {
+  beforeEach(() => {
+    setNpmToken();
+  });
+
+  it('returns undefined without a checkPackageName', async () => {
+    const logger = {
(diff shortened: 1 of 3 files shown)


src/targets/npm.ts
```diff
@@ -246,3 +263,95 @@ export class NpmTarget extends BaseTarget {
...
+  // We tag it with "old" instead of "latest"
+  if (
+    parsedNewVersion &&
+    !versionGreaterOrEqualThan(parsedNewVersion, parsedLatestVersion)
+  ) {
+    logger.warn(
+      `Detected older version than currently published version (${latestVersion}) for ${checkPackageName}`
+    );
+    logger.warn('Adding tag "old" to not make it "latest" in registry.');
+    return 'old';
```
Lms24, line 353: theoretically (not saying this should happen now), we could assign a `<major>.x` tag here if we decide we want to go this route for multi-major version support.

src/targets/npm.ts
```diff
@@ -246,3 +263,95 @@ export class NpmTarget extends BaseTarget {
...
+
+    return response.toString().trim();
+  } catch {
+    return undefined;
+  }
+}
+/**
+ * Get the tag to use for publishing to npm.
+ */
+export async function getPublishTag(
```
Lms24, line 310: l: l: I'd either document that returning `undefined` later on causes npm to assign the latest tag automatically or we return `latest` here as a default. wdyt?

Lms24 (approved):

src/targets/__tests__/npm.test.ts
```diff
@@ -0,0 +1,129 @@
+import { parseVersion } from '../../utils/version';
+import { getPublishTag, getLatestVersion } from '../npm';
+
+function setNpmToken() {
+  process.env.NPM_TOKEN = 'test-token';
+}
```
asottile-sentry, line 6: this never gets cleaned up -- there should be some context-like assignment to a global

src/targets/__tests__/npm.test.ts
```diff
@@ -0,0 +1,129 @@
...
+describe('getPublishTag', () => {
+  beforeEach(() => {
+    setNpmToken();
+  });
+
+  it('returns undefined without a checkPackageName', async () => {
+    const logger = {
+      warn: jest.fn(),
+    } as any;
+    const actual = await getPublishTag(
```
asottile-sentry, line 46: hmmm is this hitting actual public npm during test? we probably want to avoid that if possible since it makes these tests dependent on external state

mydea: yes, this hits npm. I tried to write the test in a way that it works with whatever version we get back. But yes, if NPM would be down this would fail. Any idea how else to test this? 🤔 if we don't want to hit NPM the test also gets less realistic. The only thing I can think of is to mock `spawnProcess` and have it return a static version?

mydea: @asottile-sentry I've updated the test to use a mocked process there instead!

asottile-sentry (approved):

PR #509, opened 2023-11-30
Changelog prerelease command fix

hubertdeng123:
Adds a condition so that if preReleaseCommand is empty, but changelog changes still need to be committed, craft handles it

### DIFF
--- a/src/commands/prepare.ts
+++ b/src/commands/prepare.ts
@@ -514,7 +514,12 @@ export async function prepareMain(argv: PrepareOptions): Promise<any> {
   if (preReleaseCommandRan) {
     // Commit the pending changes
     await commitNewVersion(git, newVersion);
-  } else {
+  }
+  // If changelog is set but preReleaseCommand is empty string, there may be commits we care about
+  else if (!argv.noChangelog && config.preReleaseCommand && config.preReleaseCommand.length === 0) {
+    await commitNewVersion(git, newVersion);
+  }
+  else {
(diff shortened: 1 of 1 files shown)


src/commands/prepare.ts
```diff
@@ -514,7 +514,12 @@ export async function prepareMain(argv: PrepareOptions): Promise<any> {
   if (preReleaseCommandRan) {
     // Commit the pending changes
     await commitNewVersion(git, newVersion);
-  } else {
+  }
+  // If changelog is set but preReleaseCommand is empty string, there may be commits we care about
+  else if (!argv.noChangelog && config.preReleaseCommand && config.preReleaseCommand.length === 0) {
+    await commitNewVersion(git, newVersion);
+  }
+  else {
```
chadwhitacre, line 522: Reading the code I think it would be better to always call `commitNewVersion` but update it to not hard error if there is nothing to commit. Earlier we ensure that the git status is clear before we start doing things, so we can trust that any changes that do now exist are supposed to be there.

asottile-sentry: either way this should also have a test demonstrating what you're trying to accomplish.  it's sorta unclear from the code change and topic and would be nice to ensure it doesn't regress from a future change

chadwhitacre: We should also review the documentation.

chadwhitacre:
This becomes urgent as we approach the next release, we'll hit the same issue in Vroom again otherwise.

PR #515, opened 2024-02-26
feat: powershell target

vaind:
This adds powershell publishing support

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -78,6 +78,13 @@ RUN curl -fsSL https://storage.googleapis.com/flutter_infra_release/releases/sta
   && tar xf /opt/flutter.tar.xz -C /opt \
   && rm /opt/flutter.tar.xz
 
+# https://learn.microsoft.com/en-us/powershell/scripting/install/install-debian
+RUN curl -fsSL https://github.com/PowerShell/PowerShell/releases/download/v7.4.1/powershell_7.4.1-1.deb_amd64.deb -o /opt/powershell.deb \
+  && dpkg -i /opt/powershell.deb \
+  && apt-get install -f \
+  && apt-get clean \
+  && rm /opt/powershell.deb
+
 # craft does `git` things against mounted directories as root
 RUN git config --global --add safe.directory '*'
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -65,7 +65,6 @@ RUN curl -fsSL https://packages.microsoft.com/config/debian/10/packages-microsof
   && cargo --version \
   && cargo install cargo-hack \
   && gem install -g --no-document \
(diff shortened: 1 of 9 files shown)


src/utils/version.ts
```diff
@@ -89,6 +89,8 @@ export function versionGreaterOrEqualThan(v1: SemVer, v2: SemVer): boolean {
     return true;
   } else if (v1.pre && !v2.pre) {
     return false;
+  } else if (v1.pre && v2.pre && v1.pre === v2.pre) {
+    return v1.build === v2.build;
```
asottile-sentry, line 93: I don't think this is correct for this function -- it's meant to implement `>=` so only checking strict equality isn't quite right

src/utils/version.ts
```diff
@@ -157,7 +159,6 @@ export function getPackageVersion(): string {
  * Returns the stringified version of the passed SemVer object.
  */
 export function semVerToString(s: SemVer) {
-  return `${s.major}.${s.minor}.${s.patch}${s.pre ? `-${s.pre}` : ''}${
-    s.build ? `+${s.build}` : ''
-  }`;
+  return `${s.major}.${s.minor}.${s.patch}${s.pre ? `-${s.pre}` : ''}${s.build ? `+${s.build}` : ''
```
asottile-sentry, line 162: craft is going to have some other problems with this -- currently we don't allow `+` releases anywhere -- is this necessary?

Dockerfile
```diff
@@ -65,6 +65,7 @@ RUN curl -fsSL https://packages.microsoft.com/config/debian/10/packages-microsof
   && cargo --version \
   && cargo install cargo-hack \
   && gem install -g --no-document \
+  && gem cleanup all \
```
asottile-sentry, line 68: let's decouple this from your change and do it separately

asottile-sentry (commented):
can you write some tests for your new feature

src/utils/version.ts
vaind: this is a craft version check. I've hit this when testing the compiled craft locally - it comes with version `1.11.0-dev.0` and even though you specify that as `minVersion` in `.craft.yml`, the check fails without the changes I've made here.

bruno-garcia:
Thanks @vaind for raising this!

bruno-garcia:
@asottile-sentry please let me know if we need any other change to get this through

asottile-sentry (approved):
apologies I was out sick the last week (still am) but going to approve to unblock

vaind:
Thx.  Can you please merge as well? I don't have permissions

bruno-garcia:
Thanks @vaind and @asottile-sentry !

PR #520, opened 2024-02-28
fix: Minor linter issues

vaind:
- **add missing gem cleanup**
- **autofixed settings.json**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -69,6 +69,7 @@ RUN : \
   && cargo --version \
   && cargo install cargo-hack \
   && gem install -g --no-document \
+  && gem cleanup all \
   # Install https://github.com/getsentry/symbol-collector
   && symbol_collector_url=$(curl -s https://api.github.com/repos/getsentry/symbol-collector/releases/tags/1.17.0 | \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -65,6 +65,7 @@ RUN curl -fsSL https://packages.microsoft.com/config/debian/10/packages-microsof
   && cargo --version \
   && cargo install cargo-hack \
   && gem install -g --no-document \
+  && gem cleanup all \
```
asottile-sentry, line 68: just  because I'm curious -- what does this do?  reading the output it looks like it didn't really uninstall anything:

```
#16 109.4 Cleaning up installed gems...
#16 109.4 Clean up complete
```

vaind: guess nothing in this case, it's just what linter said is missing there. To me it sounds like having apt-get clean or apt-get autoremove - you may not need it most of the time but if you do, it's there and no need to come back to the image and investigate whether you can decrease the size.

asottile-sentry: what linter ?

vaind: some extension in vscode uses these: https://github.com/projectatomic/dockerfile_lint

PR #521, opened 2024-02-28
fix: version test for equal pre-release versions

vaind:
I've hit this when testing the compiled craft locally - it comes with version 1.11.0-dev.0 and even though you specify that as minVersion in .craft.yml, the check fails without the changes I've made here.

### DIFF
--- a/src/utils/__tests__/version.test.ts
+++ b/src/utils/__tests__/version.test.ts
@@ -195,6 +195,25 @@ describe('versionGreaterOrEqualThan', () => {
     expect(() => versionGreaterOrEqualThan(v1, v2)).toThrowError();
     expect(() => versionGreaterOrEqualThan(v2, v1)).toThrowError();
   });
+
+  test('can compare pre parts that are the same', () => {
+    let v1 = parseVersion('1.2.3-dev.0')!;
+    let v2 = parseVersion('1.2.3-dev.0')!;
+    expect(versionGreaterOrEqualThan(v1, v2)).toBe(true);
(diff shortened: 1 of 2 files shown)


vaind:
> I'm pretty sure we don't want to support `minVersion: {a dev version}` since that'll never be a valid value when checked in -- is this still needed?

the code in the repo before this change is wrong, as illustrated by the added test cases. Whether you call the pre-release a "dev" release or if it's a beta, you should be able to set a `minVersion` in a `.craft.yml` to match that version.

Without the changes in this PR, this code fails, which is obviously wrong, just from the name of the function.
```js
expect(versionGreaterOrEqualThan(parseVersion('1.2.3-beta'), parseVersion('1.2.3-beta'))).toBe(true);
```

asottile-sentry:
> > I'm pretty sure we don't want to support `minVersion: {a dev version}` since that'll never be a valid value when checked in -- is this still needed?
> 
> the code in the repo before this change is wrong, as illustrated by the added test cases. Whether you call the pre-release a "dev" release or if it's a beta, you should be able to set a `minVersion` in a `.craft.yml` to match that version.
> 
> Without the changes in this PR, this code fails, which is obviously wrong, just from the name of the function:
> 
> ```js
> expect(versionGreaterOrEqualThan(parseVersion('1.2.3-beta'), parseVersion('1.2.3-beta'))).toBe(true);
> ```

we haven't, and don't plan to, release craft prereleases so there's no reason to make the change

vaind:
I just don't get why you insist on keeping the code broken, `versionGreaterOrEqualThan` saying false for identical argument is just that...

but whatever

PR #523, opened 2024-03-12
ref: upgrade to node 16

asottile-sentry:
this is a stepping stone to get to a more modern docker / node


IanWoodard (approved):

asottile-sentry:
I've determined the zip issue is only present in node 18+ so I'm going to proceed with this and try and track down the zip issue separately

PR #530, opened 2024-03-26
ref: upload to twine synchronously in a single call

asottile-sentry:
I suspect pypi can't handle parallel requests


joshuarli (approved):
looks like typescript

PR #532, opened 2024-04-11
feat: Add `createdAt` key to json in registry

HazAT:
We want to use the data from our release registry more, one info that is missing today is the timestamp of when something was released.
Yes, in theory, it exists in git, but just for convenience, we want to add a key `createdAt` into every JSON that the Registry target writes for easier access later.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 1.18.0
 
-- feat: Registry target now adds key `created` with current timestamp to json file
+- feat: Registry target now adds key `createdAt` with current timestamp to json file
 
 ## 1.17.2
 
@@ -112,7 +112,7 @@
 
 ### Various fixes & improvements
 
-- ref: Pin cocoapods version  (#496) by @brustolin
+- ref: Pin cocoapods version (#496) by @brustolin
 
 ## 1.6.0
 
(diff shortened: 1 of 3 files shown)


src/targets/registry.ts
```diff
@@ -362,7 +362,11 @@ export class RegistryTarget extends BaseTarget {
       );
     }
     // Update the manifest
-    const updatedManifest = { ...packageManifest, version };
+    const updatedManifest: { [key: string]: any } = {
+      ...packageManifest,
+      version,
+    };
+    updatedManifest.createdAt = new Date().toISOString();
```
Lms24, line 369: Just to confirm: the manifest is what ends up [here](https://github.com/getsentry/sentry-release-registry/blob/master/packages/npm/%40sentry/browser/7.109.0.json), correct? If yes, then I think adding a new field should be safe

HazAT: Yes

Lms24 (approved):

Lms24 (approved):
thx!

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.18.0
+
+- feat: Registry target now adds key `createdAt` with current timestamp to json file
+
```
asottile-sentry, line 6: craft will auto generate this based on your PR commit message so this shoudl be left out

CHANGELOG.md
```diff
@@ -108,7 +112,7 @@
 
 ### Various fixes & improvements
 
-- ref: Pin cocoapods version  (#496) by @brustolin
+- ref: Pin cocoapods version (#496) by @brustolin
```
asottile-sentry, line 115: please revert the other (unrelated, cosmetic) edits to this file -- please make these in a separate PR if you want

src/targets/registry.ts
```diff
@@ -362,7 +362,15 @@ export class RegistryTarget extends BaseTarget {
...
     // Update the manifest
-    const updatedManifest = { ...packageManifest, version };
+    const updatedManifest: {
+      version: string;
+      createdAt: string;
+      [key: string]: any;
+    } = {
+      ...packageManifest,
+      version,
+      createdAt: new Date().toISOString(),
```
asottile-sentry, line 372: is there a plan to backfill this data?

any test that needs updating?

HazAT: Added a simple unit test to check if property is there
here is the PR to backfill the data for existing files
https://github.com/getsentry/sentry-release-registry/pull/148

asottile-sentry (approved):

PR #536, opened 2024-05-16
fix(github): Fix making github releases latest or not

mydea:
This PR updates the github release creation to ensure that the release is correctly marked as latest.

We used to set this for when we created the release draft, but according to https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#create-a-release, this cannot actually be set on drafts. Instead, we need to set it when we publish the release.

There is also an option `legacy`, which, according to the docs, _seems_ like it may also work:

> `legacy` specifies that the latest release should be determined based on the release creation date and higher semantic version.

But since this option is a legacy option, _and_ since the behavior is a bit under-specced (e.g. if it determines based on creation date, it may be wrong, if it always uses semantic version, it may be correct), we decided to implement this consistently for ourselves.

### DIFF
--- a/src/targets/github.ts
+++ b/src/targets/github.ts
@@ -425,6 +425,7 @@ export class GitHubTarget extends BaseTarget {
         : 'No previous release found'
     );
 
+    // Preview versions should never be marked as latest
     const isPreview =
       this.githubConfig.previewReleases && isPreviewRelease(version);
     const makeLatest = isPreview
--- a/src/targets/github.ts
+++ b/src/targets/github.ts
@@ -111,7 +111,8 @@ export class GitHubTarget extends BaseTarget {
   public async createDraftRelease(
     version: string,
     revision: string,
-    changes?: Changeset
+    changes?: Changeset,
+    isLatest = true
   ): Promise<GitHubRelease> {
     const tag = versionToTag(version, this.githubConfig.tagPrefix);
     this.logger.info(`Git tag: "${tag}"`);
@@ -126,26 +127,6 @@ export class GitHubTarget extends BaseTarget {
         upload_url: '',
       };
     }
-
-    let latestRelease: { tag_name: string } | undefined;
-    try {
-      latestRelease = (
-        await this.github.repos.getLatestRelease({
-          owner: this.githubConfig.owner,
-          repo: this.githubConfig.repo,
-        })
-      ).data;
-    } catch (error) {
-      // if the error is a 404 error, it means that no release exists yet
-      // all other errors should be rethrown
-      if (error.status !== 404) {
-        throw error;
-      }
-    }
-
-    const isLatest = isPreview
-      ? false
-      : isLatestRelease(latestRelease, version);
 
     const { data } = await this.github.repos.createRelease({
       draft: true,
@@ -347,7 +328,7 @@ export class GitHubTarget extends BaseTarget {
    *
    * @param release Release object
    */
-  public async publishRelease(release: GitHubRelease) {
+  public async publishRelease(release: GitHubRelease, isLatest = true) {
     if (isDryRun()) {
       this.logger.info(`[dry-run] Not publishing the draft release`);
       return;
@@ -356,6 +337,7 @@ export class GitHubTarget extends BaseTarget {
     await this.github.repos.updateRelease({
       ...this.githubConfig,
       release_id: release.id,
+      make_latest: isLatest ? 'true' : 'false',
       draft: false,
     });
   }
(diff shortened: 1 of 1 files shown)


asottile-sentry:
can you describe the issue? it's unclear what you're trying to solve for (additionally this PR makes changes beyond logging)

mydea:
> can you describe the issue? it's unclear what you're trying to solve for (additionally this PR makes changes beyond logging)

The problem is that it does not seem to be working, it is still marking the stuff as latest. But I can't see anything wrong, and the existing logs make it impossible to know where this failed (e.g. could it not find the previous release? is the version from there picked up incorrectly? ...)

Lms24 (approved):
LGTM!

asottile-sentry:
probably don't need the logging -- the problem is that the draft was created correctly (non-latest) but then converted to latest when moved out of draft because this defaults to `make_latest: true`: https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#update-a-release

BYK:
I think the additional logging makes sense but it's better not to change other stuff.

mydea:
> probably don't need the logging -- the problem is that the draft was created correctly (non-latest) but then converted to latest when moved out of draft because this defaults to `make_latest: true`: https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#update-a-release

Ahh, great catch. I'll update this accordingly! I think I'll leave the logging in, it can still be helpful I'd say? At least the log for what the previous release was!

asottile-sentry:
> > probably don't need the logging -- the problem is that the draft was created correctly (non-latest) but then converted to latest when moved out of draft because this defaults to `make_latest: true`: https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#update-a-release
> 
> Ahh, great catch. I'll update this accordingly! I think I'll leave the logging in, it can still be helpful I'd say? At least the log for what the previous release was!

🤷 personally I'd rather not just because it's even more output clutter -- and it wouldn't have helped us find the answer here

mydea:
So I updated this to ensure we set `make_latest` in `publishRelease` as well, which meant that I extracted this out to only call this once!

asottile-sentry:
as I requested above, and as I've done in many of your PRs: please limit the change to just the functional change instead of coupling it with unrelated (and imo undesirable) changes.  this makes it easier to review, easier to reason about, easier to rollback, and generally is how one should approach development

mydea:
I updated this to only include minimal changes necessary to make this work. The core change is that we ensure to also set `make_latest` when we call `updateRelease` now. I also included the fix from https://github.com/getsentry/craft/pull/567 as this is moved around here.

src/targets/github.ts
```diff
@@ -345,6 +337,7 @@ export class GitHubTarget extends BaseTarget {
     await this.github.repos.updateRelease({
       ...this.githubConfig,
       release_id: release.id,
+      make_latest: isLatest ? 'true' : 'false',
```
BYK, line 340: Why is this a string?

src/targets/github.ts
```diff
@@ -111,7 +111,8 @@ export class GitHubTarget extends BaseTarget {
   public async createDraftRelease(
     version: string,
     revision: string,
-    changes?: Changeset
+    changes?: Changeset,
+    isLatest = true
```
BYK, line 115:
```suggestion
    setAsLatest = true
```

Also, not to derail your patch but boolean function arguments are frowned upon as they make it very hard to see what's going on from the callsite: `createDraftRelease('1.2.3', 'aaaef32', undefined, false)` would be quite hard to make sense of.

You may want to consider converting all of them to an arguments object or just add a final `options` object with `setAsLatest` as its only known key for now?

src/targets/github.ts
```diff
@@ -336,7 +328,7 @@ export class GitHubTarget extends BaseTarget {
    *
    * @param release Release object
    */
-  public async publishRelease(release: GitHubRelease) {
+  public async publishRelease(release: GitHubRelease, isLatest = true) {
```
BYK, line 331: Same concern as above re boolean args and also naming. Would prefer something like `setAsLatest` or `makeLatest`

BYK (changes requested):
Requesting changes not because of the code but for the lack of a PR description explaining what was going wrong, what exactly we are fixing etc.

If this is an important change and if it's not too hard, adding a test would also be nice (I'm fine not adding if there are none though)

src/targets/github.ts
mydea: happy to do that, I tried to do minimal changes in the PR, but I agree - I'll convert this to an object!

src/targets/github.ts
mydea: Here: https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#create-a-release it is documented as type `string`, for whatever reason...

PR #540, opened 2024-05-23
fix(maven): Rename module.json to dist.module

markushi:
Currently various SDKs (sentry-java, kotlin multiplatform) generate maven ready artifacts which contain a `module.json` file instead of a e.g. `sentry-java-1.0.0.module` file. 

This PR simply renames the file right before publishing to maven.


asottile-sentry (approved):
thanks for the tests!

PR #543, opened 2024-06-17
Upgrade Flutter from 3.10.0 to 3.13.0 in docker image

buenaflor:
Currently Flutter `3.10.0` is packaged with Dart `3.0.0`.
Now we require Dart `3.1.0` which is packaged starting with Flutter `3.13.0`

See: https://github.com/getsentry/publish/actions/runs/9548988953/job/26317502315#step:11:104

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -78,7 +78,7 @@ RUN : \
   && chmod +x /usr/local/bin/SymbolCollector.Console
 
 # https://docs.flutter.dev/get-started/install/linux#install-flutter-manually
-RUN curl -fsSL https://storage.googleapis.com/flutter_infra_release/releases/stable/linux/flutter_linux_3.10.0-stable.tar.xz -o /opt/flutter.tar.xz \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -78,7 +78,7 @@ RUN : \
   && chmod +x /usr/local/bin/SymbolCollector.Console
 
 # https://docs.flutter.dev/get-started/install/linux#install-flutter-manually
-RUN curl -fsSL https://storage.googleapis.com/flutter_infra_release/releases/stable/linux/flutter_linux_3.10.0-stable.tar.xz -o /opt/flutter.tar.xz \
+RUN curl -fsSL https://storage.googleapis.com/flutter_infra_release/releases/stable/linux/flutter_linux_3.13.0-stable.tar.xz -o /opt/flutter.tar.xz \
```
asottile-sentry, line 81: we should probably be checksumming this file to ensure it is what we expected though nothing else does that here so shrugs

asottile-sentry (approved):
please in the future request reviews from `team-dev-infra` instead!

buenaflor:
> please in the future request reviews from team-dev-infra instead!

ah sry, thx for the quick approve though

PR #546, opened 2024-06-26
chore: update actions

vaind:
update actions due to deprecations and upcoming phase-out


asottile-sentry:
> upcoming phase-out

fwiw github has promised they will continue to run them in modern node versions despite the warnings -- so this is not really necessary beyond silencing a few warnings

vaind:
Didn't know about that pledge, thx for letting me know. Still doesn't seem to aply to artifact actions: https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/

PR #547, opened 2024-06-26
fix: linter issue

vaind:
[@typescript-eslint/no-non-null-assertion] Forbidden non-null assertion.

### DIFF
--- a/src/utils/__tests__/version.test.ts
+++ b/src/utils/__tests__/version.test.ts
@@ -185,8 +185,10 @@ describe('versionGreaterOrEqualThan', () => {
   test('can compare pre parts', () => {
     const v1 = parseVersion('1.2.3-1');
     const v2 = parseVersion('1.2.3-2');
-    expect(versionGreaterOrEqualThan(v1, v2)).toBe(false);
-    expect(versionGreaterOrEqualThan(v2, v1)).toBe(true);
+    expect(v1).not.toBeNull();
+    expect(v2).not.toBeNull();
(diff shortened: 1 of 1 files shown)


src/utils/__tests__/version.test.ts
```diff
@@ -183,10 +183,12 @@ describe('versionGreaterOrEqualThan', () => {
...
-    const v1 = parseVersion('1.2.3-1')!;
-    const v2 = parseVersion('1.2.3-2')!;
-    expect(versionGreaterOrEqualThan(v1, v2)).toBe(false);
-    expect(versionGreaterOrEqualThan(v2, v1)).toBe(true);
+    const v1 = parseVersion('1.2.3-1');
+    const v2 = parseVersion('1.2.3-2');
+    expect(v1).not.toBeNull();
+    expect(v2).not.toBeNull();
+    expect(versionGreaterOrEqualThan(v1 as SemVer, v2 as SemVer)).toBe(false);
+    expect(versionGreaterOrEqualThan(v2 as SemVer, v1 as SemVer)).toBe(true);
```
BYK, line 191:
```suggestion
    const v1 = parseVersion('1.2.3-1') as SemVer;
    const v2 = parseVersion('1.2.3-2') as SemVer;
    expect(versionGreaterOrEqualThan(v1, v2)).toBe(false);
    expect(versionGreaterOrEqualThan(v2, v1)).toBe(true);
```

BYK (changes requested):
I don't think this test needs to check for not null condition.

PR #549, opened 2024-07-23
ci: dependency review action

mdtro:
Add dependency review action. Any new dependencies with known high or critical vulnerabilities will fail CI.

### DIFF
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -1,0 +1,19 @@
+name: 'Dependency Review'
+on:
+  pull_request:
+    branches: ['master']
+
+permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/dependency-review.yml
```diff
@@ -0,0 +1,19 @@
+name: 'Dependency Review'
```
asottile-sentry, line 1: should we apply this globally at sentry rather than adding it to each repository separately?

mdtro: Yes, yes we should. :) 
I'll work with Jeffrey to get this added as one of the global workflows and applied to the right repositories.

PR #560, opened 2024-10-28
fix: Add optional to type chain

### LINKED ISSUE #559: Bug caused by missing PR Author
This bug occurs when trying to run a craft release workflow on a repo where PRs contain ghost users (deleted accounts). For example: https://github.com/getsentry/jsonnet-libs/pull/7

Normally, the release should run properly and should just skip & not include author names for the PRs which were opened by a deleted account. 

Instead, craft errors with the message `Error:  Cannot read properties of null (reading 'login')
  at /usr/local/bin/craft:399:261` due to the author fields missing from the graphql response and being null

brian-lou:
This PR fixes #559

### DIFF
--- a/src/utils/changelog.ts
+++ b/src/utils/changelog.ts
@@ -380,7 +380,7 @@ interface CommitInfo {
     nodes: Array<{
       number: string;
       body: string;
-      author: {
+      author?: {
         login: string;
       };
       milestone: {
@@ -472,7 +472,7 @@ async function getPRAndMilestoneFromCommit(
         hash.slice(1),
         pr
           ? {
-              author: pr.author?.login ?? null,
+              author: pr.author?.login,
               pr: pr.number,
               prBody: pr.body,
               milestone: pr.milestone?.number ?? null,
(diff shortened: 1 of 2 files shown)


src/utils/changelog.ts
```diff
@@ -472,7 +472,7 @@ async function getPRAndMilestoneFromCommit(
         hash.slice(1),
         pr
           ? {
-              author: pr.author.login,
+              author: pr.author?.login ?? null,
```
asottile-sentry, line 475: I don't think `?? null` is needed here with the `?.` access (since that'll become `undefined` already) -- I think this also needs a fix to the types above because as it is right now the `?.` looks unnecessary (I'm surprised the type checker isn't complaining about this?)

asottile-sentry (commented):
should probably also include a test -- here's the actual api body for the PR which may be helpful in crafting a fake graphql response for a test https://api.github.com/repos/getsentry/jsonnet-libs/pulls/5

asottile-sentry (approved):
please add the additional context to the issue as requested on slack -- looks good and thanks for the test! (helps to show that this makes a reasonable output too!)

PR #561, opened 2024-11-11
fix(utils): Ensure `isPreviewRelease` identifies `-dump` as a preview release

Lms24:
came up in: https://github.com/getsentry/sentry-docs/issues/11716

This PR adds `dump` to our pre-release detection regex, so that `isPreviewRelease()` check returns `true` for versions like `1.2.3-dump(.)1`. 

We had an internal discussion about choosing a more "wholistic" approach to detecting preview releases, like for example checking if there's any suffix to a typical semver `<major>.<minor>.<patch>` version. I'm also happy to implement this but I'm probably missing some context as to why we chose to opt in specific strings in the past. Perhaps the answer is that we can also handle non-semver version strings? Anyway, this should at least avoid the case brought up in https://github.com/getsentry/sentry-docs/issues/11716.

### DIFF
--- a/src/utils/__tests__/version.test.ts
+++ b/src/utils/__tests__/version.test.ts
@@ -120,12 +120,19 @@ describe('parseVersion', () => {
 });
 
 describe('isPreviewRelease', () => {
-  test('accepts semver preview release', () => {
-    expect(isPreviewRelease('2.3.4-preview1')).toBe(true);
-  });
+  test.each(['preview', 'pre', 'alpha.0', 'beta', 'rc.1', 'dev'])(
+    'accepts semver preview release',
+    previewSuffix => {
+      expect(isPreviewRelease(`2.3.4-${previewSuffix}1`)).toBe(true);
+    }
+  );
 
   test('accepts Python-style preview release', () => {
     expect(isPreviewRelease('2.3.4rc0')).toBe(true);
+  });
+
+  test('accepts dump preview release', () => {
+    expect(isPreviewRelease('2.3.4-dump1')).toBe(true);
   });
 
   test('does not accept non-preview release', () => {
--- a/src/utils/version.ts
+++ b/src/utils/version.ts
@@ -104,7 +104,7 @@ export function versionGreaterOrEqualThan(v1: SemVer, v2: SemVer): boolean {
 /**
  * A regular expression to detect that a version is a pre-release version.
  */
-export const PREVIEW_RELEASE_REGEX = /(?:[^a-z])(preview|pre|rc|dev|alpha|beta|unstable|a|b)(?:[^a-z]|$)/i;
+export const PREVIEW_RELEASE_REGEX = /(?:[^a-z])(preview|pre|rc|dev|alpha|beta|unstable|a|b|dump)(?:[^a-z]|$)/i;
 
 /**
  * Checks that the provided string is a pre-release version.


src/utils/__tests__/version.test.ts
```diff
@@ -120,14 +120,21 @@ describe('parseVersion', () => {
 });
 
 describe('isPreviewRelease', () => {
-  test('accepts semver preview release', () => {
-    expect(isPreviewRelease('2.3.4-preview1')).toBe(true);
-  });
+  test.each(['preview', 'pre', 'alpha.0', 'beta', 'rc.1', 'dev'])(
```
Lms24, line 123: added some more tests here for other suffixes

src/utils/version.ts
```diff
@@ -104,7 +104,7 @@ export function versionGreaterOrEqualThan(v1: SemVer, v2: SemVer): boolean {
 /**
  * A regular expression to detect that a version is a pre-release version.
  */
-export const PREVIEW_RELEASE_REGEX = /(?:[^a-z])(preview|pre|rc|dev|alpha|beta|unstable|a|b)(?:[^a-z]|$)/i;
+export const PREVIEW_RELEASE_REGEX = /(?:[^a-z])(preview|pre|rc|dev|alpha|beta|unstable|a|b|dump)(?:[^a-z]|$)/i;
```
asottile-sentry, line 107: I've never heard of "dump" as a prerelease naming scheme -- is this standardized anywhere? I'd prefer we use actual semver names rather than making up our own here (as that would be more expected from users as well when reasoning about versions)

Lms24: As far as I understand, `dump` was chosen because it described a feature that was added to this preview release. So likely not standardized. I agree that this is a niche/edge case but "dump" kinda seems likely enough to come up again. Also noticed while [removing the entries](https://github.com/getsentry/sentry-release-registry/pull/169) of this version from the registry, we had similar naming schemes being used in the past.

I can also close this as I get your point. WDYT about checking for any suffix after the semver patch number instead?

asottile-sentry: ok then we definitely shouldn't add this to the regex

I think craft should *reject* any non-semver versions -- I thought it already did this but maybe that's only for python?

the dotnet sdk release should have named it -pre1 or one of the standard naming schemes

asottile-sentry: one important case either way is `1.2.3-1` is standardized [in python](https://packaging.python.org/en/latest/specifications/version-specifiers/#version-specifiers) as equivalent to `1.2.3.post1` and [we use that intentionally](https://github.com/getsentry/sentry-forked-django-stubs/releases) so we have to make sure that still works

Lms24: I checked the `1.2.3-1` case:

- `isPreviewRelease` returns `false`. I guess this is correct, given this is a post release? 
- `parseVersion` can parse this string and identifies the `-1` as `.pre`:
  ```js
   test('parses a Python-style post release version', () => {
     expect(parseVersion('1.2.3-1')).toEqual({
       major: 1,
       minor: 2,
       patch: 3,
       pre: '1',
     });
   });
   ```

Lms24:
Closing as `dump` is not a standardized pre-release identifier. We can think about a more versatile check for arbitrary identifiers (in a new PR) or we stick to using well-established identifiers.

src/utils/version.ts
asottile-sentry: seems fine to me!

Lms24: Ok now the specific list of pre-releases makes sense: If we want to ensure Python post releases are not identified as pre-releases, we can't check against arbitrary pre-release version identifiers. So just checking for `!!parseVersion(...).pre` would not cover the python post release correctly. 

So I guess, we don't change anything 👍 I can open a PR with some tests I wrote to have the Python case covered.

PR #562, opened 2024-11-11
test(utils): Add additional tests for version helper functions

Lms24:
This PR adds some tests for version helper functions used in multiple craft targets.

This came up after investigating https://github.com/getsentry/sentry-docs/issues/11716 where I found out that we have a specific list of pre-release suffix identifiers that we consider a pre-release, instead of considering every kind of `-x` suffix as a pre-release. 

Specifically, as mentioned in https://github.com/getsentry/craft/pull/561#discussion_r1836934373,  we don't consider the (shorthand) python post release pattern (TIL) a pre-release, although `parseVersion` actually puts it into the returned `.pre` property. 

Tests added:
- `parseVersion` parses Python post release values
- `isValidVersion` returns `true` for Python post releases
- `isPreviewVersion` returns `false` for Python post releases
- More tests for `isPreviewVersion` returning `true` for allowed identifiers  

I don't have strong feelings on getting this merged, so feel free to reject it but I figured I might as well open a PR after having added these tests locally.

### DIFF
--- a/src/utils/__tests__/version.test.ts
+++ b/src/utils/__tests__/version.test.ts
@@ -42,6 +42,10 @@ describe('isValidVersion', () => {
     expect(isValidVersion('1.2.3rc1')).toBe(true);
   });
 
+  test('accepts valid Python-style post release version', () => {
+    expect(isValidVersion('1.2.3-1')).toBe(true);
+  });
+
   test('does not accept leading "v"', () => {
     expect(isValidVersion('v1.2.3')).toBe(false);
   });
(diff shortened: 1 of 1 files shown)


src/utils/__tests__/version.test.ts
```diff
@@ -110,6 +114,15 @@ describe('parseVersion', () => {
     });
   });
 
+  test('parses a Python-style post release version', () => {
+    expect(parseVersion('1.2.3-1')).toEqual({
+      major: 1,
+      minor: 2,
+      patch: 3,
+      pre: '1',
```
asottile-sentry, line 122: might be worth a comment that it misinterprets this -- but it's ok as long as we don't utilize this function for python in the same way

asottile-sentry (approved):
fixnship

PR #567, opened 2024-11-22
fix(github): Guard against missing release

mydea:
Fixes https://github.com/getsentry/craft/pull/503#discussion_r1852239403

Note: We already have a test covering that `getLatestRelease()` accepts undefined.

### DIFF
--- a/src/targets/github.ts
+++ b/src/targets/github.ts
@@ -127,7 +127,7 @@ export class GitHubTarget extends BaseTarget {
       };
     }
 
-    let latestRelease: { tag_name: string } | undefined = undefined;
+    let latestRelease: { tag_name: string } | undefined;
     try {
       latestRelease = (
         await this.github.repos.getLatestRelease({
--- a/src/targets/github.ts
+++ b/src/targets/github.ts
@@ -135,8 +135,12 @@ export class GitHubTarget extends BaseTarget {
           repo: this.githubConfig.repo,
         })
       ).data;
-    } catch {
-      // no release yet
+    } catch (error) {
+      // if the error is a 404 error, it means that no release exists yet
+      // all other errors should be rethrown
+      if (error.status !== 404) {
+        throw error;
+      }
     }
 
     const isLatest = isPreview
--- a/src/targets/github.ts
+++ b/src/targets/github.ts
@@ -127,10 +127,17 @@ export class GitHubTarget extends BaseTarget {
       };
     }
 
-    const { data: latestRelease } = await this.github.repos.getLatestRelease({
-      owner: this.githubConfig.owner,
-      repo: this.githubConfig.repo,
-    });
+    let latestRelease: { tag_name: string } | undefined = undefined;
+    try {
+      latestRelease = (
+        await this.github.repos.getLatestRelease({
+          owner: this.githubConfig.owner,
+          repo: this.githubConfig.repo,
+        })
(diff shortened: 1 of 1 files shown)


src/targets/github.ts
```diff
@@ -127,10 +127,17 @@ export class GitHubTarget extends BaseTarget {
...
-    });
+    let latestRelease: { tag_name: string } | undefined = undefined;
+    try {
+      latestRelease = (
+        await this.github.repos.getLatestRelease({
+          owner: this.githubConfig.owner,
+          repo: this.githubConfig.repo,
+        })
+      ).data;
+    } catch {
```
BYK, line 138: The blanket `catch` here may mask other potential issues such as authentication or temporary server unavailability. I feel like we should filter these.

mydea: what would you filter here? To me it sounds reasonable to fall back to assume it is the latest release if something goes wrong here 🤔 But if you want to exclude something specific, we can of course do that, open for concrete suggestions (not really easy to test/try this...)

BYK: We should not assume lack of a release on a 401 or 500 response. I'd even say unless we get a 200 with an explicit null in it, we should not assume anything?

mydea: So you mean we should fail the whole release if we do not get a 200 here?

BYK: Yes as otherwise you create an incomplete release or worse, mark a hotfix release as latest accidentally.

If you think this behavior should be optional, we should revise the flow and make this an explicit user choice or make a breaking release.

BYK (changes requested):
See https://github.com/getsentry/craft/pull/567/files#r1899608188

In it's current form this patch creates inconsistent behavior on server related issues.

src/targets/github.ts
mydea: finally got around to give this another go. I adjusted it to only "swallow" 404 errors, which occur when no release exists. all other errors are rethrown!

src/targets/github.ts
```diff
@@ -127,10 +127,21 @@ export class GitHubTarget extends BaseTarget {
...
+      latestRelease = (
+        await this.github.repos.getLatestRelease({
+          owner: this.githubConfig.owner,
+          repo: this.githubConfig.repo,
+        })
+      ).data;
+    } catch (error) {
+      // if the error is a 404 error, it means that no release exists yet
+      // all other errors should be rethrown
+      if (error.status !== 404) {
```
mydea, line 141: @BYK adjusted this here to only handle 404 errors!

BYK: We still risk getting a 404 on bad credentials that said that would be so rare that we can keep this safely (rare because these repos are meant to be public at all times)

src/targets/github.ts
```diff
@@ -127,10 +127,21 @@ export class GitHubTarget extends BaseTarget {
       };
     }
 
-    const { data: latestRelease } = await this.github.repos.getLatestRelease({
-      owner: this.githubConfig.owner,
-      repo: this.githubConfig.repo,
-    });
+    let latestRelease: { tag_name: string } | undefined = undefined;
```
BYK, line 130:
```suggestion
    let latestRelease: { tag_name: string } | undefined;
```

BYK (approved):
🚀

PR #568, opened 2024-11-22
Downgrade CocoaPods from 1.16.2 to 1.14.2

philipphofmann:
Rollback to 1.14.2 because 1.16.2 isn't working.


philipphofmann:
Closed in favor of https://github.com/getsentry/craft/pull/569.

PR #570, opened 2024-11-22
chore: Comment why we can't use cocoapods 1.16.2

philipphofmann:
Explain why we can't use cocoapods 1.16.2 and link a related GH issue.

Follow up on https://github.com/getsentry/craft/pull/569.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,10 @@
 # Pin CocoaPods Version to avoid that bugs in CocoaPods like
 # https://github.com/CocoaPods/CocoaPods/issues/12081 break our release
 # workflow.
-# 1.16.2 has a bug that breaks the release workflow
+# 1.16.2 has a bug that breaks the release workflow. Publishing with craft then
+# fails with:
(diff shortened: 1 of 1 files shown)


kahest (approved):

Gemfile
```diff
@@ -2,6 +2,10 @@
 # Pin CocoaPods Version to avoid that bugs in CocoaPods like
 # https://github.com/CocoaPods/CocoaPods/issues/12081 break our release
 # workflow.
-# 1.16.2 has a bug that breaks the release workflow
+# 1.16.2 has a bug that breaks the release workflow. Publishing with craft then
+# fails with:
+# - ERROR | [iOS] unknown: Encountered an unknown error (Unable to locate the
+#   executable `rsync`) during validation.
```
asottile-sentry, line 8: reading the error message -- do you not just want to install rsync?

asottile-sentry: yes that would be the place -- you can build the docker image locally and play around with it too

philipphofmann: I will do that before updating CocoaPods. Thanks.

PR #571, opened 2024-11-24
feat: dotnet publish re-entrant

bruno-garcia:
Use `--skip-duplicate` to allow the job to be retried

[From docs](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-nuget-push):

> When pushing multiple packages to an HTTP(S) server, treats any 409 Conflict response as a warning so that other pushes can continue.

### DIFF
--- a/src/targets/nuget.ts
+++ b/src/targets/nuget.ts
@@ -73,6 +73,7 @@ export class NugetTarget extends BaseTarget {
       '--api-key',
       '${NUGET_API_TOKEN}',
       '--source',
+      '--skip-duplicate',
       this.nugetConfig.serverUrl,
     ]);
   }


bruno-garcia:
Got it. I'm this case we end up with half the packages release and half not. Which is a pain to fix by hand. With this change we'll be able to finish the step at least.

BYK:
@bruno-garcia when you check off a target in `publish`, it is skipped. Is this multiple packages in a single target? If so, this may be grounds for granting an exception maybe?

bruno-garcia:
> @bruno-garcia when you check off a target in `publish`, it is skipped. Is this multiple packages in a single target? If so, this may be grounds for granting an exception maybe?

It's a single target. It's just 1 command `dotnet publish` and it does the job on all packages.
The command is made re-entrant with the flag I added in this PR. 

We don't really gain anything by not having that flag in, right? But definitely would be helpful having the flag in.

cleptric (approved):

BYK:
>either fails if it's already there, or just ignores it (the behavior I need with this PR)

Just a heads up, I think we should clearly document this as the "just ignores it" is the behavior we explicitly did not want in Craft. Given your circumstances, I think this case deserves the exception but we should clearly document that this is a departure from the norm.

PS: We can also revisit the norm, just saying we should note the odd one out.

PR #585, opened 2025-02-10
fix:(GitHub Auth): Deprecate getAuthUsername

Jeffreyhung:
Currently we are using `https://username:token@github/owner/repo` to perform auth will talking to GitHub, it does not work well with GitHub Apps, hence deprecating it and bake the authentication token into authentication headers.

### DIFF
--- a/src/utils/githubApi.ts
+++ b/src/utils/githubApi.ts
@@ -12,10 +12,6 @@ export class GitHubRemote {
   public readonly owner: string;
   /** GitHub repository name */
   public readonly repo: string;
-  /** GitHub username */
-  protected username?: string;
-  /** GitHub personal authentication token */
-  protected apiToken?: string;
   /** GitHub hostname */
   protected readonly GITHUB_HOSTNAME: string = 'github.com';
   /** Protocol prefix */
@@ -26,8 +22,6 @@ export class GitHubRemote {
   public constructor(
     owner: string,
     repo: string,
-    username?: string,
-    apiToken?: string
   ) {
     this.owner = owner;
     this.repo = repo;
--- a/src/utils/githubApi.ts
+++ b/src/utils/githubApi.ts
@@ -45,7 +45,7 @@ export class GitHubRemote {
   /**
    * Returns an HTTP-based git remote with embedded HTTP basic auth
    *
-   * Using dummy username as it does not matter for cloning
+   * Using placeholder username as it does not matter for cloning
    *
    * It MAY contain sensitive information (e.g. API tokens)
    */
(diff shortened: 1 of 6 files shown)


src/utils/githubApi.ts
```diff
@@ -31,22 +31,9 @@ export class GitHubRemote {
   ) {
     this.owner = owner;
     this.repo = repo;
-    if (username && apiToken) {
-      this.setAuth(username, apiToken);
-    }
```
asottile-sentry, line 36: should be able to remove these parameters entirely -- or maybe take the api key and do the header modification here?

src/utils/githubApi.ts
```diff
@@ -87,6 +55,20 @@ export function getGitHubApiToken(): string {
   return githubApiToken;
 }
 
+/**
+ * Returns the GitHub auth header
+ *
+ * @returns GitHub auth header
+ */
+export function getGitHubAuthHeader(): Array<string> {
```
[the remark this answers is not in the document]
asottile-sentry: afaict both ways work with user tokens or app tokens -- fun fact the username does not matter for the clone-url approach just that it cannot be empty

asottile-sentry (approved):

src/utils/githubApi.ts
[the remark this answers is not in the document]
BYK: Wowza, alright then let's go with the old way as it seems simpler?

Jeffreyhung:
Did some testing locally and figured the auth header won't work as expected.
Switching back to original method with a dummy username as it doesn't matter for clone (🥔 @asottile-sentry )

src/utils/githubApi.ts
```diff
@@ -60,12 +45,14 @@ export class GitHubRemote {
...
+   * Using dummy username as it does not matter for cloning
+   *
    * It MAY contain sensitive information (e.g. API tokens)
    */
   public getRemoteStringWithAuth(): string {
     const authData =
-      this.username && this.apiToken
-        ? `${this.username}:${this.apiToken}@`
+      this.apiToken
+        ? `dummyusername:${this.apiToken}@`
```
asottile-sentry, line 55: placeholderusername is more appropriate from an ableist-language perspective

Jeffreyhung: Updated!

src/utils/githubApi.ts
```diff
@@ -60,12 +45,14 @@ export class GitHubRemote {
   /**
    * Returns an HTTP-based git remote with embedded HTTP basic auth
    *
+   * Using dummy username as it does not matter for cloning
```
asottile-sentry, line 48: here too

asottile-sentry (approved):

PR #592, opened 2025-03-10
feat: Install most recent Erlang from RabbitMQ packages

BYK:
This should fix the Elixir release issues: https://github.com/getsentry/publish/actions/runs/13697376301/job/38302799587#step:12:94

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:22-bookworm-slim as builder
+FROM node:22-bookworm-slim AS builder
 
 WORKDIR /usr/local/lib
 
@@ -27,20 +27,20 @@ ENV DEBIAN_FRONTEND=noninteractive \
 
 RUN apt-get -qq update \
   && apt-get install -y --no-install-recommends \
-    apt-transport-https \
-    build-essential \
-    curl \
-    default-jdk-headless \
-    dirmngr \
-    gnupg \
-    git \
-    jq \
-    python3-packaging \
-    python3-venv \
-    rsync \
-    ruby-full \
-    unzip \
-    maven \
+  apt-transport-https \
+  build-essential \
+  curl \
+  default-jdk-headless \
+  dirmngr \
+  gnupg \
+  git \
+  jq \
+  python3-packaging \
+  python3-venv \
+  rsync \
+  ruby-full \
+  unzip \
+  maven \
   && apt-get clean \
   && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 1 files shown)


asottile-sentry:
this does not seem like a good idea -- why not just upgrade instead of using some unofficial build

BYK:
@asottile-sentry would love to but the versions on the Debian repos are too old for our stuff.

asottile-sentry:
> @asottile-sentry would love to but the versions on the Debian repos are too old for our stuff.

? on the link in the deleted comment I see 27.3 available which is the absolute latest

whatyouhide (approved):
Nice!!

BYK:
@asottile-sentry I did check that before. The problem is, they do not publish the package listed there (hence our failures). We can try to download and install the `.deb` file manually but that felt like a less stable solution, especially after seeing this: https://github.com/esl/packages/issues/70

I thin the RabbitMQ packages are used and relied on enough to make it here but if you disagree, I am okay to try downloading and installing the `.deb` file from there.

whatyouhide:
The hex.pm team maintains [one the most widely-used Docker images](https://hub.docker.com/r/hexpm/elixir) for Elixir development. It supports Debian—[here](https://hub.docker.com/layers/hexpm/elixir/1.18.3-erlang-27.3-debian-buster-20240612-slim/images/sha256-ba46873c7794839ec879edaabcef09613ef509571ac08c3b291591cfc4a9f24b) is an example of the latest Erlang/Elixir running on Debian.

sl0thentr0py:
@whatyouhide the problem is we have a [single container based on `node:22-bookworm-slim`](https://github.com/getsentry/craft/blob/master/Dockerfile) where we add all the language specific tooling later.

Is this the correct Dockerfile that build the images you linked? Here they're building from source.
https://github.com/hexpm/bob/blob/main/priv/scripts/docker/erlang-debian-bookworm.dockerfile

I discussed having separate images per publishing tool, but I don't think anyone has time/motivation/investment justification to make such a large change to craft.

The larger problem remains that maintaining this Dockerfile is a PITA and we break the release flow everytime someone attempts to do so.

whatyouhide:
@sl0thentr0py yes that's the Dockerfile for Erlang, and then Elixir gets installed on top. You could also pin the Erlang version if it's just for releasing. You could also do asdf on the particular Elixir image you build yeah.

BYK:
@sl0thentr0py I think I may help with that improvement (not using an omnibus docker image for everything)

It'll take a few days at least though. What's the best way to unblock right now?

sl0thentr0py:
current fix is okay then for me! will leave it to @asottile-sentry to approve

sl0thentr0py:
people are waiting on the release, could we get this out? @asottile-sentry

asottile-sentry:
> people are waiting on the release, could we get this out? @asottile-sentry

I'm not blocking anything -- it just seems like a bad idea to use rabbitmq's builds rather than official ones

whatyouhide:
If @BYK is going to work on making Docker images configurable, this seems like a good patch to get releases out in the meantime?

BYK:
To make it clear here: I did search for official builds for quite a while and that's how I ended up with the RabbitMQ builds. If anyone has a better idea, happy to go with that.

@whatyouhide @sl0thentr0py - What's the main issue with the omnibus Debian image? What would you have used if it was not this?

sl0thentr0py:
just the hexpm/elixir base image directly

BYK:
@sl0thentr0py would love to replicate what they do in our own image but I can't wrap my head around it (probably because Elixir is alien to me). Any pointers? Bob the builder doesn't really help 😅

BYK:
We can definitely do that. I checked their Dockerfile and looks like that's what they do?

whatyouhide:
@sl0thentr0py @BYK releases are still blocked for the Elixir SDK. Anything I can do to help?

sl0thentr0py (approved):

sl0thentr0py:
i approved let's just try this

whatyouhide:
@sl0thentr0py any news here?

sl0thentr0py:
we just need to approve the release again, but Stephanie is out so I'll ping some other manager

PR #597, opened 2025-04-23
pin dotnet sdk 9 version to 9.0.203

bruno-garcia:
Tried to verify this locally with docker on with the same base image and not sure why but failed through several steps.
Just using CI as crutch to see if this works.


bruno-garcia:
> 4.136 E: Version '9.0.203' for 'dotnet-sdk-9.0' was not found

can't pin the version. 

New plan: We'll change the nuget target to get rid of any `global.json` when running `dotnet nuget push` since we don't care about the version of the SDK at that time. We do when building/packaging only.

PR #601, opened 2025-05-01
fix: Avoid global.json when running `dotnet push`

### LINKED ISSUE #598: Temporarily disable global.json when pushing NuGet packages
# Problem description

We recently [pinned the version of dotnet](https://github.com/getsentry/sentry-dotnet/blob/d86fe28d931aed859abc9bf95fa6888b6181ecaa/global.json#L4) that we use in the sentry-dotnet repo. This ensures that when we're building the solution in local development environments we get the same versions of workloads (for things like MAUI) that are used when building the solution in CI. Without this, .NET will roll forward to the latest available workload versions, which can change the behaviour of the resulting application (or even break the builds).

However this may not be the same version of .NET that is installed on the docker containers used by Craft when publishing NuGet packages. 

 _Originally posted by @github-actions in [#5421](https://github.com/getsentry/publish/issues/5421#issuecomment-2824998861)_

> dotnet:       A compatible .NET SDK was not found.
> dotnet: 
> dotnet: Requested SDK version: 9.0.203
> dotnet: global.json file: /github/workspace/__repo__/global.json

This is because when the craft workflow downloads the build artefacts it extracts both the NuGet packages that need publishing and the global.json file that we use for development. So any `dotnet` commands that Craft tries to execute, at that point, are bound by the constraints specified in the global.json file.

# Possible solution

Ultimately craft is trying to [publish the (already built) NuGet packages to NuGet ](https://github.com/getsentry/craft/blob/2f923ebd89154dce7db9c87ca475a2b25b1babdf/src/targets/nuget.ts#L62-L7)and the specific version of dotnet that is used for this operation is unimportant.

We could temporarily disable the global.json file (if present) by renaming it prior to this operation and then re-enable it again afterwards... so something like:

```diff
  public async uploadAsset(path: string): Promise<any> {
+   mv global.json global.json.disabled
-   return spawnProcess(NUGET_DOTNET_BIN, [
+   var result = spawnProcess(NUGET_DOTNET_BIN, [
      'nuget',
      'push',
      path,
      '--api-key',
      '${NUGET_API_TOKEN}',
      '--source',
      this.nugetConfig.serverUrl,
    ]);
+  mv global.json.disabled global.json
+  return result
  }
```

That logic would have to be converted into valid typescript code, of course, but you get the idea.

There doesn't seem to be any other way of [disabling global.json when executing a specific command in .NET](https://github.com/dotnet/sdk/issues/10311#issuecomment-1254912466).

jamescrosswell:
Resolves #598:
- https://github.com/getsentry/craft/issues/598

Pinning the version of of the .NET SDK used to build the sentry-dotnet repo prevents us from making relases, since that version of .NET is not installed on the machine that publishes NuGet packages.

We can run `dotnet push` using any version of .NET so this PR simply runs that command from outside the repository working directory.

### DIFF
--- a/src/targets/nuget.ts
+++ b/src/targets/nuget.ts
@@ -66,7 +66,7 @@ export class NugetTarget extends BaseTarget {
    * @returns A promise that resolves when the upload has completed
    */
   public async uploadAsset(path: string): Promise<any> {
-    return spawnProcess(NUGET_DOTNET_BIN, [
+    const args = [
       'nuget',
       'push',
       path,
@@ -74,7 +74,11 @@ export class NugetTarget extends BaseTarget {
       '${NUGET_API_TOKEN}',
       '--source',
       this.nugetConfig.serverUrl,
-    ]);
+    ];
+    // Run this outside the repository folder to avoid global.json constraints
+    // (we don't need specific dotnet/workload versions to upload to nuget)
+    const spawnOptions = { cwd: '/tmp/' };
+    return spawnProcess(NUGET_DOTNET_BIN, args, spawnOptions);
   }
 
   /**
--- a/src/targets/nuget.ts
+++ b/src/targets/nuget.ts
@@ -75,9 +75,9 @@ export class NugetTarget extends BaseTarget {
       '--source',
       this.nugetConfig.serverUrl,
     ];
-    // Run this outside the repository folder to avoid global.json constraints
-    // (we don't need specific dotnet/workload versions to upload to nuget)
-    const spawnOptions = { cwd: '/tmp/' };
+    // Run outside the repository folder to avoid global.json constraints
+    // (we don't need specific dotnet/workload versions just to upload to nuget)
+    const spawnOptions = { cwd: '/' };
     return spawnProcess(NUGET_DOTNET_BIN, args, spawnOptions);
   }
 


src/targets/nuget.ts
jamescrosswell, line 1: @asottile-sentry any tips on how to test this? As far as I can tell, this would only ever be run when we try to make a release from the `sentry-dotnet` repo. 

`spawnProcess` is well and truly battle tested. `uploadAsset` is simply a wrapper for this (passing specific arguments). There's no real logic to test. Ultimately, it "passes" the test if those arguments result in a real NuGet package being uploaded to nuget.org... I can't think of any meaningful tests other than that.

src/targets/nuget.ts
```diff
@@ -66,15 +66,19 @@ export class NugetTarget extends BaseTarget {
...
       path,
       '--api-key',
       '${NUGET_API_TOKEN}',
       '--source',
       this.nugetConfig.serverUrl,
-    ]);
+    ];
+    // Run this outside the repository folder to avoid global.json constraints
+    // (we don't need specific dotnet/workload versions to upload to nuget)
+    const spawnOptions = { cwd: '/tmp/' };
```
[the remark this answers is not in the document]
asottile-sentry: I would use `/` or `..` instead personally

src/targets/nuget.ts
asottile-sentry: yeah there's not really much to test without copying the implementation details into the test

src/targets/nuget.ts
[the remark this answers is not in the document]
jamescrosswell: k, I've just gone with the root directory...

src/targets/nuget.ts
jamescrosswell: In that case, I think this commit is ready. We can make a new release from the [sentry-dotnet](https://github.com/getsentry/sentry-dotnet) repository as soon as it's merged to make sure it all hangs together.

asottile-sentry (approved):

jamescrosswell:
@asottile-sentry I suspect you might need to merge this (I don't think I have permission... I certainly don't have any merge button).

PR #602, opened 2025-05-02
fix: Don't fail when there are no releases

### LINKED ISSUE #342: Prepare dies if there are no tags yet - fatal: No names found, cannot describe anything
### Steps to Reproduce

1. Start a new repo.
1. Commit a simple `.craft.yml` file.
1. Run `craft prepare 0.0.0`.

### Expected Result

Successful completion.

### Actual Result

```
$ craft prepare 0.0.0
ℹ Checking the local repository status...                                                                            11:45:29
ℹ Releasing version 0.0.0 from main                                                                                  11:45:29
ℹ Preparing to release the version: 0.0.0                                                                            11:45:29
ℹ Created a new release branch: "release/0.0.0"                                                                      11:45:29
ℹ Switched to branch "release/0.0.0"                                                                                 11:45:29
here

 ERROR  fatal: No names found, cannot describe anything.                                                             11:45:29


  
  at Object.action (/Users/chadwhitacre/workbench/getsentry/craft/dist/craft:154:11690)
  at Bae.exec (/Users/chadwhitacre/workbench/getsentry/craft/dist/craft:154:12163)
  at /Users/chadwhitacre/workbench/getsentry/craft/dist/craft:154:18969
  at new Promise (<anonymous>)
  at Xae.handleTaskData (/Users/chadwhitacre/workbench/getsentry/craft/dist/craft:154:18858)
  at Xae.<anonymous> (/Users/chadwhitacre/workbench/getsentry/craft/dist/craft:154:18435)
  at Generator.next (<anonymous>)
  at o (/Users/chadwhitacre/workbench/getsentry/craft/dist/craft:154:16995)
  at processTicksAndRejections (internal/process/task_queues.js:95:5)

$
```

Here's the call that fails:

https://github.com/getsentry/craft/blob/2ff325e4d0ac9d4a8410b8107cede9144c2b91b6/src/utils/git.ts#L35-L38

See:

```
$ git describe --tags --abbrev=0
fatal: No names found, cannot describe anything.
$
```

BYK:
Fixes #342.

### DIFF
--- a/src/utils/git.ts
+++ b/src/utils/git.ts
@@ -43,7 +43,9 @@ export async function getLatestTag(git: SimpleGit): Promise<string> {
     // If there are no tags, return an empty string
     if (
       err instanceof Error &&
-      err.message.startsWith('fatal: No names found')
+      (
+        err.message.startsWith('fatal: No names found') ||
+        err.message.startsWith('Nothing to describe'))
     ) {
       return '';
     }
--- a/src/utils/git.ts
+++ b/src/utils/git.ts
@@ -1,4 +1,4 @@
-import simpleGit, { SimpleGit } from 'simple-git';
+import simpleGit, { type SimpleGit, type LogOptions, type Options } from 'simple-git';
 
 import { getConfigFileDir } from '../config';
 import { ConfigurationError } from './errors';
@@ -20,6 +20,8 @@ export interface GitChange {
 // https://docs.github.com/en/rest/commits/commits#list-pull-requests-associated-with-a-commit
 export const PRExtractor = /(?<=\(#)\d+(?=\)$)/;
 
+export const defaultInitialTag = '0.0.0';
+
 export async function getDefaultBranch(
   git: SimpleGit,
   remoteName: string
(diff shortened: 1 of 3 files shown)


asottile-sentry:
I was pretty sure the conclusion from #342 was that this wasn't really fixable for two reasons:
- the bump script has two required positional arguments (old version, new version) and passing nonsense is a breaking change
- the changelog generation requires a previous version

BYK:
It is fixable (at least to me), here's my reasoning:

>the bump script has two required positional arguments (old version, new version) and passing nonsense is a breaking change

Not necessarily as this _only_ applies to the very first release. At that point one can tweak the script. In any case, I use `0.0.0` as the default version instead of an empty string just in case. That said the "breaking" part doesn't really apply as this would not affect any existing project (otherwise you are right, it is a marked depart from the original behavior, which is crashing).

>the changelog generation requires a previous version

The workaround for this is just getting all the commits up to `HEAD` which just works?

asottile-sentry:
I'd rather we fix this properly and deprecate the 2-arg release bump script and make a bump-version-v2.sh or whatever

> The workaround for this is just getting all the commits up to HEAD which just works?

I assume you mean to the beginning of the history? `HEAD` is the current revision

BYK:
>I assume you mean to the beginning of the history? HEAD is the current revision

Up to `HEAD` from beginning. The entire commit history in short.

BYK:
>I'd rather we fix this properly and deprecate the 2-arg release bump script and make a bump-version-v2.sh or whatever

Why though? Having the old version available is definitely useful and making the script be called with 1 arg for the initial release and 2 args for the rest seems like adding unnecessary complexity.

What's your main concern here? I am lax as this only applies to new projects and shouldn't have any impact on existing ones.

asottile-sentry:
I don't think the old version is useful.   and I don't think sending a bogus value for the first release makes sense either

BYK:
@asottile-sentry I cannot justify the change you are asking for from a cost benefit standpoint. Making a deprecation for all existing projects, forcing an analysis and migration with a way to manage a breaking change for a fix that only affects the first release of new projects doesn't seem cost effective to me.

What am I missing from your perspective?

src/utils/git.ts
```diff
@@ -57,7 +69,12 @@ export async function getChangesSince(
     // this should still return all commits for individual repos when run from
     // the repo root.
     file: '.',
-  });
+  };
+
+  if (rev) {
+			gitLogArgs.from = rev;
```
hubertdeng123: nit: weird tab/spacing here

hubertdeng123 (commented):
> I'd rather we fix this properly and deprecate the 2-arg release bump script and make a bump-version-v2.sh or whatever

I agree with this, but it feels like despite the logic here being pretty hacky, it's a better experience for the end user who we're building for and nobody should release a 0.0.0 version. Anyone who is setting up craft for their release pipeline will encounter this bug. We can opt for a more in extensive rework in the future

WDYT @asottile-sentry?

PR #603, opened 2025-06-11
fix(aws): make reading/writing of aws manifest consistent with other packages

sl0thentr0py:
see the corresponding implementation here for all other packages:
https://github.com/getsentry/craft/blob/f4949f4b275e495e949bfd88e50125f284680446/src/utils/registry.ts#L41-L43
https://github.com/getsentry/craft/blob/f4949f4b275e495e949bfd88e50125f284680446/src/utils/registry.ts#L62

mainly the newline at the end was missing which was causing the aws endpoint failures I think.


BYK (approved):
This should probably be extracted out as a helper to help with consistency?

sl0thentr0py:
the helpers already exist in utils.. I don't know why whoever wrote the aws stuff didn't use them.
Don't really have time for a refactor here right now, sorry!

PR #604, opened 2025-06-27
gha: pin action versions to sha and set permissions

mdtro:
- Pin the GitHub Actions used to avoid dependency attacks.
- Set appropriate `permissions` scope for action steps.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,16 +9,18 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
     strategy:
       matrix:
         node: ['20','22']
     name: Node ${{ matrix.node }}
     steps:
-      - uses: actions/checkout@v3
-      - uses: actions/setup-node@v3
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -9,16 +9,18 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
```
asottile-sentry, line 13: this doesn't actually improve posture and is one of those annoying things security vendors always try and push.  if we actually want to improve posture here we should be setting the defaults org wide instead of sprinkling it into every workflow

mdtro: This improves it from our default posture, which is read/write.

I agree though, I would like to set this globally at the organization level, but there would likely be a lot of failed workflows as a result. I'm working my way through the findings and critical repos to get these specific clauses added to mitigate that before flipping the default to `contents: read & packages: read`. 

image

asottile-sentry: so let's leave it off where it would make no improvement otherwise it's just more noise

PR #608, opened 2025-07-02
feat(maven): Add Central repository support

romtsn:
Sonatype changed their APIs, so this PR adapts to the new endpoints:
- Change how we retrieve the repo_id by parsing the key
- Change API endpoint to get the repository status
- Change the body for the requests that close/promote repositories
- Add polling of the new endpoint to get the status of deployment with the deployment_id 

More details in this internal doc: https://www.notion.so/sentry/Nexus-Central-migration-2238b10e4b5d80368abbd74dd0b5ca46

### DIFF
--- a/src/targets/__tests__/maven.test.ts
+++ b/src/targets/__tests__/maven.test.ts
@@ -8,6 +8,7 @@ import {
   POM_DEFAULT_FILENAME,
   targetOptions,
   targetSecrets,
+  CENTRAL_API_BASE_URL,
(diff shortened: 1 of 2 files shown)


buenaflor (commented):
do we need to add support for [autoPublish](https://central.sonatype.org/publish/publish-portal-maven/#autopublish)?

afaik in sonatype when packages are released by default they transition into a validated state and then need to be manually deployed

PR #611, opened 2025-07-08
feat(maven): Skip already published artifacts

romtsn:
We're hitting a new endpoint now to check whether a certain artifact has already been published or not, e.g.:
```
https://central.sonatype.com/api/v1/publisher/published?namespace=io.sentry&name=sentry&version=8.17.0
```

This will help when we cancel releases half-way, so craft can pick it up from where it was aborted and continue to a successful release. This will result in the following output in craft:
image

### DIFF
--- a/src/targets/maven.ts
+++ b/src/targets/maven.ts
@@ -188,7 +188,7 @@ export class MavenTarget extends BaseTarget {
     if (!this.config.android) {
       throw new ConfigurationError(
         'Required Android configuration was not found in the configuration file. ' +
-        'See the documentation for more details'
+          'See the documentation for more details'
       );
     }
 


BYK:
My main concern here would be about how this affects the built-in protection and assumption that Craft simply doesn't allow repeat publishing for the same version. This behavior has been explicitly built-in and carried over years: if the version you are trying to publish already exists Craft fails and refuses, asking you to create a new release.

So my questions is: does this PR change that dynamic for Maven because it looks like it does to me.

asottile-sentry (changes requested):
craft is intentionally not idempotent for uploading artifacts.

the recommendation is to make a new release instead

romtsn:
> @romtsn I do undestand the issue. There are two important questions we need to answer though:
> 
> 1. Why was the release done half-way. Why was it interrupted?
> 2. Why can we not deprecate the broken release and cut a new one?

1. We wanted to include one more PR into the release and added "#retract" but it was too late
2. I wouldn't see it as broken, but rather "unfinished". Basically it was successful with maven and all we need now is to make sure it's also published to github and release-registry. We tried to select "maven" [here](https://github.com/getsentry/publish/issues/5882) in order to skip it, but it still tries to publish there. Perhaps we have to look into that instead?

PR #612, opened 2025-07-21
feat(maven): add new config to process klib distributions differently from root and apple

buenaflor:
For KMP we are going to publish new targets: `js`, `wasm-js`, `linux`, `mingw` and all the artifacts to publish are currently not compatible with the KMP implementation in craft -> they would be processed incorrectly with the current implementation

All these distributions have the same structure so we can add a new config that targets these. They all contain 5 files: 1 -source.jar, 1 -javadoc.jar, 1 .klib, 1 .module and 1 pom file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,6 @@ Options:
   -h, --help     Show help                                             [boolean]
 ```
 
-
 ### Version naming conventions
 
 Craft currently supports [semantic versioning (semver)](https://semver.org)-like versions for the `NEW-VERSION` argument passed to its `prepare` and `publish` commands. This means, releases made with craft need to follow a general pattern as follows:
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1164,6 +1164,7 @@ targets:
     kmp:
       rootDistDirRegex: /sentry-kotlin-multiplatform-[0-9]+.*$/
       appleDistDirRegex: /sentry-kotlin-multiplatform-(macos|ios|tvos|watchos).*/
+      klibDistDirRegex: /sentry-kotlin-multiplatform-(js|wasm-js).*/
```
buenaflor, line 1167: I'm also open for another name for this config

BYK:
@romtsn I don't really have an opinion here unless it changes a fundamental Craft principle like not publishing the same version twice etc. I don't have an opinion specifically about Maven.

buenaflor:
👍 , no changes here that alter craft behaviours, only adding support for publishing new targets on maven with KMP

I'll merge this then

asottile-sentry (approved):

PR #616, opened 2025-08-27
feat(aws-lambda): Add description including SDK version to layer metadata

msonnb:
This allows users to easily identify which SDK version they use by looking at the description of the layer.
Also boosts confidence that you have input the correct ARN when adding the layer.

### DIFF
--- a/src/utils/__tests__/awsLambdaLayerManager.test.ts
+++ b/src/utils/__tests__/awsLambdaLayerManager.test.ts
@@ -14,7 +14,8 @@ function getTestAwsLambdaLayerManager(): awsManager.AwsLambdaLayerManager {
     'test layer name',
     'test license',
     Buffer.alloc(0),
-    AWS_TEST_REGIONS
+    AWS_TEST_REGIONS,
+    '0.0.0'
   );
 }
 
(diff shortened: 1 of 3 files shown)


src/utils/__tests__/awsLambdaLayerManager.test.ts
```diff
@@ -14,7 +14,8 @@ function getTestAwsLambdaLayerManager(): awsManager.AwsLambdaLayerManager {
     'test layer name',
     'test license',
     Buffer.alloc(0),
-    AWS_TEST_REGIONS
+    AWS_TEST_REGIONS,
+    '0.0.0'
```
andreiborza, line 18: Can we assert the version in a test too?

andreiborza (approved):

src/utils/__tests__/awsLambdaLayerManager.test.ts
andreiborza: Realised this is just on our abstraction, we can just leave this.

msonnb: as discussed offline: not really since we don't return it in `PublishedLayer`

PR #617, opened 2025-09-04
ref: Log descriptive error message when no tags are found during release preparation

Lms24:
A common pitfall when first setting up craft and our publishing workflow in a new repo is that there's no tag yet in this repo. While this is [mentioned as a step](https://develop.sentry.dev/sdk/processes/releases/#the-first-release) in the develop docs, it happens often enough that we don't do it and end up with a less than helpful error in the release preparation flow. This PR simply adds a more descriptive error message so that users have a better idea what to do.

### DIFF
--- a/src/utils/git.ts
+++ b/src/utils/git.ts
@@ -39,7 +39,7 @@ export async function getLatestTag(git: SimpleGit): Promise<string> {
     return (await git.raw('describe', '--tags', '--abbrev=0')).trim();
   } catch (e) {
     logger.error(
-      "Couldn't get the latest tag! If you're releasing for the first time, check if your repo contains any tags. If not, add one manually and try again."
+      'Couldn\'t get the latest tag! If you\'re releasing for the first time, check if your repo contains any tags. If not, add one manually and try again: `git tag 0.0.0 "$(git log -1 --reverse --format=%h)"'
     );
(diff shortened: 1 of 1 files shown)


src/utils/git.ts
```diff
@@ -35,7 +35,15 @@ export async function getDefaultBranch(
 
 export async function getLatestTag(git: SimpleGit): Promise<string> {
   // This part is courtesy of https://stackoverflow.com/a/7261049/90297
-  return (await git.raw('describe', '--tags', '--abbrev=0')).trim();
+  try {
+    return (await git.raw('describe', '--tags', '--abbrev=0')).trim();
+  } catch (e) {
+    logger.error(
+      "Couldn't get the latest tag! If you're releasing for the first time, check if your repo contains any tags. If not, add one manually and try again."
```
asottile-sentry: perhaps even include the command here? `git tag 0.0.0 "$(git log -1 --reverse --format=%h)"`

asottile-sentry (approved):

src/utils/git.ts
Lms24: good idea, thanks! [57245bf](https://github.com/getsentry/craft/pull/617/commits/57245bf8d884a29ec0802f5209762ebed94bb0a7)

PR #619, opened 2025-10-01
chore(nuget): clarify publish behavior reentrant

bruno-garcia:
See:

* https://github.com/getsentry/craft/pull/571#issuecomment-3358211891

#skip-changelog

### DIFF
--- a/README.md
+++ b/README.md
@@ -753,7 +753,8 @@ targets:
 ### NuGet (`nuget`)
 
 Uploads packages to [NuGet](https://www.nuget.org/) via [.NET Core](https://github.com/dotnet/core).
-By default, `craft` publishes all packages with `.nupkg` extension.
+By default, `craft` publishes all packages with `.nupkg` extension by using `dotnet nuget publish --skip-duplicate`
+so re-running `craft` means it will not fail, but instead continue to publish any package it failed.
 
 **Environment**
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -753,7 +753,8 @@ targets:
 ### NuGet (`nuget`)
 
 Uploads packages to [NuGet](https://www.nuget.org/) via [.NET Core](https://github.com/dotnet/core).
-By default, `craft` publishes all packages with `.nupkg` extension.
+By default, `craft` publishes all packages with `.nupkg` extension by using `dotnet nuget publish --skip-duplicate`
+so re-running `craft` means it will not fail, but instead continue to publish any package it failed.
```
bruno-garcia: @BYK Or how about instead?

```suggestion
By default, `craft` publishes all packages with `.nupkg` extension. The target is re-entrant because of `--skip-duplicate`, which allows working around flakiness on the NuGet API by re-running the `craft`.
```

BYK: I wanna put the emphasis on the divergent behavior. What do you think of the following:

```suggestion
Normally, `craft` targets raise an exception when trying to release a version that already exists. *This target diverges from the norm and allows re-entrant publishing* as it can publish multiple packages at once and the processes might get interrupted. This behavior allows us to finalize half-finished releases without having to publish a new version and play cat & mouse with the flaky upstream package repository.
```

BYK (approved):
Added the comment myself for future reference as I have no idea what `--skip-duplicate` would do and wouldn't expect anyone to intuitively understand that if they don't know about NuGet already.

PR #623, opened 2025-11-14
Migrate changelog to GitHub releases

BYK:
Replaces the milestone-based changelog generation with a system that uses GitHub's `release.yml` categories.

This change leverages [GitHub's native release notes configuration](https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes#configuration-options) for more flexible and powerful categorization of PRs based on labels and exclusion rules, moving away from the less flexible milestone-based approach.

### DIFF
--- a/src/utils/__tests__/changelog.test.ts
+++ b/src/utils/__tests__/changelog.test.ts
@@ -550,13 +550,13 @@ describe('generateChangesetFromGit', () => {
       [
         '### Better drivetrain',
         '',
-        '- Upgraded the manifold (#123) by @alice in [#123](https://github.com/test-owner/test-repo/pull/123)',
+        '- Upgraded the manifold by @alice in [#123](https://github.com/test-owner/test-repo/pull/123)',
         '- Refactored the crankshaft by @bob in [#456](https://github.com/test-owner/test-repo/pull/456)',
         '',
         '### Better driver experience',
         '',
-        '- Upgrade the HUD (#789) by @charlie in [#789](https://github.com/test-owner/test-repo/pull/789)',
-        '- Upgrade the steering wheel (#900) by @charlie in [#900](https://github.com/test-owner/test-repo/pull/900)',
+        '- Upgrade the HUD by @charlie in [#789](https://github.com/test-owner/test-repo/pull/789)',
+        '- Upgrade the steering wheel by @charlie in [#900](https://github.com/test-owner/test-repo/pull/900)',
         '',
         '### Other',
         '',
@@ -669,11 +669,11 @@ describe('generateChangesetFromGit', () => {
       [
         '### Better drivetrain',
         '',
-        '- Upgraded the manifold (#123) by @alice in [#123](https://github.com/test-owner/test-repo/pull/123)',
+        '- Upgraded the manifold by @alice in [#123](https://github.com/test-owner/test-repo/pull/123)',
         '',
         '### Better driver experience',
         '',
-        '- Upgrade the HUD (#789) by @charlie in [#789](https://github.com/test-owner/test-repo/pull/789)',
+        '- Upgrade the HUD by @charlie in [#789](https://github.com/test-owner/test-repo/pull/789)',
         '',
         '### Other',
         '',
@@ -749,13 +749,13 @@ describe('generateChangesetFromGit', () => {
       [
         '### Better drivetrain',
         '',
-        '- Upgraded the manifold (#123) by @alice in [#123](https://github.com/test-owner/test-repo/pull/123)',
+        '- Upgraded the manifold by @alice in [#123](https://github.com/test-owner/test-repo/pull/123)',
         '- Refactored the crankshaft by @bob in [#456](https://github.com/test-owner/test-repo/pull/456)',
         '  This is important and we\'ll include the __body__ for attention.',
         '',
         '### Better driver experience',
         '',
-        '- Upgrade the HUD (#789) by @charlie in [#789](https://github.com/test-owner/test-repo/pull/789)',
+        '- Upgrade the HUD by @charlie in [#789](https://github.com/test-owner/test-repo/pull/789)',
         '',
         '### Other',
         '',
@@ -921,7 +921,7 @@ describe('generateChangesetFromGit', () => {
       // PR #1 is excluded from Features category but should appear in Other
       // (category-level exclusions only exclude from that specific category)
       expect(changes).toContain('#1');
-      expect(changes).not.toContain('Feature PR (#1) by @alice in [#1]');
+      expect(changes).not.toContain('Feature PR by @alice in [#1]');
       expect(changes).toContain('#2');
       expect(changes).toContain('### Features');
     });
@@ -951,7 +951,7 @@ describe('generateChangesetFromGit', () => {
 
       const changes = await generateChangesetFromGit(dummyGit, '1.0.0', 3);
       expect(changes).toContain('### All Changes');
-      expect(changes).toContain('Any PR (#1) by @alice in [#1](https://github.com/test-owner/test-repo/pull/1)');
+      expect(changes).toContain('Any PR by @alice in [#1](https://github.com/test-owner/test-repo/pull/1)');
     });
 
     it('should fallback to Other when no config exists', async () => {
--- a/src/utils/changelog.ts
+++ b/src/utils/changelog.ts
@@ -587,16 +587,21 @@ export async function generateChangesetFromGit(
 
     const prEntries = category.prs.map(pr => {
       const prLink = `${prLinkBase}/${pr.number}`;
-      const prTitle = escapeLeadingUnderscores(pr.title);
+      // Strip PR number suffix like "(#123)" since we already include the PR link
+      const prSuffix = `(#${pr.number})`;
+      const title = pr.title.endsWith(prSuffix)
+        ? pr.title.slice(0, -prSuffix.length).trimEnd()
+        : pr.title;
+      const prTitle = escapeLeadingUnderscores(title);
       let entry = `- ${prTitle} by @${pr.author} in [#${pr.number}](${prLink})`;
-      
+
       if (pr.body?.includes(BODY_IN_CHANGELOG_MAGIC_WORD)) {
         const body = pr.body.replace(BODY_IN_CHANGELOG_MAGIC_WORD, '').trim();
         if (body) {
           entry += `\n  ${body}`;
         }
       }
-      
+
       return entry;
     });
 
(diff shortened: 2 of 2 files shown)


src/utils/changelog.ts
```diff
@@ -223,18 +226,193 @@ interface Commit {
...
+  const { exclude } = config.changelog;
+
+  // Check label exclusions
+  if (exclude.labels) {
+    for (const excludeLabel of exclude.labels) {
+      if (labels.includes(excludeLabel)) {
+        return true;
+      }
+    }
+  }
```
BYK: Might be cleaner with `exclude.labels.some(label => labels.includes(label))`. Also we may wanna convert `labels` to a `Set()` for efficient look up if this is happenning frequently.

src/utils/changelog.ts
```diff
@@ -223,18 +226,193 @@ interface Commit {
...
+
+  // Check author exclusions
+  if (exclude.authors && author) {
+    if (exclude.authors.includes(author)) {
+      return true;
+    }
+  }
```
BYK: Like my previous comment: probably better to make `exclude.authors` a `Set()` right after parsing

src/utils/changelog.ts
```diff
@@ -223,18 +226,193 @@ interface Commit {
...
+
+  return false;
+}
+
+/**
+ * Checks if a PR is excluded at the category level
+ * Category-level exclusions completely remove the PR from the changelog
```
BYK: This is incorrect. A category level exclusion just removes the PR from that category. `shouldExcludePR` covers the case where a PR should be completely omitted from changelogs.

src/utils/changelog.ts
```diff
@@ -223,18 +226,193 @@ interface Commit {
...
+  author: string | undefined,
+  config: ReleaseConfig | null
+): boolean {
+  if (!config?.changelog?.categories) {
+    return false;
+  }
+
+  for (const category of config.changelog.categories) {
+    // Check if PR matches category labels
+    if (!category.labels || category.labels.length === 0) {
```
BYK: Let's normalize this data right after parsing the config to either `category.labels` being undefined if it's an empty array or it is always an empty array if it is undefined in the actual config.

src/utils/changelog.ts
```diff
@@ -223,18 +226,193 @@ interface Commit {
...
+      }
+      if (labels.includes(categoryLabel)) {
+        matchesCategory = true;
+        break;
+      }
+    }
+
+    if (!matchesCategory) {
+      continue;
+    }
```
BYK: Do we really need to check if a PR matches a category? I mean if it is excluded, that's enough to _NOT_ have it here. Would that not be more efficient?

src/utils/changelog.ts
```diff
@@ -223,18 +226,193 @@ interface Commit {
...
+        }
+      }
+
+      if (category.exclude.authors && author) {
+        if (category.exclude.authors.includes(author)) {
+          return true; // Excluded at category level
+        }
+      }
+    }
+  }
```
BYK: Like my earlier comments, it's clear that we'd benefit from having `labels` and `category.exclude.authors` as `Set()` objects.

src/utils/changelog.ts
```diff
@@ -223,18 +226,193 @@ interface Commit {
...
+    if (!category.labels || category.labels.length === 0) {
+      continue;
+    }
+
+    let matchesCategory = false;
+    for (const categoryLabel of category.labels) {
+      if (categoryLabel === '*') {
+        matchesCategory = true;
+        break;
+      }
```
BYK: Don't think this logic is correct. Here's what the spec says:
>Use * as a catch-all for pull requests that didn't match any of the previous categories.

That means we should check this last.

src/utils/changelog.ts
```diff
@@ -288,31 +469,61 @@ export async function generateChangesetFromGit(
...
+
+    // Apply global exclusions
+    if (shouldExcludePR(labels, author, releaseConfig)) {
+      continue;
+    }
+
+    // Apply category-level exclusions (these completely hide the PR)
+    if (isCategoryLevelExcluded(labels, author, releaseConfig)) {
+      continue;
+    }
```
BYK: As mentioned earlier, this logic is incorrect. This just hides the PR from a specific category, not necessarily from everything else.

src/utils/changelog.ts
```diff
@@ -288,31 +469,61 @@ export async function generateChangesetFromGit(
...
+
+    // Match PR to category
+    const categoryTitle = matchPRToCategory(labels, releaseConfig);
```
BYK: I think it makes more sense to bake the `isCategoryLevelExcluded` logic into `matchPRToCategory` so it just returns whatever category it matches the PR to, taking the exclusions into account.

src/utils/changelog.ts
```diff
@@ -288,31 +469,61 @@ export async function generateChangesetFromGit(
...
       };
-      // We _know_ the PR exists as milestones are attached to PRs
-      milestone.prs.push({
-        author: commit.author as string,
-        number: commit.pr as string,
-        body: commit.prBody as string,
-      });
-      milestones[commit.milestone] = milestone;
+      // If we have both PR and author, add to category PRs list
+      // Otherwise, add to leftovers (e.g., PR without author, or commit without PR)
```
BYK: Nit: you should never have a PR without an author?

src/utils/changelog.ts
```diff
@@ -323,28 +534,21 @@ export async function generateChangesetFromGit(
...
-  const milestonesInfo = await getMilestoneInfo(Object.keys(milestones));
-
   const changelogSections = [];
-  for (const milestoneNum of Object.keys(milestones)) {
-    const milestone = milestonesInfo[milestoneNum];
-    if (milestone == null) {
-      // XXX(BYK): This case should never happen in real life
-      throw new Error(`Cannot get information for milestone #${milestoneNum}`);
+  // Generate sections for each category
+  for (const categoryTitle of Object.keys(categories)) {
```
BYK: Mighte be better to make `categories` a `Map()`

src/utils/changelog.ts
```diff
@@ -252,6 +253,28 @@ interface ReleaseConfig {
...
+/**
+ * Normalized release config with Sets for efficient lookups
+ */
+interface NormalizedReleaseConfig {
+  changelog: {
+    exclude?: {
+      labels?: Set<string>;
+      authors?: Set<string>;
+    };
+    categories?: NormalizedCategory[];
```
BYK: Don't make anything optional in normalized configs. Empty sets and arrays are easier.

src/utils/changelog.ts
```diff
@@ -252,6 +253,28 @@ interface ReleaseConfig {
...
+  };
+}
+
+interface NormalizedCategory {
+  title: string;
+  labels: string[]; // Normalized: always an array, empty if undefined in config
+  exclude?: {
+    labels?: Set<string>;
+    authors?: Set<string>;
+  };
```
BYK: Same as above, don't make anything optional in normalized configs. Use empty sets when needed.

src/utils/__tests__/changelog.test.ts
```diff
@@ -552,11 +552,13 @@ describe('generateChangesetFromGit', () => {
       [
         '### Better drivetrain',
         '',
-        'By: @alice (#123), @bob (#456)',
+        '- Upgraded the manifold (#123) by @alice in [#123](https://github.com/test-owner/test-repo/pull/123)',
```
BYK: We should strip the PR no at the end of commit logs or PR titles when they appear like this: `(#<PR_NO>)`

src/utils/changelog.ts
```diff
@@ -282,132 +306,162 @@ function readReleaseConfig(): ReleaseConfig | null {
...
+function normalizeReleaseConfig(config: ReleaseConfig | null): NormalizedReleaseConfig | null {
+  if (!config?.changelog) {
+    return null;
+  }
+
+  const normalized: NormalizedReleaseConfig = {
+    changelog: {
+      exclude: {
+        labels: new Set<string>(),
+        authors: new Set<string>(),
```
BYK, line 321: Avoid creating these new sets unless they are actually missing?

src/utils/changelog.ts
```diff
@@ -282,132 +306,162 @@ function readReleaseConfig(): ReleaseConfig | null {
...
-        matchesCategory = true;
-        break;
-      }
-      if (labels.includes(categoryLabel)) {
-        matchesCategory = true;
-        break;
+  if (wildcardCategory) {
+    if (wildcardCategory.exclude.labels.size > 0) {
+      for (const excludeLabel of wildcardCategory.exclude.labels) {
+        if (labels.has(excludeLabel)) {
```
BYK: You should be able to simplify this to `wildcard.Category.exclude.labels.intersection(labels)` (Supported on Node 22+)

src/utils/changelog.ts
```diff
@@ -282,132 +306,162 @@ function readReleaseConfig(): ReleaseConfig | null {
...
+          return null;
+        }
       }
     }
 
-    if (!matchesCategory) {
-      continue;
+    if (wildcardCategory.exclude.authors.size > 0 && author) {
```
BYK: Don't need the size checks, just use `.has` directly.

src/utils/__tests__/changelog.test.ts
```diff
@@ -581,70 +603,12 @@ describe('generateChangesetFromGit', () => {
...
-        '',
-        '### Better Engine',
-        '',
-        'By: @alice (#456)',
-      ].join('\n'),
+      null,
+      '### Other\n\n- Serialized \\_meta (#123)',
     ],
     [
       `should skip commits & prs with the magic ${SKIP_CHANGELOG_MAGIC_WORD}`,
```
BYK, line 610: Add a note that this being redundant now as we can skip PRs with certain labels.

src/utils/changelog.ts
```diff
@@ -323,52 +582,41 @@ export async function generateChangesetFromGit(
...
   if (nLeftovers > 0) {
-    // TODO: read .github/release.yml and its `changelog` section
-    //   if empty, keep using the default `Various fixes & improvements`
-    //   ip not, follow https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes#configuration-options
-    //   to group the leftovers
-
     changelogSections.push(
-      markdownHeader(SUBSECTION_HEADER_LEVEL, 'Various fixes & improvements')
+      markdownHeader(SUBSECTION_HEADER_LEVEL, 'Other')
     );
```
BYK: If there are no other sections, we should skip adding this section title.

PR #639, opened 2025-12-01
fix: $HOME and some others should be passed in env

BYK:
Fixes the `fatal: detected dubious ownership in repository at` errors we got when running `git` operations in post release scripts due to missing `/home/byk` in env. It also passes the Git identity vars for this step as they are overridden in publish repo.

Sample fail: https://github.com/getsentry/publish/actions/runs/19486535961/job/55769925123


hubertdeng123:
Seems like the link you have in the PR is actually a success, this one is a failure: https://github.com/getsentry/publish/actions/runs/19833269111/job/56824349845

PR #641, opened 2025-12-02
fix(changelog): default matcher should match scopes with dashes

BYK:
Fixes conventional commit parsing for scopes with dashes and expands internal changelog categories to include `refactor` and `meta` types.

The previous regex for conventional commit scopes (`\(\w+\)`) did not correctly match scopes containing hyphens (e.g., `fix(my-component):`). This PR updates the regex to properly handle such cases and also adds `refactor` and `meta` commit types to the "Build / dependencies / internal" changelog category for more comprehensive internal change tracking.

### DIFF
--- a/src/utils/__tests__/changelog.test.ts
+++ b/src/utils/__tests__/changelog.test.ts
@@ -1541,6 +1541,8 @@ describe('generateChangesetFromGit', () => {
       expect(changes).toContain('fix(my-component): resolve issue');
       expect(changes).toContain('feat(some-other-scope): add feature');
       expect(changes).toContain('docs(multi-part-scope): update docs');
+      // Should NOT appear in Other section - all commits should be categorized
+      expect(changes).not.toContain('### Other');
     });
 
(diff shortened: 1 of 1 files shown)


src/utils/__tests__/changelog.test.ts
```diff
@@ -1489,5 +1489,149 @@ describe('generateChangesetFromGit', () => {
...
+      );
+
+      const changes = await generateChangesetFromGit(dummyGit, '1.0.0', 10);
+
+      expect(changes).toContain('### Bug Fixes');
+      expect(changes).toContain('### New Features');
+      expect(changes).toContain('### Documentation');
+      expect(changes).toContain('fix(my-component): resolve issue');
+      expect(changes).toContain('feat(some-other-scope): add feature');
+      expect(changes).toContain('docs(multi-part-scope): update docs');
```
BYK, line 1543: We should also ensure these don't end up under `### Other` section. We can probably just assert that the section doesn't exist.

hubertdeng123 (approved):

PR #645, opened 2025-12-03
feat(npm): Add workspaces support

BYK:
Adds support for auto-expanding workspaces to avoid config bloat in monorepos like sentry-javascript

### DIFF
--- a/src/targets/npm.ts
+++ b/src/targets/npm.ts
@@ -225,7 +225,12 @@ export class NpmTarget extends BaseTarget {
         includeNames,
       };
 
-      // Copy over common npm target options
+      // Copy over common target options
+      if (config.excludeNames) {
+        expandedTarget.excludeNames = config.excludeNames;
+      }
+
+      // Copy over npm-specific target options
       if (config.access) {
         expandedTarget.access = config.access;
       }
--- a/src/targets/__tests__/npm.test.ts
+++ b/src/targets/__tests__/npm.test.ts
@@ -1,4 +1,9 @@
-import { getPublishTag, getLatestVersion, NpmTarget } from '../npm';
+import {
+  getPublishTag,
+  getLatestVersion,
+  NpmTarget,
+  NpmPackageAccess,
+} from '../npm';
 import * as system from '../../utils/system';
 import * as workspaces from '../../utils/workspaces';
 
@@ -285,5 +290,47 @@ describe('NpmTarget.expand', () => {
     // Should only include the public package
     expect(result).toHaveLength(1);
     expect(result[0].id).toBe('@sentry/browser');
+  });
+
+  it('propagates excludeNames and other options to expanded targets', async () => {
+    discoverWorkspacesMock = jest
+      .spyOn(workspaces, 'discoverWorkspaces')
+      .mockResolvedValue({
+        type: 'npm',
+        packages: [
+          {
+            name: '@sentry/browser',
+            location: '/root/packages/browser',
+            private: false,
+            hasPublicAccess: true,
+            workspaceDependencies: [],
+          },
+          {
+            name: '@sentry/node',
+            location: '/root/packages/node',
+            private: false,
+            hasPublicAccess: true,
+            workspaceDependencies: [],
+          },
+        ],
+      });
+
+    const config = {
+      name: 'npm',
+      workspaces: true,
+      excludeNames: '/.*-debug\\.tgz$/',
+      access: NpmPackageAccess.PUBLIC,
+      checkPackageName: '@sentry/browser',
+    };
+    const result = await NpmTarget.expand(config, '/root');
+
+    expect(result).toHaveLength(2);
+
+    // Both expanded targets should have the propagated options
+    for (const target of result) {
+      expect(target.excludeNames).toBe('/.*-debug\\.tgz$/');
+      expect(target.access).toBe(NpmPackageAccess.PUBLIC);
+      expect(target.checkPackageName).toBe('@sentry/browser');
+    }
   });
 });


src/targets/npm.ts
```diff
@@ -44,6 +53,29 @@ export interface NpmTargetConfig extends TargetConfig {
...
+   * Enable workspace discovery to auto-generate npm targets for all workspace packages.
+   * When enabled, this target will be expanded into multiple targets, one per workspace package.
+   */
+  workspaces?: boolean;
+  /**
+   * Regex pattern to filter which workspace packages to include.
+   * Only packages matching this pattern will be published.
+   * Example: /^@sentry\//
+   */
+  includeWorkspaces?: string;
```
Lms24, line 66: L: the JSDoc suggests a RegExp but the type requires a string. I see though that we later convert to regexp. Is this to allow regular expression syntax in .craft.yml? Let's just clarify this. Might be as simple as this:
```suggestion
  /**
   * Regex pattern to filter which workspace packages to include.
   * Only packages matching this pattern will be published.
   * Example: '/^@sentry\//'
   */
  includeWorkspaces?: string;
```

src/targets/npm.ts
```diff
@@ -44,6 +53,29 @@ export interface NpmTargetConfig extends TargetConfig {
...
+  /**
+   * Regex pattern to filter which workspace packages to exclude.
+   * Packages matching this pattern will not be published.
+   * Example: /^@sentry-internal\//
+   */
+  excludeWorkspaces?: string;
```
Lms24, line 72: same q here

src/targets/npm.ts
```diff
@@ -44,6 +53,29 @@ export interface NpmTargetConfig extends TargetConfig {
...
+  /**
+   * Template for generating artifact filenames from package names.
+   * Variables: {{name}} (full package name), {{simpleName}} (without @scope/), {{version}}
+   * Default convention: @sentry/browser -> sentry-browser-{version}.tgz
+   */
+  artifactTemplate?: string;
```
Lms24, line 78: q: do we need this for any packages I'm not aware of? in the JS SDK repo we always use the default names, so I don't think we need this customization option here. No objections to keeping it around though, whatever you prefer.

src/targets/npm.ts
```diff
@@ -77,6 +109,120 @@ export class NpmTarget extends BaseTarget {
...
+    }
+
+    const filteredPackages = filterWorkspacePackages(
+      result.packages,
+      includePattern,
+      excludePattern
+    );
+
+    // Also filter out private packages by default (they shouldn't be published)
+    const publishablePackages = filteredPackages.filter(pkg => !pkg.private);
```
Lms24, line 155: This is an important check, thanks for adding (we mark our WIP packages as private for example). 

M: I think we should also check against 

```json
  "publishConfig": {
    "access": "public"
  },
```

correct? I think NPM requires this for publishing. At least [we set it](https://github.com/getsentry/sentry-javascript/blob/fdf22a31afce27e71d74ae46db7de253e53fcb6f/docs/new-sdk-release-checklist.md?plain=1#L74-L75) in every package.json of our SDKs

src/targets/npm.ts
```diff
@@ -77,6 +109,120 @@ export class NpmTarget extends BaseTarget {
...
+    if (result.type === 'none' || result.packages.length === 0) {
+      logger.warn(
+        'npm target has workspaces enabled but no workspace packages were found'
+      );
+      return [];
+    }
+
+    logger.info(
+      `Discovered ${result.packages.length} ${result.type} workspace packages`
+    );
```
Lms24: L: should we debug log all found packages? Might be helpful at least until we know this works

src/utils/__tests__/workspaces.test.ts
```diff
@@ -0,0 +1,263 @@
...
+
+    const sorted = topologicalSortPackages(packages);
+    const names = sorted.map(p => p.name);
+
+    // Verify order: types -> core -> (browser, node) -> integrations
+    expect(names.indexOf('@sentry/types')).toBeLessThan(names.indexOf('@sentry/core'));
+    expect(names.indexOf('@sentry/core')).toBeLessThan(names.indexOf('@sentry/browser'));
+    expect(names.indexOf('@sentry/core')).toBeLessThan(names.indexOf('@sentry/node'));
+    expect(names.indexOf('@sentry/browser')).toBeLessThan(names.indexOf('@sentry/integrations'));
+    expect(names.indexOf('@sentry/node')).toBeLessThan(names.indexOf('@sentry/integrations'));
```
Lms24: this is a good test and depicts something like `@sentry/nextjs` really well that also depends on browser and node, thanks for adding! 

L: Could we add `@sentry/node-core` as a dependency of `@sentry/node` (where node-core also depends on core)? Then we have a second "branch" in the graph.

Super-L: WDYT about checking against the `names` array directly? I think the indexOf is good enough but unless there's something that makes the ordering non-deterministic, the "equals array" (or inline snapshot, whatever you prefer) check makes it super easy to read the test.

(Only suggesting because this is probably the most crucial part to this change. Feel free to disregard)

src/config.ts
```diff
@@ -388,3 +391,57 @@ export function getChangelogConfig(): NormalizedChangelogConfig {
...
+ * @returns The expanded list of target configs
+ */
+export async function expandWorkspaceTargets(
+  targets: TargetConfig[]
+): Promise<TargetConfig[]> {
+  // Lazy import to avoid circular dependency
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const { getTargetByName } = require('./targets');
+
+  const rootDir = getConfigFileDir() || process.cwd();
```
Lms24, line 432: L: just to sanity-check: `.craft.yml` must always be in the root dir, correct? (It is in our repos but was just wondering if there's some other convention)

src/targets/npm.ts
BYK: We can do that with craft targets so not sure if this is actually useful

src/targets/npm.ts
BYK: Fixed, thanks!

src/targets/npm.ts
BYK: Don't think we need to change it but let's keep it for flexibility for now.

src/targets/npm.ts
BYK: Addressed

src/config.ts
BYK: Yes, it must be in root dir based on getConfigFileDir() :)

src/utils/__tests__/workspaces.test.ts
BYK: Added the `node-core` test. Will see for the other request :)

PR #656, opened 2025-12-09
feat: Add automatic version bumping based on conventional commits

BYK:
## Summary

Add automatic semantic version calculation and version bump types, triggered by:
- `craft prepare auto` - determine version from conventional commit analysis
- `craft prepare major` - bump major version
- `craft prepare minor` - bump minor version
- `craft prepare patch` - bump patch version

All options require `minVersion: '2.14.0'` (or higher) in `.craft.yml`.

## Key changes

- Add `semver` field to release config categories (`major`/`minor`/`patch`)
- Update `DEFAULT_RELEASE_CONFIG` with conventional commit semver mappings:
  - Breaking Changes → `major`
  - New Features → `minor`
  - Bug Fixes, Documentation, Build/internal → `patch`
- Create `autoVersion.ts` with `BumpType` enum for efficient max comparison and early exit
- Add `requiresMinVersion()` helper to gate feature
- Update prepare command to accept `auto` and bump type keywords
- Update README with comprehensive documentation

## Behavior

**Auto-versioning:** Finds the highest bump type across matched commits with early exit when a major bump is found. Throws an error if no commits match categories with semver fields defined.

**Version bump types:** Gets the latest tag and applies the specified bump (major/minor/patch) using the semver library.

## Usage

```bash
# Automatic version from commit analysis
craft prepare auto

# Explicit version bumps
craft prepare major   # 1.2.3 -> 2.0.0
craft prepare minor   # 1.2.3 -> 1.3.0
craft prepare patch   # 1.2.3 -> 1.2.4
```

Requires `minVersion: '2.14.0'` (or higher) in `.craft.yml`.

### DIFF
--- a/src/commands/prepare.ts
+++ b/src/commands/prepare.ts
@@ -30,7 +30,8 @@ import { getGitClient, getDefaultBranch, getLatestTag } from '../utils/git';
 import {
   getChangelogWithBumpType,
   calculateNextVersion,
-  BUMP_TYPES,
+  validateBumpType,
+  isBumpType,
   type BumpType,
 } from '../utils/autoVersion';
 import { isDryRun, promptConfirmation } from '../utils/helpers';
@@ -135,7 +136,7 @@ export function checkVersionOrPart(argv: Arguments<any>, _opt: any): boolean {
   }
 
   // Allow version bump types (major, minor, patch)
-  if (BUMP_TYPES.has(version)) {
+  if (isBumpType(version)) {
     return true;
   }
 
@@ -511,7 +512,7 @@ export async function prepareMain(argv: PrepareOptions): Promise<any> {
   }
 
   // Handle automatic version detection or version bump types
-  const isVersionBumpType = BUMP_TYPES.has(newVersion);
+  const isVersionBumpType = isBumpType(newVersion);
 
   if (newVersion === 'auto' || isVersionBumpType) {
     if (!requiresMinVersion(AUTO_VERSION_MIN_VERSION)) {
@@ -527,12 +528,16 @@ export async function prepareMain(argv: PrepareOptions): Promise<any> {
     const latestTag = await getLatestTag(git);
 
     // Determine bump type - either from arg or from commit analysis
-    // Note: getChangelogWithBumpType is memoized, so calling it here and later
-    // in prepareChangelog won't result in duplicate GitHub API calls
-    const bumpType: BumpType =
-      newVersion === 'auto'
-        ? (await getChangelogWithBumpType(git, latestTag)).bumpType
-        : (newVersion as BumpType);
+    // Note: generateChangesetFromGit is memoized, so calling getChangelogWithBumpType
+    // here and later in prepareChangelog won't result in duplicate GitHub API calls
+    let bumpType: BumpType;
+    if (newVersion === 'auto') {
+      const changelogResult = await getChangelogWithBumpType(git, latestTag);
+      validateBumpType(changelogResult); // Throws if no valid bump type
+      bumpType = changelogResult.bumpType;
+    } else {
+      bumpType = newVersion as BumpType;
+    }
 
     // Calculate new version from latest tag
     const currentVersion =
--- a/src/utils/__tests__/autoVersion.test.ts
+++ b/src/utils/__tests__/autoVersion.test.ts
@@ -18,7 +18,11 @@ import type { SimpleGit } from 'simple-git';
 import * as config from '../../config';
 import { getChangesSince } from '../git';
 import { getGitHubClient } from '../githubApi';
-import { calculateNextVersion, getChangelogWithBumpType } from '../autoVersion';
+import {
+  calculateNextVersion,
+  getChangelogWithBumpType,
+  validateBumpType,
+} from '../autoVersion';
 import { clearChangesetCache } from '../changelog';
 
 const getConfigFileDirMock = config.getConfigFileDir as jest.MockedFunction<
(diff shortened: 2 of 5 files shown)


src/commands/prepare.ts
```diff
@@ -485,6 +517,46 @@ export async function prepareMain(argv: PrepareOptions): Promise<any> {
...
+
+    // Calculate new version from latest tag
+    const currentVersion =
+      latestTag && latestTag.replace(/^v/, '').match(/^\d/)
+        ? latestTag.replace(/^v/, '')
+        : '0.0.0';
+
+    newVersion = calculateNextVersion(currentVersion, bumpType);
+    logger.info(
+      `Version bump: ${currentVersion || '(none)'} -> ${newVersion} (${bumpType} bump)`
```
BYK: Current version cannot be empty per the ternary on line 550 above? It would either be "something" or `0.0.0`

src/commands/prepare.ts
```diff
@@ -485,6 +517,46 @@ export async function prepareMain(argv: PrepareOptions): Promise<any> {
...
+    }
+
+    const latestTag = await getLatestTag(git);
+
+    // For 'auto', analyze commits to determine both bump type AND generate changelog
+    // This avoids duplicate GitHub API calls later when preparing the changelog
+    let bumpType: BumpType;
+    if (newVersion === 'auto') {
+      cachedChangelogResult = await getChangelogWithBumpType(git, latestTag);
+      bumpType = cachedChangelogResult.bumpType!; // Already validated non-null by getChangelogWithBumpType
```
BYK: Let's either change the type to never be `null` then or move the check here (which might make more sense)

src/utils/__tests__/autoVersion.test.ts
```diff
@@ -0,0 +1,218 @@
...
+  test('ordered by priority: major > minor > patch', () => {
+    expect(BUMP_TYPES).toEqual(['major', 'minor', 'patch']);
+  });
+
+  test('major has lowest index (highest priority)', () => {
+    expect(BUMP_TYPES.indexOf('major')).toBe(0);
+    expect(BUMP_TYPES.indexOf('minor')).toBe(1);
+    expect(BUMP_TYPES.indexOf('patch')).toBe(2);
+  });
+});
```
BYK: These tests are useless, remove them.

src/utils/changelog.ts
```diff
@@ -642,11 +664,26 @@ function formatChangelogEntry(entry: ChangelogEntry): string {
...
+  /** Number of commits that matched a category with a semver field */
+  matchedCommitsWithSemver: number;
+}
+
 export async function generateChangesetFromGit(
   git: SimpleGit,
   rev: string,
   maxLeftovers: number = MAX_LEFTOVERS
-): Promise<string> {
+): Promise<ChangelogResult> {
```
BYK, line 737: Instead of passing this function's result around for caching, we should memoize this function. This should be easy as none of the arguments (`git`, `rev`, `maxLeftovers`) can be changed throughout a single run (they are the inputs and we don't do multiple passes on different repos etc). Still use proper memoization with taking argument values into account.

Since we are returning a promise, we should memoize the promise so any concurrent calls while the promise is still pending get coalesced.

src/utils/changelog.ts
```diff
@@ -663,6 +700,9 @@ export async function generateChangesetFromGit(
   const leftovers: Commit[] = [];
   const missing: Commit[] = [];
 
+  // Track bump types for auto-versioning
+  const foundBumpTypes = new Set<BumpType>();
```
BYK: You don't need to keep track of a set. Instead of making `BUMP_TYPES` an array, we probably should make it an object:

```
const BUMP_TYPES = {
  "major": 0,
  "minor": 1,
  "patch": 2,
} as const;
```

Then simply look up the value of the bump type we get and use `min()`

```
let bumpVal = null;

bumpVal = min(bumpVal ?? 2, BUMP_TYPES[bumpType]);
```

At the end, if `bumpVal` is `null`, that means we couldn't find anything. If `bumpVal` ever becomes `0`, we can stop checking for it (not sure if this is worth trying to optimize as we'll go through all commits anyway).

src/utils/changelog.ts
```diff
@@ -855,7 +910,12 @@ export async function generateChangesetFromGit(
     }
   }
 
-  return changelogSections.join('\n\n');
+  return {
+    changelog: changelogSections.join('\n\n'),
+    bumpType,
+    totalCommits: gitCommits.length,
+    matchedCommitsWithSemver: foundBumpTypes.size,
```
BYK: This value dosen't seem to be correct as `foundBumpTypes` can only be 3 max (there are only 3 bump types and foundBumpTypes is a set).

src/utils/autoVersion.ts
```diff
@@ -44,13 +52,13 @@ export function calculateNextVersion(
  *
  * @param git The SimpleGit instance
  * @param rev The revision (tag) to analyze from
- * @returns The changelog result containing both changelog and bump type
- * @throws Error if no commits match categories with semver fields
+ * @returns The changelog result with validated non-null bumpType
+ * @throws Error if no commits found or none match categories with semver fields
```
BYK: `getChangelogWithBumpType` should not throw as we may have manual version bumping with auto changelogs.

src/utils/changelog.ts
```diff
@@ -679,10 +687,48 @@ export interface ChangelogResult {
   matchedCommitsWithSemver: number;
 }
 
+// Memoization cache for generateChangesetFromGit
+// Caches the promise to coalesce concurrent calls
+let changesetCache: {
+  key: string;
+  promise: Promise<ChangelogResult>;
+} | null = null;
```
BYK: This should be a `const` and a `Map` instead of an object?

src/utils/changelog.ts
```diff
@@ -679,10 +687,48 @@ export interface ChangelogResult {
...
+function getChangesetCacheKey(rev: string, maxLeftovers: number): string {
+  return `${rev}:${maxLeftovers}`;
+}
+
+/**
+ * Clears the memoization cache for generateChangesetFromGit.
+ * Primarily used for testing.
+ */
+export function clearChangesetCache(): void {
+  changesetCache = null;
```
BYK: If `changesetCache` becomes a `Map()` as I suggested earlier, we can do `changesetCache.clear()`

PR #657, opened 2025-12-10
feat(docker): Add support for multiple registries

### LINKED ISSUE #591: Support multiple docker registries
Currently craft always assumes dockerhub when using the docker target. Ideally we can have multiple docker targets targeting different registries, like Github's GHCR or Google's Artifact Registry.

For this we either need more targets or extend support for the existing docker target to support multiple registries and credentials for these registries.

BYK:
## Summary

Extend the Docker target to support publishing to multiple container registries (Docker Hub, GHCR, GCR, etc.) with per-registry credentials, including cross-registry publishing from private sources.

### Multi-Registry Support
- **Auto-detect registry** from image paths (e.g., `ghcr.io/user/image` → `ghcr.io`)
- **Per-registry credentials** via `DOCKER_<REGISTRY>_USERNAME/PASSWORD` env vars
- **Built-in GHCR defaults** using `GITHUB_ACTOR`/`GITHUB_TOKEN` for zero-config in GitHub Actions
- **Registry override** to share credentials across regional registries
- Treat `docker.io`, `index.docker.io`, `registry-1.docker.io` as Docker Hub aliases

### Cross-Registry Publishing
- **Source registry authentication** for pulling from private registries
- Login to both source and target registries when credentials are available
- If no source credentials found, assumes source is public

### New Nested Object Config Format
Both `source` and `target` now support string or object format:

```yaml
# String shorthand
source: ghcr.io/org/image
target: getsentry/craft

# Object format with options
source:
  image: ghcr.io/org/image
  registry: ghcr.io
  format: "{{{source}}}:sha-{{{revision}}}"
  usernameVar: MY_SOURCE_USER
  passwordVar: MY_SOURCE_PASS
target:
  image: us.gcr.io/project/image
  registry: gcr.io  # Share creds across regions
```

### Security Improvements
- **Secure password handling** using `--password-stdin`

## Example Configurations

Cross-registry publishing:
```yaml
targets:
  - name: docker
    source:
      image: ghcr.io/myorg/private-image
      usernameVar: GHCR_USER
      passwordVar: GHCR_PASS
    target: getsentry/craft
```

Multi-registry publishing:
```yaml
targets:
  # Docker Hub
  - name: docker
    source: ghcr.io/getsentry/craft
    target: getsentry/craft

  # GHCR (zero-config in GitHub Actions)
  - name: docker
    source: ghcr.io/getsentry/craft
    target: ghcr.io/getsentry/craft

  # GCR with shared regional credentials  
  - name: docker
    source: ghcr.io/getsentry/craft
    target:
      image: us.gcr.io/project/image
      registry: gcr.io
```

Closes #591

### DIFF
--- a/src/targets/docker.ts
+++ b/src/targets/docker.ts
@@ -1,6 +1,6 @@
-import * as fs from 'fs';
-import * as path from 'path';
-import * as os from 'os';
+import { existsSync } from 'node:fs';
+import { homedir } from 'node:os';
+import { join } from 'node:path';
 
 import { TargetConfig } from '../schemas/project_config';
 import { BaseArtifactProvider } from '../artifact_providers/base';
@@ -58,19 +58,19 @@ export function hasGcloudCredentials(): boolean {
   ];
 
   for (const credPath of credPaths) {
-    if (credPath && fs.existsSync(credPath)) {
+    if (credPath && existsSync(credPath)) {
       return true;
     }
   }
 
   // Check default Application Default Credentials location
-  const defaultAdcPath = path.join(
-    os.homedir(),
+  const defaultAdcPath = join(
+    homedir(),
     '.config',
     'gcloud',
     'application_default_credentials.json'
   );
-  if (fs.existsSync(defaultAdcPath)) {
+  if (existsSync(defaultAdcPath)) {
     return true;
   }
 
@@ -348,12 +348,12 @@ export class DockerTarget extends BaseTarget {
         const registryHint = registry
           ? `DOCKER_${registryToEnvPrefix(registry)}_USERNAME/PASSWORD or `
           : '';
-        throw new ConfigurationError(
-          `Cannot perform Docker release: missing credentials.
+      throw new ConfigurationError(
+        `Cannot perform Docker release: missing credentials.
 Please use ${registryHint}DOCKER_USERNAME and DOCKER_PASSWORD environment variables.`.replace(
-            /^\s+/gm,
-            ''
-          )
+          /^\s+/gm,
+          ''
+        )
         );
       }
       return undefined;
--- a/src/targets/__tests__/docker.test.ts
+++ b/src/targets/__tests__/docker.test.ts
@@ -1,5 +1,5 @@
-import * as fs from 'fs';
-import * as os from 'os';
+import * as fs from 'node:fs';
+import * as os from 'node:os';
 
 import {
   DockerTarget,
@@ -18,8 +18,8 @@ jest.mock('../../utils/system', () => ({
   spawnProcess: jest.fn().mockResolvedValue(Buffer.from('')),
 }));
 
-jest.mock('fs');
-jest.mock('os');
+jest.mock('node:fs');
+jest.mock('node:os');
 
 describe('normalizeImageRef', () => {
   it('normalizes string source to object with image property', () => {


README.md
```diff
@@ -1064,44 +1064,101 @@ targets:
...
+
+Publishing to multiple registries:
+
+```yaml
+targets:
+  # Docker Hub
+  - name: docker
+    source: ghcr.io/getsentry/craft
+    target: getsentry/craft
+    # Uses DOCKER_USERNAME / DOCKER_PASSWORD
```
aldy505, line 1171: Can we also support `docker.io/getsentry/craft` for DockerHub too?

BYK: Not sure what you mean. Can you provide more details?

aldy505: If you specify your Docker image to be `getsentry/craft`, it's basically an alias for `docker.io/getsentry/craft` (for DockerHub). But in your code, I don't see (or I might don't understand) that if someone put `docker.io/` as the registry prefix, it won't be recognized as DockerHub.

README.md
```diff
@@ -1107,44 +1064,160 @@ targets:
...
+
+  # GHCR (auto-detected, zero-config in GitHub Actions)
+  - name: docker
+    source: ghcr.io/getsentry/craft
+    target: ghcr.io/getsentry/craft
+    # Uses DOCKER_GHCR_IO_* or GITHUB_ACTOR/GITHUB_TOKEN
+
+  # GCR with shared credentials across regions
+  - name: docker
```
hubertdeng123, line 1180: For gcr, it's likely that we're going to need some other way to authenticate than just username + password. We use workload identity provider in GHA right now

src/targets/docker.ts
```diff
@@ -12,22 +12,151 @@ const DEFAULT_DOCKER_BIN = 'docker';
...
+const DOCKER_HUB_REGISTRIES = ['docker.io', 'index.docker.io', 'registry-1.docker.io'];
+
+/**
+ * Extracts the registry host from a Docker image path.
+ *
+ * @param imagePath Docker image path (e.g., "ghcr.io/user/image" or "user/image")
+ * @returns The registry host if present (e.g., "ghcr.io"), undefined for Docker Hub
+ */
+export function extractRegistry(imagePath: string): string | undefined {
+  const parts = imagePath.split('/');
```
hubertdeng123, line 99: It's common for dockerhub registries to be addressed like `getsentry/sentry`, seems like that would break here

aldy505: I don't think it will, the code below checks if `parts.length >= 2` and since `getsentry/sentry` only has 2 in length, it should fall into the `undefined` return value.

BYK: @hubertdeng123 nah it's fine as @aldy505 mentioned this case is covered. There are also tests for this exact scenario.

README.md
BYK: Okay updated our approach:

## How it works

For Google Cloud registries (`gcr.io`, `*.gcr.io`, `*-docker.pkg.dev`), craft now:

1. **Detects gcloud credentials** by checking:
   - `GOOGLE_APPLICATION_CREDENTIALS` env var
   - `GOOGLE_GHA_CREDS_PATH` env var (set by [google-github-actions/auth](https://github.com/google-github-actions/auth))
   - `CLOUDSDK_AUTH_CREDENTIAL_FILE_OVERRIDE` env var
   - Default ADC location: `~/.config/gcloud/application_default_credentials.json`

2. **Checks if gcloud CLI is available** by running `gcloud --version`

3. **Auto-runs `gcloud auth configure-docker`** with the detected registries (comma-separated if multiple)

4. **Skips the regular `docker login`** for those registries

## Usage

With [google-github-actions/auth](https://github.com/google-github-actions/auth), it's now zero-config:

```yaml
# GitHub Actions workflow
- uses: google-github-actions/auth@v2
  with:
    workload_identity_provider: ${{ vars.WORKLOAD_IDENTITY_PROVIDER }}
    service_account: ${{ vars.SERVICE_ACCOUNT }}

- run: craft publish ...

# .craft.yml - no credentials needed!
targets:
  - name: docker
    source: ghcr.io/myorg/image
    target: us-docker.pkg.dev/my-project/repo/image
```

The `skipLogin` option is still available for cases where you've already configured auth externally or need to bypass automatic detection.

aldy505: Did you vibe coded that? 🤣

BYK: How dare you! I _love_ spending multiple hours on reverse-engineering obscure and underdocumented enterprise authentication tools to provide people a good UX.

(if the joke is not obvious, yes I did vibe code it and I wouldn't do it any other way)

src/targets/docker.ts
```diff
@@ -1,3 +1,7 @@
+import * as fs from 'fs';
```
hubertdeng123: Should we be more specific about imports here?

aldy505:
@BYK do you have any plans to cut a release for this one?

PR #667, opened 2025-12-23
feat(github): Integrate action-prepare-release into Craft repo

BYK:
Move the action-prepare-release GitHub Action into the Craft repository root,
enabling usage as `getsentry/craft@v2`. This provides better version alignment
and allows Craft to dogfood its own action.

- Add `floatingTags` config option to create/update floating tags (e.g., v2)
- Support placeholders: {major}, {minor}, {patch}
- Example: `floatingTags: ['v{major}']` creates v2 tag for version 2.15.0

- Parametrized inputs replacing hardcoded Sentry-specific values:
  - `blocker_label` (default: release-blocker)
  - `publish_repo` (default: ${{ github.repository_owner }}/publish)
  - `git_user_name`, `git_user_email` (default to GITHUB_ACTOR)
- Smart Craft installation:
  - Downloads from build artifact when in getsentry/craft repo
  - Falls back to release download using github.action_ref
- Outputs resolved version for downstream steps

- build.yml: Add workflow_call trigger for reusability
- release.yml: Call build workflow first, then use local action for dogfooding

- Add `floatingTags: ['v{major}']` to github target in .craft.yml

Repos using `getsentry/action-prepare-release@v1` can migrate to:
```yaml
uses: getsentry/craft@v2
with:
  version: auto
  git_user_name: getsentry-bot  # if needed
  git_user_email: bot@sentry.io  # if needed
```

### DIFF
--- a/src/commands/prepare.ts
+++ b/src/commands/prepare.ts
@@ -643,13 +643,13 @@ export async function prepareMain(argv: PrepareOptions): Promise<any> {
   // Handle --config-from: load config from remote branch
   if (argv.configFrom) {
     logger.info(`Loading configuration from remote branch: ${argv.configFrom}`);
-    await git.fetch([argv.remote, argv.configFrom]);
     try {
+      await git.fetch([argv.remote, argv.configFrom]);
       const configContent = await git.show([
         `${argv.remote}/${argv.configFrom}:${CONFIG_FILE_NAME}`,
       ]);
       loadConfigurationFromString(configContent);
-    } catch (error) {
+    } catch (error: any) {
       throw new ConfigurationError(
         `Failed to load ${CONFIG_FILE_NAME} from branch "${argv.configFrom}": ${error.message}`
       );


src/commands/prepare.ts
```diff
@@ -95,12 +100,20 @@ export const builder: CommandBuilder = (yargs: Argv) =>
       description: 'The git remote to use when pushing',
       type: 'string',
     })
+    .option('config-from', {
+      description: 'Load .craft.yml from the specified remote branch instead of local file',
+      type: 'string',
+    })
+    .option('calver-offset', {
```
hubertdeng123, line 107: Under what circumstance would this be needed?

hubertdeng123:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

.github/workflows/release.yml
```diff
@@ -10,11 +10,16 @@ on:
       force:
         description: Force a release even when there are release-blockers (optional)
         required: false
-      craft_version:
-        description: Craft version to use for the release
 
 jobs:
+  build:
```
hubertdeng123, line 15: nice, we ensure build before release

hubertdeng123 (approved):

src/commands/prepare.ts
BYK: For us, never as we always use the default value `14`. But since Craft is meant to be a generic tool, some other folks may want to tweak this value.

PR #681, opened 2025-12-29
fix(github): Clean up orphaned draft releases on publish failure

### LINKED ISSUE #388: Release drafts persist after failed publish attempt
When publish action fails, either due to an error, or a timeout, a pre-created draft release will be left untouched.

This means, that whenever someone reapproves the release, they can end up with multiple drafts and a single published release, with the same tag names, which is polluting our GitHub releases page.

It's not possible to query for non-published releases by tag, nor are they listed in `/releases` endpoint, so we cannot just remove or reuse an existing one, as for this, we need a release ID value, which we don't know.

It'd be best to call `deleteRelease` in the `catch` clause of the main `publish` step if anything goes wrong.
However now, due to the timeouts issue that we are trying to resolve, it would not work as well, so leaving this issue description here for future reference.

BYK:
## Summary

When a publish action fails (due to error or timeout) after creating a draft release, the draft is left orphaned on GitHub. This causes multiple draft releases with the same tag name to accumulate when someone re-runs the release.

## Changes

- Add `deleteRelease` method to `GitHubTarget` that safely deletes draft releases
- The method refuses to delete non-draft releases as a safety measure
- Wrap the publish flow in try-catch to clean up drafts on failure
- Apply the same cleanup pattern to the UPM target
- Add tests for draft release cleanup behavior

## Testing

- Added 4 new tests for `deleteRelease` method (draft check, non-draft refusal, undefined handling, dry-run)
- Added 4 tests for publish cleanup behavior (publishRelease failure, deleteRelease failure, success case, uploadAsset failure)
- All 641 tests pass

Fixes #388

### DIFF
--- a/src/targets/__tests__/github.test.ts
+++ b/src/targets/__tests__/github.test.ts
@@ -1,5 +1,7 @@
 import { vi, type Mock, type MockInstance, type Mocked, type MockedFunction } from 'vitest';
-import { isLatestRelease } from '../github';
+import { isLatestRelease, GitHubTarget } from '../github';
+import { NoneArtifactProvider } from '../../artifact_providers/none';
+import { setGlobals } from '../../utils/helpers';
 
 describe('isLatestRelease', () => {
   it('works with missing latest release', () => {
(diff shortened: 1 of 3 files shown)


BYK:
@aldy505 this is only for draft releases during a GitHub publish action. We need to clean up stale publish issues ourselves. I'll go over them and close as there are many old issues that are still open.

PR #682, opened 2025-12-29
fix(brew): skip prereleases and add mustache templating for formula names

### LINKED ISSUE #391: Prevent releasing beta versions and newer major overrides for Brew releases
1. Our Brew releasing workflow doesn't understand to not publish itself when a newer major is present, and it overrides it, which makes patches for older versions override the latest version.
2. Brew doesn't understand pre-releases, and it tries to install `2.0.0-beta.0` which is not available at the time of doing Crafts pre-prelease publish.

beta ref: https://github.com/getsentry/homebrew-tools/pull/3
major ref: https://github.com/getsentry/homebrew-tools/pull/5

BYK:
## Summary

This PR addresses two issues with the Brew target from #391:

1. **Prerelease skip**: Homebrew doesn't understand prerelease versions like `2.0.0-beta.0`, so the Brew target now skips publishing for any preview/prerelease version.

2. **Mustache templating for formula names**: The `formula` config option now supports Mustache-style templates with version variables, allowing versioned formula names that prevent older major patches from overriding newer versions.

### Usage

```yaml
brew:
  tap: getsentry/tools
  formula: 'sentry-cli-v{{{major}}}'  # Creates sentry-cli-v1, sentry-cli-v2, etc.
  template: |
    class SentryCliV{{ major }} < Formula
      # ...
    end
```

### Supported template variables

- `{{{version}}}`: Full version string (e.g., "10.2.3")
- `{{{major}}}`: Major version number (e.g., "10")
- `{{{minor}}}`: Minor version number (e.g., "2")
- `{{{patch}}}`: Patch version number (e.g., "3")

This follows the same pattern as the AWS Lambda Layer target's `layerName` templating.

Fixes #391.

### DIFF
--- a/src/commands/prepare.ts
+++ b/src/commands/prepare.ts
@@ -30,7 +30,7 @@ import {
   handleGlobalError,
   reportError,
 } from '../utils/errors';
-import { getGitClient, getDefaultBranch, getLatestTag } from '../utils/git';
+import { getGitClient, getDefaultBranch, getLatestTag, isRepoDirty } from '../utils/git';
 import {
   getChangelogWithBumpType,
   calculateNextVersion,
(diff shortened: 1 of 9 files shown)


src/targets/brew.ts
```diff
@@ -147,11 +176,18 @@ export class BrewTarget extends BaseTarget {
...
   public async publish(version: string, revision: string): Promise<any> {
-    const { formula, path, template, tapRepo } = this.brewConfig;
-    const { owner, repo } = this.githubRepo;
+    // Skip pre-release versions as Homebrew doesn't understand them
+    if (isPreviewRelease(version)) {
+      this.logger.info(
+        'Skipping Homebrew release for pre-release version: ' + version
+      );
+      return undefined;
+    }
```
armcknight, line 185: 👍🏻

armcknight (approved):
The strategy looks good to help mitigate the last-write-wins issue with homebrew. I can't really help with typescript language/framework details here though.

PR #648, opened 2025-12-04
feat(changelog): Allow changelog entry to be overridden by PR description section

szokeasaurusrex:
## Changelog Entry

 - Add support for custom changelog entries from PR descriptions
 - And with support for multiple entries
   - and nested items

## Description

Implement the ability to override changelog entries using a "Changelog Entry" section in the PR description.

This allows PR authors to provide more detailed, user-facing release notes directly within the PR body, improving clarity and flexibility for changelog generation. If no such section is present, the PR title is used as the changelog entry as before.

### How it works

Add a `## Changelog Entry` section to your PR description with a list of items:

```markdown
## Changelog Entry

- First custom changelog entry
- Second custom changelog entry (optional nested content)
```

Each bullet point becomes a separate changelog entry, allowing a single PR to contribute multiple items to the changelog. This is useful for PRs that contain multiple user-facing changes.

### Example

If your PR title is "refactor: internal changes and add new feature", but you want the changelog to show:

- Add support for X
- Improve performance of Y

Simply add a `## Changelog Entry` section as shown above.

This PR itself demonstrates the feature - look at the changelog preview comment to see how the custom entry is rendered!

PR #690, opened 2025-12-30
fix(changelog): deduplicate merged PR entries in preview

BYK:
## Problem

When viewing the changelog preview for a PR that is already merged (e.g., after editing its title/description), the PR appears twice in the changelog:
- Once from git history (via `fetchRawCommitInfo`)
- Once from the current PR fetch (with `highlight: true`)

Additionally, in repositories using **rebase-merge** workflows, multiple commits from a single PR can each be associated with the same PR number via GitHub's `associatedPullRequests` API, causing duplicate changelog entries.

Fixes #648

## Solution

Added deduplication at two levels:

1. **`fetchRawCommitInfo`**: Deduplicates commits with the same PR number from git history using a `Set`. This handles rebase-merge workflows where multiple commits are associated with the same PR. Keeps the first (newest) occurrence since git log returns commits newest-first.

2. **`generateChangelogWithHighlight`**: Filters the current PR from git history before prepending the fresh highlighted version, ensuring we use up-to-date data (e.g., if the PR title was edited after merge).

## Changes

- `src/utils/changelog.ts`: Added deduplication in `fetchRawCommitInfo` and updated comments in `generateChangelogWithHighlight`
- `src/utils/__tests__/changelog-generate.test.ts`: Added 3 new tests:
  - Merged PR deduplication scenario
  - Rebase-merge workflow with multiple commits per PR
  - Commits without PR association are preserved

### DIFF
--- a/src/utils/changelog.ts
+++ b/src/utils/changelog.ts
@@ -1408,21 +1408,29 @@ export async function generateChangelogWithHighlight(
   // Step 5: Fetch raw commit info up to base branch
   const rawCommits = await fetchRawCommitInfo(git, rev, baseRef);
 
-  // Step 6: Add current PR to the list with highlight flag (at the beginning)
-  const currentPRCommit: RawCommitInfo = {
-    hash: prInfo.headSha,
-    title: prInfo.title.trim(),
-    body: prInfo.body,
-    author: prInfo.author,
-    pr: String(prInfo.number),
-    prTitle: prInfo.title,
-    prBody: prInfo.body,
-    labels: prInfo.labels,
-    highlight: true,
-  };
-  // Prepend current PR to make it newest in the list.
-  // Git log returns commits newest-first, so this maintains that order.
-  const allCommits = [currentPRCommit, ...rawCommits];
+  // Step 6: Build combined commit list with current PR highlighted at the beginning.
+  // Filter out the current PR from git history if already merged - this ensures we use
+  // the freshly-fetched PR data (with highlight) instead of potentially stale data
+  // from git history (e.g., if PR title was edited after merge).
+  const currentPRStr = String(currentPRNumber);
+  const allCommits: RawCommitInfo[] = [
+    {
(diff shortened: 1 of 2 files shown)


src/utils/changelog.ts
```diff
@@ -1420,9 +1420,16 @@ export async function generateChangelogWithHighlight(
...
+  // Step 6b: Filter out the current PR from git history if already merged.
+  // This ensures we use the freshly-fetched PR data (with highlight) instead of
+  // potentially stale data from git history (e.g., if PR title was edited after merge).
+  const currentPRStr = String(currentPRNumber);
+  const dedupedRawCommits = rawCommits.filter(c => c.pr !== currentPRStr);
+
   // Prepend current PR to make it newest in the list.
   // Git log returns commits newest-first, so this maintains that order.
-  const allCommits = [currentPRCommit, ...rawCommits];
+  const allCommits = [currentPRCommit, ...dedupedRawCommits];
```
BYK: Wouldn't it be more efficient to do a single for-of loop and add all commits from `rawCommits` to `allCommits` with the filter baked in?

src/utils/changelog.ts
```diff
@@ -1489,6 +1496,18 @@ async function fetchRawCommitInfo(
...
+  // This handles:
+  // 1. Rebase-merge workflows where multiple commits share the same PR number
+  // 2. Already-merged PRs that appear in git history when generating previews
+  const seenPRs = new Set<string>();
+  return allCommits.filter(commit => {
+    if (!commit.pr) return true; // Keep commits without PR association
+    if (seenPRs.has(commit.pr)) return false;
+    seenPRs.add(commit.pr);
+    return true;
+  });
```
BYK: Similar to above, instead of using `.map` and `.filter`, why not have `const allCommits = []` first and then do a for-of loop on gitCommits and keep tabs on `seenPRs` during that loop, skipping the rest and saving work on them?

PR #707, opened 2026-01-12
feat: Automatic version bumping from configured targets

### LINKED ISSUE #76: Bump versions directly for known package managers
It would be nice if you could avoid having to write a `scripts/bump-version.sh` script for projects that have a `Cargo.toml` or a `package.json`.

Ideally, this should be configurable, in case you actually have a rust project, but with a `package.json` which might only be used for development.

BYK:
## Summary

When `minVersion >= 2.21.0` and no custom `preReleaseCommand` is defined, Craft automatically bumps version numbers based on configured publish targets. This eliminates the need for a `scripts/bump-version.sh` script in most cases.

## How It Works

1. Craft examines configured `targets` in `.craft.yml`
2. For each target that supports version bumping, it updates the appropriate project files
3. Targets are processed in config order, deduplicated by type

## Supported Targets

| Target | Detection | Version Bump Method |
|--------|-----------|---------------------|
| `npm` | `package.json` | `npm version --no-git-tag-version` |
| `pypi` | `pyproject.toml` | hatch, poetry, setuptools-scm, or direct edit |
| `crates` | `Cargo.toml` | `cargo set-version` (cargo-edit) |
| `gem` | `*.gemspec` | Direct edit of gemspec + `lib/**/version.rb` |
| `pub-dev` | `pubspec.yaml` | Direct edit |
| `hex` | `mix.exs` | Direct edit |
| `nuget` | `*.csproj` | dotnet-setversion or direct XML edit |

## Python (pypi) Detection Priority

1. `[tool.hatch]` → `hatch version <version>`
2. `[tool.poetry]` → `poetry version <version>`
3. `[tool.setuptools_scm]` → No-op (version from git tags)
4. `[project]` with version → Direct TOML edit

## Breaking Changes

None. Existing projects continue to work unchanged. The feature is gated behind `minVersion: "2.21.0"`.

Fixes #76

### DIFF
--- a/docs/src/content/docs/configuration.md
+++ b/docs/src/content/docs/configuration.md
@@ -136,6 +136,28 @@ If automatic version bumping fails:
 
 In all error cases, Craft suggests defining a custom `preReleaseCommand` as a fallback.
 
+### Recovery from Failed Prepare
+
+If version bumping succeeds but `craft prepare` fails mid-way (e.g., during changelog generation or git operations), you may need to clean up manually:
+
+1. **Check the release branch**: If a release branch was created, you can delete it:
+
+   ```bash
+   git branch -D release/<version>
+   ```
+
+2. **Revert version changes**: If files were modified but not committed, reset them:
+
+   ```bash
+   git checkout -- package.json pyproject.toml Cargo.toml  # or whichever files were changed
+   ```
+
+3. **Re-run prepare**: Once the issue is fixed, run `craft prepare` again. Version bumping is idempotent—running it multiple times with the same version is safe.
+
+:::tip
+Use `craft prepare --dry-run` first to preview what changes will be made without modifying any files.
+:::
+
 ## Post-release Command
 
 This command runs after a successful `craft publish`. Default: `bash scripts/post-release.sh`.
--- a/src/commands/prepare.ts
+++ b/src/commands/prepare.ts
@@ -323,17 +323,24 @@ export async function runPreReleaseCommand(
     targets.length > 0
   ) {
     logger.info('Running automatic version bumping from targets...');
-    const anyBumped = await runAutomaticVersionBumps(
-      targets,
-      rootDir,
-      newVersion,
-    );
+    const result = await runAutomaticVersionBumps(targets, rootDir, newVersion);
 
-    if (!anyBumped) {
-      logger.warn('No targets support automatic version bumping');
+    if (!result.anyBumped) {
+      if (result.bumpableTargets.length === 0) {
+        logger.warn(
+          'None of your configured targets support automatic version bumping. ' +
+            'Consider adding a preReleaseCommand to bump versions manually.',
+        );
+      } else {
+        logger.warn(
+          `Targets [${result.skippedTargets.join(', ')}] support version bumping ` +
+            'but did not find applicable files in your project. ' +
+            'Consider adding a preReleaseCommand if you need custom version bumping.',
+        );
+      }
     }
 
-    return anyBumped;
+    return result.anyBumped;
   }
 
   return runCustomPreReleaseCommand(oldVersion, newVersion, undefined);
--- a/src/__tests__/versionBump.test.ts
+++ b/src/__tests__/versionBump.test.ts
@@ -68,29 +68,35 @@ describe('runAutomaticVersionBumps', () => {
     vi.resetAllMocks();
   });
 
-  test('returns false when no targets are provided', async () => {
+  test('returns empty result when no targets are provided', async () => {
     const result = await runAutomaticVersionBumps([], tempDir, '1.0.0');
-    expect(result).toBe(false);
+    expect(result.anyBumped).toBe(false);
+    expect(result.bumpableTargets).toEqual([]);
+    expect(result.skippedTargets).toEqual([]);
   });
 
-  test('returns false when target does not support version bumping', async () => {
+  test('returns no bumpable targets when target does not support version bumping', async () => {
     // 'github' target doesn't have bumpVersion
     const result = await runAutomaticVersionBumps(
       [{ name: 'github' }],
       tempDir,
       '1.0.0',
     );
-    expect(result).toBe(false);
+    expect(result.anyBumped).toBe(false);
+    expect(result.bumpableTargets).toEqual([]);
+    expect(result.skippedTargets).toEqual([]);
   });
 
-  test('returns false when target detection fails', async () => {
+  test('returns skipped target when target detection fails', async () => {
     // npm target but no package.json
     const result = await runAutomaticVersionBumps(
       [{ name: 'npm' }],
       tempDir,
       '1.0.0',
     );
-    expect(result).toBe(false);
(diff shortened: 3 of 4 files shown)


src/targets/gem.ts
```diff
@@ -26,6 +30,85 @@ export class GemTarget extends BaseTarget {
...
+   *
+   * Looks for version patterns in:
+   * 1. .gemspec files (s.version = "x.y.z")
+   * 2. lib/.../version.rb files (VERSION = "x.y.z")
+   */
+  public static async bumpVersion(
+    rootDir: string,
+    newVersion: string
+  ): Promise<boolean> {
+    const gemspecFiles = readdirSync(rootDir).filter(f => f.endsWith('.gemspec'));
```
sl0thentr0py: `sentry-ruby` is a monorepo of several gems where each gemspec is not in the root dir but a subdir, example: https://github.com/getsentry/sentry-ruby/blob/master/sentry-ruby/sentry-ruby.gemspec

docs/src/content/docs/configuration.md
```diff
@@ -41,6 +41,91 @@ export npm_config_git_tag_version=false
...
+| `pub-dev` | `pubspec.yaml` exists | Direct edit of pubspec.yaml |
+| `hex` | `mix.exs` exists | Direct edit of mix.exs |
+| `nuget` | `*.csproj` exists | dotnet-setversion or direct XML edit |
+
+### npm Workspace Support
+
+For npm/yarn/pnpm monorepos, Craft automatically detects and bumps versions in all workspace packages:
+
+- **npm 7+**: Uses `npm version --workspaces` to bump all packages at once
+- **yarn/pnpm or npm < 7**: Falls back to bumping each non-private package individually
```
betegon, line 79: where's my dear Bun?  we can add it later, i think we could just look for Bun lockfiles `bun.lockb` / `bun.lock`, add `BUN_BIN`, `USE_BUN` env vars, and `bun publish`support.

betegon: it supports workspaces too https://bun.com/docs/pm/workspaces

BYK:
**Re: Gem monorepo support**

Updated! The gem target now searches for `.gemspec` files up to 2 levels deep, respecting `.gitignore` patterns. This supports monorepo structures like:
- `sentry-ruby/sentry-ruby.gemspec` (depth 1)
- `packages/gem-name/gem-name.gemspec` (depth 2)

The implementation uses the `ignore` npm package for gitignore parsing and async `fs/promises` APIs.

---

**Re: Bun support**

Great suggestion! Bun support (lockfile detection via `bun.lockb`/`bun.lock`, `BUN_BIN`/`USE_BUN` env vars, `bun publish`, and workspace support) would be a valuable addition. I'll address this in a follow-up PR to keep this one focused on the automatic version bumping feature.

src/commands/prepare.ts
```diff
@@ -67,6 +73,9 @@ const DEFAULT_BUMP_VERSION_PATH = join('scripts', 'bump-version.sh');
 /** Minimum craft version required for auto-versioning */
 const AUTO_VERSION_MIN_VERSION = '2.14.0';
 
+/** Minimum craft version required for automatic version bumping from targets */
+const AUTO_BUMP_MIN_VERSION = '2.19.0';
```
BYK:
```suggestion
const AUTO_BUMP_MIN_VERSION = '2.21.0';
```

src/targets/pypi.ts
```diff
@@ -36,6 +44,130 @@ export class PypiTarget extends BaseTarget {
...
+        `Cannot find "${HATCH_BIN}" for version bumping. ` +
+          'Install hatch or define a custom preReleaseCommand in .craft.yml'
+      );
+    }
+
+    logger.debug(`Running: ${HATCH_BIN} version ${newVersion}`);
+    await spawnProcess(HATCH_BIN, ['version', newVersion], { cwd: rootDir });
+
+    return true;
+  }
```
BYK: This "if has executable run it with these commands, otherwise throw error" pattern repeats on almost all targests. We should probably abstract it out and provide a helper.

src/utils/files.ts
```diff
@@ -145,3 +147,102 @@ export function detectContentType(artifactName: string): string | undefined {
...
+  const results: string[] = [];
+
+  let entries;
+  try {
+    entries = await readdirAsync(currentDir, { withFileTypes: true });
+  } catch {
+    return results;
+  }
+
+  for (const entry of entries) {
```
BYK: Using `opendir` with an async for loop might be better here: https://nodejs.org/api/fs.html#fspromisesopendirpath-options

docs/src/content/docs/configuration.md
```diff
@@ -48,6 +48,91 @@ export npm_config_git_tag_version=false
...
+2. **Poetry** - If `[tool.poetry]` section exists → `poetry version <version>`
+3. **setuptools-scm** - If `[tool.setuptools_scm]` section exists → No-op (version derived from git tags)
+4. **Direct edit** - If `[project]` section with `version` field exists → Edit `pyproject.toml` directly
+
+### Enabling Automatic Version Bumping
+
+To enable automatic version bumping, ensure your `.craft.yml` has:
+
+```yaml
+minVersion: "2.19.0"
```
BYK:
```suggestion
minVersion: "2.21.0"
```

docs/src/content/docs/configuration.md
```diff
@@ -48,6 +48,91 @@ export npm_config_git_tag_version=false
...
+And either:
+- Remove any custom `preReleaseCommand`, or
+- Don't define `preReleaseCommand` at all
+
+### Disabling Automatic Version Bumping
+
+To disable automatic version bumping while still using minVersion 2.19.0+:
+
+```yaml
+minVersion: "2.19.0"
```
BYK:
```suggestion
minVersion: "2.21.0"
```

docs/src/content/docs/configuration.md
```diff
@@ -48,6 +48,91 @@ export npm_config_git_tag_version=false
...
+preReleaseCommand: ""  # Explicitly set to empty string
+```
+
+Or define a custom script:
+
+```yaml
+minVersion: "2.19.0"
```
BYK:
```suggestion
minVersion: "2.21.0"
```

src/targets/pypi.ts
BYK: Done in 9ad56d0 and eba3c11:

- Added `runWithExecutable()` helper in `system.ts`
- Extended to support array of executables with `findFirstExecutable()` and `requireFirstExecutable()`
- Updated pypi.ts, crates.ts, and npm.ts to use the new abstraction

The npm target now uses `requireFirstExecutable([NPM_CONFIG, YARN_CONFIG])` to try npm first, then yarn.

src/utils/files.ts
BYK: Done in e65c06b - switched to `opendir` with async `for await...of` loop and proper cleanup via `try/finally`.

BYK:
@cursoragent review

PR #709, opened 2026-01-13
feat: Configure GitHub artifact provider to filter by workflow and name

BYK:


betegon (approved):

src/schemas/project_config.ts
```diff
@@ -82,6 +82,36 @@ export const BaseArtifactProviderSchema = z.object({
...
+const ArtifactPatternsSchema = z.union([z.string(), z.array(z.string())]);
+
+/**
+ * Artifacts config for GitHub artifact provider.
+ * Accepts string, array of strings, or object mapping workflow names to artifact patterns.
+ */
+export const GitHubArtifactsConfigSchema = z
+  .union([
+    z.string(),
+    z.array(z.string()),
```
BYK:
```suggestion
    ArtifactPatternsSchema,
```

src/artifact_providers/github.ts
```diff
@@ -17,12 +18,85 @@ import {
...
+
+/**
+ * Converts a string to a RegExp if it's wrapped in slashes, otherwise creates an exact match RegExp
+ */
+export function patternToRegexp(pattern: string): RegExp {
+  if (pattern.startsWith('/') && pattern.lastIndexOf('/') > 0) {
+    return stringToRegexp(pattern);
+  }
+  // Exact match (escape special regex characters)
+  const escaped = pattern.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
```
BYK: Don't we already have a regex escape helper function somewhere? If not we should probably make this a utility. Same with `patternToRegexp` here.

src/artifact_providers/github.ts
```diff
@@ -230,12 +304,245 @@ export class GitHubArtifactProvider extends BaseArtifactProvider {
...
+          run_id: run.id,
+          per_page,
+          page,
+        });
+
+        for (const artifact of response.data.artifacts) {
+          for (const filter of filters) {
+            if (filter.workflow && !filter.workflow.test(workflowName)) {
+              continue;
+            }
```
BYK: Don't we already filter workflow names before coming here? Why do we have this additional check. If we are gonna check this here, I think the optimization we had by defining workflow names to reduce number of API calls is meaningless.

src/artifact_providers/github.ts
```diff
@@ -230,12 +304,245 @@ export class GitHubArtifactProvider extends BaseArtifactProvider {
...
+
+            const matches = filter.artifacts.some(pattern =>
+              pattern.test(artifact.name)
+            );
+            if (matches) {
+              this.logger.debug(
+                `Artifact "${artifact.name}" matches filter from workflow "${workflowName}"`
+              );
+              matchingArtifacts.push(artifact);
+              break;
```
BYK: Why do we break here, what if we have multiple artifacts matching the name. I think we should be collecting all, not the first one matching.

src/artifact_providers/github.ts
```diff
@@ -230,12 +304,245 @@ export class GitHubArtifactProvider extends BaseArtifactProvider {
...
+    artifactItems: ArtifactItem[]
+  ): Promise<RemoteArtifact[]> {
+    const limit = pLimit(DOWNLOAD_CONCURRENCY);
+    const allArtifacts: RemoteArtifact[] = [];
+
+    const downloadPromises = artifactItems.map(item =>
+      limit(async () => {
+        this.logger.debug(`Downloading artifact "${item.name}"...`);
+        const url = await this.getArchiveDownloadUrl(item);
+        const artifacts = await this.downloadAndUnpackArtifacts(url);
```
BYK: I think we should separate downloading and unpacking as disk I/O and CPU for unpacking should not block network operations. These should happen in two steps in a pipeline: download all in parallel -> unpack all in parallel.

src/artifact_providers/github.ts
```diff
@@ -18,9 +18,12 @@ import {
 } from '../utils/files';
 import { extractZipArchive } from '../utils/system';
 import { sleep } from '../utils/async';
-import { stringToRegexp } from '../utils/filters';
+import { patternToRegexp } from '../utils/filters';
 import { GitHubArtifactsConfig } from '../schemas/project_config';
 
+// Re-export for backward compatibility with tests
+export { patternToRegexp } from '../utils/filters';
```
BYK: Nope, just update the tests.

src/schemas/project_config.ts
BYK: I proposed replacing the `z.string(), z.array(z.string())` bit with `ArtifactPatternsSchema` as that expresses our intent much more clearly.

src/utils/filters.ts
BYK, line 17: See, we already have `stringToRegexp` here. We should be using this instead of writing a new `patternToRegexp`.

PR #714, opened 2026-01-13
fix(aws-lambda): Skip layer publication for pre-release versions

### LINKED ISSUE #609: Skip publishing lambda layers for pre-releases
### Description

We published an alpha release of the sentry-javascript SDKs earlier today and despite it being an alpha-release, the Lambda layer was published.

We brought this up previously in [this issue](https://github.com/getsentry/craft/issues/537) but thought this function prevents publishing the layer:

https://github.com/getsentry/craft/blob/532b8d1950b351a4ca66f825ab906238a1cbf563/src/targets/awsLambdaLayer.ts#L196C11-L208 

Upon further reflection, the function is just used to prevent publishing to our release-registry but it does not prevent publishing the layer to AWS.

### Expected Result

Pre-releases not to publish our production AWS Lambda layers on AWS.

### Actual Result

Pre-release published a new version of our AWS Lambda layer on AWS.

BYK:
## Summary

- Skip AWS Lambda layer publication for pre-release versions by default
- Adds early check in `publish()` to return before any artifact fetching or layer publishing for prereleases
- Respects existing `linkPrereleases` config option to allow prerelease publishing when explicitly enabled
- Adds tests for the new prerelease skipping behavior

This aligns the publish behavior with the existing `isPushableToRegistry` logic which already skips registry updates for prereleases.

Fixes #609.

### DIFF
--- a/src/targets/__tests__/awsLambda.test.ts
+++ b/src/targets/__tests__/awsLambda.test.ts
@@ -1,4 +1,10 @@
-import { vi, type Mock, type MockInstance, type Mocked, type MockedFunction } from 'vitest';
+import {
+  vi,
+  type Mock,
+  type MockInstance,
+  type Mocked,
+  type MockedFunction,
+} from 'vitest';
 import { NoneArtifactProvider } from '../../artifact_providers/none';
 import { ConfigurationError } from '../../utils/errors';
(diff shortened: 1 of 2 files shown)


andreiborza (approved):
This'll make our (sdk maintainers) lives so much easier, thanks for taking care of this @BYK 😃

PR #716, opened 2026-01-14
fix(action): Use environment variables for complex inputs

BYK:
## Description

This PR refactors the 'Set git user' and 'Request publish' steps in `action.yml` to use environment variables for potentially complex data instead of direct string interpolation in bash scripts.

## Problem

Injecting content like `${{ steps.craft.outputs.changelog }}` directly into a bash script using single quotes fails if the content contains single quotes or other special characters (e.g., `syntax error near unexpected token '('`).

## Solution

Pass these values as environment variables, which is the recommended and safe way to handle complex string data in GitHub Actions.

## Changes

- Updated 'Set git user' to use `INPUT_GIT_USER_NAME` and `INPUT_GIT_USER_EMAIL` env vars.
- Updated 'Request publish' to use environment variables for `CHANGELOG`, `TARGETS`, `VERSION`, etc.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -77,20 +77,9 @@ runs:
     - name: Set git user
       shell: bash
       env:
-        INPUT_GIT_USER_NAME: ${{ inputs.git_user_name }}
-        INPUT_GIT_USER_EMAIL: ${{ inputs.git_user_email }}
+        GIT_USER_NAME: ${{ inputs.git_user_name || github.actor }}
+        GIT_USER_EMAIL: ${{ inputs.git_user_email || format('{0}+{1}@users.noreply.github.com', github.actor_id, github.actor) }}
       run: |
-        # Use provided values or fall back to triggering actor
-        GIT_USER_NAME="${INPUT_GIT_USER_NAME}"
-        GIT_USER_EMAIL="${INPUT_GIT_USER_EMAIL}"
-
-        if [[ -z "$GIT_USER_NAME" ]]; then
-          GIT_USER_NAME="${GITHUB_ACTOR}"
-        fi
-        if [[ -z "$GIT_USER_EMAIL" ]]; then
-          GIT_USER_EMAIL="${GITHUB_ACTOR_ID}+${GITHUB_ACTOR}@users.noreply.github.com"
-        fi
-
         echo "GIT_COMMITTER_NAME=${GIT_USER_NAME}" >> $GITHUB_ENV
         echo "GIT_AUTHOR_NAME=${GIT_USER_NAME}" >> $GITHUB_ENV
         echo "EMAIL=${GIT_USER_EMAIL}" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


action.yml
```diff
@@ -76,10 +76,13 @@ runs:
 
     - name: Set git user
       shell: bash
+      env:
+        INPUT_GIT_USER_NAME: ${{ inputs.git_user_name }}
+        INPUT_GIT_USER_EMAIL: ${{ inputs.git_user_email }}
```
BYK: Why not doing this
```suggestion
        GIT_USER_NAME: ${{ inputs.git_user_name }}
        GIT_USER_EMAIL: ${{ inputs.git_user_email }}
```

And then getting rid of the assigments on lines 84 and 85?

action.yml
BYK: Pretty sure we can also fold this logic into the `env` block, making it safer and tidier? The less bash the better?

BYK, line 85: I wonder if we can lift these up to job level `env` block and completly get rid of the `Set git user` step?

action.yml
```diff
@@ -194,21 +197,31 @@ runs:
...
-          merge_target='${{ inputs.merge_target }}'
+        if [[ -n "$INPUT_MERGE_TARGET" ]]; then
+          merge_target="$INPUT_MERGE_TARGET"
         else
           merge_target='(default)'
         fi
 
         # Use resolved version from Craft output
-        RESOLVED_VERSION="${{ steps.craft.outputs.version }}"
+        RESOLVED_VERSION="$VERSION"
```
BYK: Do we need the `RESOLVED_VERSION` - `VERSION` indirection? Why not map the value directly to `RESOLVED_VERSION` in the env block above?

action.yml
```diff
@@ -194,21 +197,31 @@ runs:
...
+          subdirectory="/$INPUT_PATH"
         fi
 
-        if [[ -n '${{ inputs.merge_target }}' ]]; then
-          merge_target='${{ inputs.merge_target }}'
+        if [[ -n "$INPUT_MERGE_TARGET" ]]; then
+          merge_target="$INPUT_MERGE_TARGET"
         else
           merge_target='(default)'
         fi
```
BYK: I think this logic can also be folded in to the `env` block above.

action.yml
```diff
@@ -234,17 +247,16 @@ runs:
...
+        RELEASE_BRANCH="$BRANCH"
+        RELEASE_SHA="$SHA"
+        RELEASE_PREVIOUS_TAG="$PREVIOUS_TAG"
 
         # Fall back to HEAD if no previous tag
-        if [[ -z "$PREVIOUS_TAG" ]]; then
-          PREVIOUS_TAG="HEAD"
+        if [[ -z "$RELEASE_PREVIOUS_TAG" ]]; then
+          RELEASE_PREVIOUS_TAG="HEAD"
         fi
```
BYK: Same like others: this entire logic block can be put into the `env` block.

BYK:
I have addressed the review comments by refactoring the bash logic into GitHub Actions expressions in the `env` blocks. This makes the steps tidier and safer.

PR #722, opened 2026-01-15
feat(github): Add comment option to changelog-preview workflow

BYK:
## Summary

Adds a `comment: true|false` input to the changelog-preview workflow to reduce notification noise. When set to `false`, the workflow creates a commit status instead of posting PR comments.

## Changes

- **New Input**: `comment` (boolean, defaults to `true` for backward compatibility)
- **Status Check Mode** (`comment: false`):
  - Creates a commit status named "Changelog Preview / Semver Impact"
  - Shows semver impact clearly (e.g., "Minor", "Major", "Patch")
  - Status appears independently in PR checks (not grouped with other workflows)
  - Links to GitHub Actions job summary with full changelog preview
  - Reduces notification noise significantly
- **Comment Mode** (`comment: true`, default):
  - Preserves existing PR comment behavior
  - Maintains backward compatibility
- **Dogfooding**: Craft's own PRs automatically use status check mode

## Implementation Details

Uses the **Commit Status API** rather than Check Runs API because:
- Works reliably with `GITHUB_TOKEN` (no GitHub App required)
- Appears as independent status (no grouping under other workflows)
- Simpler implementation
- Status description limited to plain text (no emoji support)

The full changelog is always available in the GitHub Actions job summary (accessible via the status link).

## Documentation

Updated `docs/src/content/docs/github-actions.md` with:
- New `comment` input documentation
- "Output Modes" section explaining both modes with pros/cons
- Examples for both comment and status check modes
- Updated permissions requirements (`statuses: write` for status check mode)

## Example

This PR demonstrates status check mode. Look for:
```
Changelog Preview / Semver Impact — Minor
```
in the checks section below.

### DIFF
--- a/docs/src/content/docs/github-actions.md
+++ b/docs/src/content/docs/github-actions.md
@@ -204,7 +204,6 @@ jobs:
 - Cleaner PR interface
 - Semver impact visible in status checks section
 - Full changelog available in Actions job summary
-- Status appears independently (not grouped with other checks)
 
 **Cons:**
 


.github/workflows/changelog-preview.yml
```diff
@@ -137,9 +157,82 @@ jobs:
...
+              BUMP_SHORT="Major"
+              BUMP_EMOJI="🔴"
+              ;;
+            minor) 
+              BUMP_SHORT="Minor"
+              BUMP_EMOJI="🟡"
+              ;;
+            patch) 
+              BUMP_SHORT="Patch"
+              BUMP_EMOJI="🟢"
```
betegon, line 172: that color codes are WORLDWIDE known to signal: 
green: go ahead, yellow: warning, and red: don't.

what about using arbitrary colors instead ☮️

docs/src/content/docs/github-actions.md
```diff
@@ -142,9 +142,86 @@ jobs:
...
+    secrets: inherit
+```
+
+**Pros:**
+
+- Minimal notification noise
+- Cleaner PR interface
+- Semver impact visible in status checks section
+- Full changelog available in Actions job summary
+- Status appears independently (not grouped with other checks)
```
BYK:
```suggestion
```

.github/workflows/changelog-preview.yml
BYK: That was kind of the intention though?

Patch -> Always safe
Minor -> Might be safe
Major -> Danger, be careful!

PR #748, opened 2026-02-11
fix(artifacts): Support glob patterns and validate all configured patterns match

BYK:
## Summary

Fixes a class of silent misconfiguration bugs in artifact pattern matching, motivated by [getsentry/cli#230](https://github.com/getsentry/cli/pull/230) where `'sentry-*'` in `.craft.yml` was silently treated as an exact match for the literal string `"sentry-*"` instead of matching `"sentry-linux-x64"` etc.

### Changes

**Glob pattern support** (`src/utils/filters.ts`):
- `patternToRegexp()` now supports glob wildcards: `*` maps to `.*`, `?` maps to `.` in regex
- Plain strings and `/regex/` syntax remain unchanged
- New `globToRegex()` helper function

**All-patterns-must-match validation** (`src/artifact_providers/github.ts`):
- New `validateAllPatternsMatched()` method ensures every configured workflow pattern matched at least one workflow run, and every artifact pattern matched at least one artifact
- Runs after the existing retry/polling loop succeeds — "wait for CI" behavior is preserved
- Error messages include available workflow/artifact names for easy debugging
- Reports all mismatches at once

**Publish-level safeguard** (`src/commands/publish.ts`):
- When `artifactProvider.config.artifacts` is set but zero artifacts are found, `reportError()` stops the publish early instead of proceeding to targets that will individually fail with confusing messages

### Backward Compatibility

- Plain strings without `*` or `?`: behavior unchanged (exact match)
- `/regex/` syntax: behavior unchanged
- Strings with `*` or `?`: previously silently wrong (escaped the chars), now correctly treated as globs — this is a bugfix
- Legacy SHA-based artifact lookup: completely unaffected

Refs: https://github.com/getsentry/cli/pull/230

### DIFF
--- a/src/commands/publish.ts
+++ b/src/commands/publish.ts
@@ -510,22 +510,6 @@ export async function publishMain(argv: PublishOptions): Promise<any> {
 
   await printRevisionSummary(artifactProvider, revision);
 
-  // If artifact filters are explicitly configured, ensure we actually found
-  // artifacts. The artifact provider itself validates that all patterns matched,
-  // but this is a belt-and-suspenders check at the publish level to catch any
-  // edge case where zero artifacts slip through without an error.
-  const artifactsConfig = config.artifactProvider?.config?.artifacts;
-  if (artifactsConfig) {
-    const artifacts = await artifactProvider.listArtifactsForRevision(revision);
-    if (artifacts.length === 0) {
-      reportError(
-        'No artifacts found for the revision, but artifact filters are ' +
-          'configured in .craft.yml. Check that your workflow names and ' +
-          'artifact names match the patterns in artifactProvider.config.artifacts.',
-      );
-    }
-  }
-
   await checkRequiredArtifacts(artifactProvider, revision, config.requireNames);
 
   // Find targets


src/artifact_providers/github.ts
```diff
@@ -499,13 +512,77 @@ export class GitHubArtifactProvider extends BaseArtifactProvider {
...
+    filters: NormalizedArtifactFilter[],
+    allRuns: WorkflowRun[],
+    matchingArtifacts: ArtifactItem[],
+  ): void {
+    const errors: string[] = [];
+
+    for (const filter of filters) {
+      // Validate workflow pattern matched at least one run
+      if (filter.workflow) {
+        const matchedRuns = allRuns.filter(run =>
```
BYK: Why not `allRuns.some()`?

BYK: Good call — switched to `.some()` with a negated check in e19fc46.

src/artifact_providers/github.ts
```diff
@@ -557,32 +634,38 @@ export class GitHubArtifactProvider extends BaseArtifactProvider {
...
-          `No artifacts matching filters found for revision "${revision}" (tries: ${MAX_TRIES})`
+          `No artifacts matching filters found for revision "${revision}" (tries: ${MAX_TRIES})`,
         );
       }
 
+      // Validate that ALL configured patterns matched, not just some.
+      // This catches cases like: config says ['craft-binary', 'craft-docs']
+      // but only 'craft-binary' was found. Without this check, the publish
+      // would silently proceed with missing artifacts.
+      this.validateAllPatternsMatched(filters, allRuns, matchingArtifacts);
```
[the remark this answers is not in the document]
BYK: Fixed in dbcaec4. `validateAllPatternsMatched` now returns errors instead of throwing, so the retry loop retries when not all patterns match yet (e.g. API propagation delay). Only on the final attempt does it throw.

src/artifact_providers/github.ts
```diff
@@ -499,13 +512,77 @@ export class GitHubArtifactProvider extends BaseArtifactProvider {
...
+          const availableNames = matchingArtifacts.map(a => a.name).join(', ');
+          const workflowDesc = filter.workflow
+            ? ` (from workflow ${filter.workflow})`
+            : '';
+          errors.push(
+            `Artifact pattern ${artifactPattern}${workflowDesc} did not match any artifacts. ` +
+              `Available artifact names: ${availableNames || '(none)'}`,
+          );
+        }
+      }
```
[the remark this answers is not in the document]
BYK: Fixed in dbcaec4. Artifact validation is now scoped per workflow filter — each filter's artifact patterns are checked only against artifacts from matching workflow runs, using `artifact.workflow_run.id` to correlate. Added a dedicated test for this scenario.

src/artifact_providers/__tests__/github.test.ts
```diff
@@ -866,10 +879,249 @@ describe('GitHub Artifact Provider', () => {
...
+    });
+
+    test('glob pattern in workflow names matches correctly', () => {
+      const result = normalizeArtifactsConfig({
+        'Build*': 'output',
+      });
+      expect(result).toHaveLength(1);
+      expect(result[0].workflow?.test('Build')).toBe(true);
+      expect(result[0].workflow?.test('Build & Test')).toBe(true);
+      expect(result[0].workflow?.test('Lint')).toBe(false);
```
BYK: Why `workflow?` instead of `workflow.test`? We should not have conditionals in tests where possible anyways.

BYK: Good call — changed all 10 occurrences to `workflow!.test()` in 23df07e. Also applied to the pre-existing tests that used the same pattern for consistency.

src/commands/publish.ts
```diff
@@ -509,6 +509,23 @@ export async function publishMain(argv: PublishOptions): Promise<any> {
...
+  if (artifactsConfig) {
+    const artifacts = await artifactProvider.listArtifactsForRevision(revision);
+    if (artifacts.length === 0) {
+      reportError(
+        'No artifacts found for the revision, but artifact filters are ' +
+          'configured in .craft.yml. Check that your workflow names and ' +
+          'artifact names match the patterns in artifactProvider.config.artifacts.',
+      );
+    }
+  }
```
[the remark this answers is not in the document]
BYK: Correct — removed in a51a44e. The provider-level validation in `doListArtifactsWithFilters` already covers this: it either throws (stopping execution at `printRevisionSummary`) or returns non-empty, making this check always false.

PR #796, opened 2026-04-21
fix(security): Move workflow to pull_request and do not persist creds

geoffg-sentry:
No need for pull_request_target in this job, but the downstream callers do need it. Moved the permissions to only the job that needs it.


BYK (approved):
Yes pls

PR #614, opened 2025-07-28
fix: Avoid `global.json` when running all `dotnet` commands

Flash0ver:
### Summary

The [Sentry .NET SDK](https://github.com/getsentry/sentry-dotnet) uses a [global.json](https://learn.microsoft.com/dotnet/core/tools/global-json) file to pin a specific version of the .NET SDK and .NET Workloads to ensure reproducible builds.
When pushing the Sentry NuGet packages and Symbol packages, we can use any version of the .NET SDK,
as long as we ensure that the actual `dotnet build`, `dotnet test`, and creating the packages via `dotnet pack` run with the pinned version of the .NET SDK.

In a previous PR, see #601,
we applied `spawnOptions` to the invocation of `dotnet nuget push` via `spawnProcess`,
with which we set the Working Directory outside of the `sentry-dotnet` repository,
so that the `global.json` does not get picked up / enforced.

However,
we forgot to apply the same `spawnOptions` to the other invocations of `dotnet` commands.

This manifests in failure when invoking `dotnet --version`:
```TEXT
...
[info] [[target/nuget]] .NET Version:
Error:  Process "dotnet" errored with code 145
...
dotnet: 8.0.412 [/usr/share/dotnet/sdk]
dotnet: 9.0.303 [/usr/share/dotnet/sdk]
...
dotnet: Requested SDK version: 9.0.301
dotnet: global.json file: global.json
...
```

<details>
<summary>Publish using Craft</summary>

```TEXT
[info] [[target/nuget]] .NET Version:
Error:  Process "dotnet" errored with code 145

STDOUT: dotnet: 8.0.412 [/usr/share/dotnet/sdk]
dotnet: 9.0.303 [/usr/share/dotnet/sdk]
dotnet: 


STDERR:dotnet: The command could not be loaded, possibly because:
dotnet:   * You intended to execute a .NET application:
dotnet:       The application '--version' does not exist.
dotnet:   * You intended to execute a .NET SDK command:
dotnet:       A compatible .NET SDK was not found.
dotnet: 
dotnet: Requested SDK version: 9.0.301
dotnet: global.json file: /github/workspace/__repo__/global.json
dotnet: 
dotnet: Installed SDKs:
dotnet: 
dotnet: Install the [9.0.301] .NET SDK or update [/github/workspace/__repo__/global.json] to match an installed SDK.
dotnet: 
dotnet: Learn about SDK resolution:
dotnet: https://aka.ms/dotnet/sdk-not-found
dotnet: 

  
[...]

PR #820, opened 2026-05-21
fix(nuget): move global.json aside during `dotnet setversion`

### LINKED ISSUE #819: Regression in nuget automatic version bump breaks dotnet releases
## Summary

The automatic version bumping added in #707 re-introduces the `global.json` pickup problem that #601 and #614 previously eliminated for the `nuget` target.

https://github.com/getsentry/craft/blob/ca38805e30f4111a9694512a2e58554d8e55e9b5/src/targets/nuget.ts#L77-L82

We've deliberately used `DOTNET_SPAWN_OPTIONS = { cwd: '/' }` elsewhere in this file to avoid this until now. I don't think that can work with setversion though, because it operates on files in cwd.

## Prior occurrences

- getsentry/craft#598 / getsentry/craft#601 — first hit on `dotnet nuget push`. Fixed by spawning with `cwd: '/'` so `global.json` isn't picked up.
- getsentry/craft#614 — same issue on `dotnet --version`; the `cwd: '/'` workaround was applied to **all** `dotnet` invocations in `src/targets/nuget.ts`.
- getsentry/craft#707 — added "Automatic version bumping from configured targets". The new `NugetTarget.bumpVersion` spawns `dotnet setversion` with `cwd: rootDir`, re-introducing the very pattern the previous PRs eliminated, on a new code path.

So this re-surfaces whenever the SDK pinned in `global.json` is newer than (or otherwise absent from) the image Craft runs on.

See linked issue in sentry-dotnet repo:
- https://github.com/getsentry/sentry-dotnet/issues/5250

## Brainstorming

1. **Skip `dotnet-setversion` and use the existing XML-edit fallback already implemented in the same method.** It doesn't shell out to `dotnet` at all, so `global.json` becomes irrelevant.
2. Temporarily move `global.json` aside for the duration of the `dotnet setversion` call (can't use `cwd: '/'` here because the tool operates on the cwd).

## Related

Downstream tracking issue:
- https://github.com/getsentry/sentry-dotnet/issues/5250

### LINKED ISSUE #5250: Release pipeline fails: `craft prepare` picks up pinned SDK via global.json
## Symptom

When the SDK pinned in `global.json` is newer than (or otherwise absent from) the image Craft runs on, the release workflow fails during `craft prepare` → automatic version bumping. e.g. [run 26224458396](https://github.com/getsentry/sentry-dotnet/actions/runs/26224458396/job/77167675585) (releasing 6.5.1).

```
[error] Automatic version bump failed for "nuget" target: Process "dotnet" errored with code 155
STDERR: dotnet: A compatible .NET SDK was not found.
        Requested SDK version: 10.0.203
        global.json file: /home/runner/work/sentry-dotnet/sentry-dotnet/global.json
        Installed SDKs: 8.0.127, 8.0.206, 8.0.319, 8.0.421, 9.0.117, 9.0.205, 9.0.314, 10.0.108, 10.0.204, 10.0.300
```

## Root cause

Craft (v2.26.5) [invokes `dotnet setversion <version>`](https://github.com/getsentry/craft/blob/ca38805e30f4111a9694512a2e58554d8e55e9b5/src/targets/nuget.ts#L75-L93) with `cwd: rootDir` (the repo). That picks up our `global.json`, which pins the .NET version (to ensure deterministic builds). 

If the GitHub-hosted runner doesn't carry the specific version of .NET we're using (currently `10.0.203` but that shifts frequently) the dotnet host aborts before `setversion` runs.

## Prior occurrences

- getsentry/craft#598 / getsentry/craft#601 — first hit on `dotnet nuget push`. Fixed by spawning with `cwd: '/'` so `global.json` isn't picked up.
- getsentry/craft#614 — same issue on `dotnet --version`; the `cwd: '/'` workaround was applied to **all** `dotnet` invocations in `src/targets/nuget.ts`.
- getsentry/craft#707 — added "Automatic version bumping from configured targets". The new `NugetTarget.bumpVersion` spawns `dotnet setversion` with `cwd: rootDir`, re-introducing the very pattern the previous PRs eliminated, on a new code path.

## Proposed fix

This really needs to be addressed in `getsentry/craft`, but some options:

1. **Skip `dotnet-setversion` and use the existing XML-edit fallback already implemented in the same method.** It doesn't shell out to `dotnet` at all, so `global.json` becomes irrelevant. This is the cleanest fix.
2. Temporarily move `global.json` aside for the duration of the `dotnet setversion` call (can't use `cwd: '/'` here because the tool operates on the cwd).

## Links

- Failing run: https://github.com/getsentry/sentry-dotnet/actions/runs/26224458396/job/77167675585
- [Craft source introducing the regression](https://github.com/getsentry/craft/blob/ca38805e30f4111a9694512a2e58554d8e55e9b5/src/targets/nuget.ts#L75-L93)
- Tracking issue raised in sentry-craft:
  - https://github.com/getsentry/craft/issues/819

jamescrosswell:
## Summary

Fixes #819. Refs getsentry/sentry-dotnet#5250.

The automatic version bumping added in #707 invokes `dotnet setversion` with `cwd: rootDir`, so the dotnet host reads any `global.json` in the consumer repo. The release runner won't always have that exact SDK installed — `dotnet` then refuses to launch and `craft prepare` aborts.

Other `dotnet` invocations in this target dodge this by spawning with `cwd: '/'` (see #601, #614). That isn't viable here because `dotnet-setversion` operates on files in cwd.

Instead, this PR temporarily renames `global.json` to `global.json.craft-bak` for the duration of the call and restores it in a `finally`. The XML-fallback path is unchanged and unaffected.

## Notes for reviewers

- An alternative workaround would be to skip `dotnet-setversion` altogether and rely on the xml based version update logic that is currently a fallback. Hard to say which workaround is better.
- No existing test file for `nuget.ts` (`src/targets/__tests__/` has no `nuget.test.ts`), so this PR doesn't add one. Happy to add one if you'd like — let me know the preferred shape.

### DIFF
--- a/src/targets/nuget.ts
+++ b/src/targets/nuget.ts
@@ -291,9 +291,9 @@ export class NugetTarget extends BaseTarget {
 
         this.logger.info(
           `Uploading file "${file.filename}" via "dotnet nuget"` +
-          (symbolFile
-            ? `, including symbol file "${symbolFile.filename}"`
-            : ''),
+            (symbolFile
+              ? `, including symbol file "${symbolFile.filename}"`
+              : ''),
         );
         return this.uploadAsset(path);
       }),
--- a/src/targets/nuget.ts
+++ b/src/targets/nuget.ts
@@ -1,4 +1,10 @@
-import { existsSync, readFileSync, readdirSync, writeFileSync } from 'fs';
+import {
+  existsSync,
+  readFileSync,
+  readdirSync,
+  renameSync,
+  writeFileSync,
+} from 'fs';
 import { join } from 'path';
 
 import { TargetConfig, TypedTargetConfig } from '../schemas/project_config';
@@ -78,6 +84,19 @@ export class NugetTarget extends BaseTarget {
     }
 
     if (hasExecutable(NUGET_DOTNET_BIN)) {
+      // `dotnet-setversion` operates on files in cwd, so we must invoke it with
+      // cwd=rootDir. That means the dotnet host will read any global.json in
+      // rootDir and may refuse to launch if the pinned SDK isn't installed on
+      // this runner (consumer repos often pin SDK/workload versions with
+      // rollForward: disable for deterministic builds — see #614, #707, and
+      // getsentry/sentry-dotnet#5250). Move global.json aside for the duration
+      // of the call so the host can pick any installed SDK.
+      const globalJsonPath = join(rootDir, 'global.json');
+      const globalJsonBackup = `${globalJsonPath}.craft-bak`;
+      const globalJsonMoved = existsSync(globalJsonPath);
+      if (globalJsonMoved) {
+        renameSync(globalJsonPath, globalJsonBackup);
+      }
       try {
         const result = await spawnProcess(
           NUGET_DOTNET_BIN,
@@ -99,6 +118,10 @@ export class NugetTarget extends BaseTarget {
         logger.debug(
           'dotnet-setversion not available, falling back to manual edit',
         );
+      } finally {
+        if (globalJsonMoved) {
+          renameSync(globalJsonBackup, globalJsonPath);
+        }
       }
     }
 
--- a/src/targets/nuget.ts
+++ b/src/targets/nuget.ts
@@ -84,13 +84,9 @@ export class NugetTarget extends BaseTarget {
     }
 
     if (hasExecutable(NUGET_DOTNET_BIN)) {
-      // `dotnet-setversion` operates on files in cwd, so we must invoke it with
-      // cwd=rootDir. That means the dotnet host will read any global.json in
-      // rootDir and may refuse to launch if the pinned SDK isn't installed on
-      // this runner (consumer repos often pin SDK/workload versions with
-      // rollForward: disable for deterministic builds — see #614, #707, and
-      // getsentry/sentry-dotnet#5250). Move global.json aside for the duration
-      // of the call so the host can pick any installed SDK.
+      // `dotnet-setversion` operates in cwd and will pickup global.json,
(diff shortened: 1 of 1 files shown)


src/targets/nuget.ts
```diff
@@ -1,4 +1,10 @@
-import { existsSync, readFileSync, readdirSync, writeFileSync } from 'fs';
```
jamescrosswell, line 1: Not related to the fix - this gets made automatically by prettier

jamescrosswell:
The Change log Preview fails, but I suspect that's because I'm making the PR from a fork (I'm a contractor so don't have permission to create a PR on the repo directly).

src/targets/nuget.ts
```diff
@@ -78,6 +84,15 @@ export class NugetTarget extends BaseTarget {
     }
 
     if (hasExecutable(NUGET_DOTNET_BIN)) {
+      // `dotnet-setversion` operates in cwd and will pickup global.json,
+      // which breaks if the pinned SDK isn't installed on the craft runner.
+      // See: https://github.com/getsentry/craft/issues/819
+      const globalJsonPath = join(rootDir, 'global.json');
+      const globalJsonBackup = `${globalJsonPath}.craft-bak`;
```
BYK, line 91: We should probably use a unique name per instance to avoid any potential clashes but since the probabiliy is very low, I'll merge it as it. Would appreciate a follow up to fix this tho.

BYK (approved):
Thanks so much and sorry for the trouble!

src/targets/nuget.ts
jamescrosswell: No worries - follow up PR:
- https://github.com/getsentry/craft/pull/823

PR #823, opened 2026-05-24
fix(nuget): use unique suffix for global.json backup

jamescrosswell:
## Summary

Follow-up to #820, addressing [BYK's review note](https://github.com/getsentry/craft/pull/820#discussion_r3287646962).

The backup path was hardcoded to `global.json.craft-bak`. If two craft invocations ever ran against the same working directory, the second would clobber the first's backup. Probability is very low in normal use (craft typically runs in a release-specific checkout), but a unique suffix costs nothing.

Uses `crypto.randomUUID()` to make the backup name unique per invocation.

### DIFF
--- a/src/targets/nuget.ts
+++ b/src/targets/nuget.ts
@@ -6,6 +6,7 @@ import {
   writeFileSync,
 } from 'fs';
 import { join } from 'path';
+import { randomUUID } from 'crypto';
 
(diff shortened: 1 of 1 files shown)


BYK (approved):
Sorry it took me so long and thanks a lot for the follow up!!!

PR #843, opened 2026-07-15
feat(targets): add cloudflare deploy target

BYK:
## Summary

Adds a new `cloudflare` release target that deploys a release artifact to Cloudflare via the [`wrangler`](https://developers.cloudflare.com/workers/wrangler/) CLI. Part of the getsentry/toolkit monorepo work (#842) — enables deploying the CLI docs website to Cloudflare on release.

Two config-selectable deploy modes:

- **`deployType: pages`** (default) → `wrangler pages deploy <dir> --project-name <name> --branch <productionBranch> ...`. Always targets **production** by passing the configured `productionBranch` (default `main`) as `--branch` (the Cloudflare environment selector, *not* a git branch), plus `--commit-hash`/`--commit-message`/`--commit-dirty` provenance so wrangler doesn't infer git state from the temp dir.
- **`deployType: worker`** → `wrangler deploy` using a `wrangler.toml` in the artifact (run with `cwd` = deploy dir).

## Configuration

| Option | Description |
|--------|-------------|
| `deployType` | `pages` (default) or `worker` |
| `projectName` | Pages project name (required for `pages`) |
| `productionBranch` | Pages production branch, passed as `--branch`. Default `main` |
| `wranglerCliPath` | wrangler binary path (default `wrangler`, or `WRANGLER_BIN`) |
| `workingDir` | Subdir within the artifact to deploy from |

Secrets `CLOUDFLARE_API_TOKEN` and `CLOUDFLARE_ACCOUNT_ID` are passed via **env, never argv**.

## Safety

- **Dry-run:** the deploy is a remote, irreversible operation, so it is explicitly skipped in dry-run — including worktree mode, where `spawnProcess` would otherwise run for real. (Local extraction still happens; only the remote deploy is skipped.)
- **Secret injection:** config values that would expand to `${VAR}` via `spawnProcess`'s env-substitution are rejected, so a config string can't be interpolated into a secret.

## Docker

`wrangler@4.111.0` is installed globally in the runtime stage (x64-only image; `workerd` prebuilt binary is fine) with a `wrangler --version` smoke check, matching how other target toolchains are shipped.

## Other changes

- Refactored ghPages' single-top-level-directory flatten logic into a shared `extractZipArchiveWithFlattening` util reused by both targets (behavior unchanged).
- New docs page `targets/cloudflare.md` + overview table row.

## Testing

- `pnpm test` — full suite green (17 new cloudflare tests, incl. dry-run guard, secret-in-env, both deploy types, config validation).
[...]

Issue #842, opened 2026-07-15
Support Cloudflare deploy target + prefixed multi-product releases (for getsentry/toolkit)


BYK:
Part of the **sentry-cli + sentry-mcp -> `getsentry/toolkit` monorepo merge**.
Full plan: https://gist.github.com/BYK/8c0ce6cec64608f41fe5654b7827228e

## Ask
The merged `toolkit` monorepo needs to deploy the **CLI documentation website to
Cloudflare** as a **release-gated** step (docs must stay in sync with the
published CLI version — same guarantee the current `gh-pages` Craft target
provides). GitHub Pages doesn't scale to multiple sites in one monorepo, so the
CLI website is moving to Cloudflare (where the MCP worker already lives). We want
to drive this from Craft on CLI release.

## Question / request for Craft
- Does Craft have (or would it accept) a **`cloudflare` target** — Pages
  (`wrangler pages deploy`) and/or Worker static assets (`wrangler deploy`)?
- If not, confirm the recommended pattern for a **custom / `command`-style
  target** invoking `wrangler` with `CLOUDFLARE_API_TOKEN` + `CLOUDFLARE_ACCOUNT_ID`,
  gated to release tags, consuming a prebuilt docs artifact.
- Guidance on **prefixed release tags** (`cli@x.y.z` / `mcp@x.y.z`) for two
  independently-versioned products in a single repo — confirm tag-prefix support
  and per-product `.craft.yml` invocation.

## Cross-links
- Master: getsentry/cli (link added once master issue is filed)

BYK:
Master tracking issue: https://github.com/getsentry/cli/issues/1248 (getsentry/cli#1248)

Issue #649, opened 2025-12-04
Add workspaces support for dotnet/nuget


Flash0ver:
Similar to #645.

We currently publish all (currently 16) NuGet packages, which are versioned together, in a single publish step. In some cases, this has caused reentrancy problems, when NuGet.org was temporary unavailable or unresponsive (see #571 and #619).
"Exploding" our package hierarchy (mind dependencies it publish in the correct order) could solve this problem without the current "hack" we have employed for dotnet/nuget.

### TODOs
- Provide the algorithm of gathering all packages to publish (keep the order in mind for partial publish / resume / retry)
  - either as a small script (e.g. PowerShell or file-based app (if available)), max 30 lines
  - or textually to be built into craft
  - @Flash0ver

Issue #394, opened 2022-05-11
Resume publishing a Rust crates workspace


jan-auer:
In #392, we implemented a workaround to skip already published crates. This comes with the downside of hiding errors when running a release that already exists. 

A better solution, as proposed by @BYK, would be to remember the crates that have been published in craft's lockfile, and resume publishing for the rest.

PR #844, opened 2026-07-16
feat: support prefixed tags for monorepo multi-product releases

BYK:
## Summary

Enables several independently-versioned products to live in one repository (e.g. `cli@1.2.3`, `mcp@2.0.0`) without cross-contaminating each other's latest-tag/changelog-base detection or release branches. Part of the getsentry/toolkit monorepo work (#842) — the second of two PRs (after the `cloudflare` target in #843).

The **write** path already honored `github.tagPrefix` (setting `tagPrefix: "cli@"` already produces `cli@1.2.3`). This PR fills the gaps on the **read** paths.

## Changes

- **`getLatestTag(git, tagPrefix)`** (`src/utils/git.ts`) — scopes `git describe --tags --abbrev=0` with `--match '<prefix>*'` so the latest tag is resolved per-product instead of picking whatever is newest across all interleaved tags. Backward compatible (empty prefix → current behavior).
- **Thread the prefix through read paths** — `prepare.ts` auto-version base (`:762`) and changelog/oldVersion base (`:870`) now pass `getGitTagPrefix()`; the `changelog` command (`changelog.ts`) does too, guarded by a config-file check so it still runs standalone.
- **`getGitTagPrefix()`** (`src/config.ts`) — now warns when a *single* config declares multiple `github` targets with *differing* `tagPrefix` values (ambiguous), returning the first. The intended layout is one `.craft.yml` per product.
- **`findReleaseBranches`** (`src/utils/git.ts`) — handles slashed release-branch prefixes (`release/cli` → `release/cli/1.2.3`) so the failed-checkout error hints work with the per-product `releaseBranchPrefix` collision fix. Verified regression-free for single-segment prefixes.

**Already prefix-aware, no code change (now covered by tests):** CalVer scan (`calver.ts`), `versionToTag`, and `getVersion`/`parseVersion` (which correctly extract `1.2.3` from `cli@1.2.3`, `mcp@2.0.0-dev.1`, `sentry-cli@10.20.30`, etc.).

## Monorepo model

One `.craft.yml` per product, each with its own `tagPrefix` + slashed `releaseBranchPrefix`:

```yaml
# CLI
releaseBranchPrefix: release/cli
targets:
  - name: github
    tagPrefix: "cli@"
```
```yaml
# MCP
releaseBranchPrefix: release/mcp
targets:
  - name: github
    tagPrefix: "mcp@"
```

→ tags `cli@1.2.3` / `mcp@1.2.3` on branches `release/cli/1.2.3` / `release/mcp/1.2.3`. No collisions.

## Known limitation (documented)

GitHub tracks a single **repo-wide** "Latest" release; `isLatestRelease` is not prefix-scoped, so the "Latest" badge can move between products. Tags, changelogs, release branches, and version detection remain correctly per-product. Documented in `github.md`.

## Testing

- `pnpm test` — full suite green (1055 passed). New/updated tests: `getLatestTag` with/without prefix + `--match` assertion; slashed-prefix `findReleaseBranches` (incl. the versionless-branch edge); `getVersion`/`parseVersion` on prefixed tags; `getGitTagPrefix` single/same/conflicting/mixed-undefined cases.
- `pnpm lint` / `tsc --noEmit` clean. Docs build clean.

## Review

Adversarial subagent review: **no CRITICAL/MAJOR bugs in changed code**; the `findReleaseBranches` refactor verified regression-free case-by-case. Findings M1 (repo-wide "Latest" limitation) and M3 (mixed-prefix test) were addressed in this PR.

🤖 Generated with [opencode](https://opencode.ai)

### DIFF
--- a/docs/src/content/docs/configuration.md
+++ b/docs/src/content/docs/configuration.md
@@ -176,6 +176,22 @@ releaseBranchPrefix: publish
 
 Full branch name: `{releaseBranchPrefix}/{version}`
 
+The prefix may contain slashes, which is useful for monorepos that release
+several independently-versioned products from one repository. Pairing a slashed
+`releaseBranchPrefix` with a per-product `github.tagPrefix` keeps each product's
+release branches and tags separate:
+
+```yaml
+releaseBranchPrefix: release/cli
+targets:
+  - name: github
+    tagPrefix: "cli@"
+```
+
+This produces branches like `release/cli/1.2.3` and tags like `cli@1.2.3`. See
+the [GitHub target docs](./targets/github/#monorepo-independently-versioned-products)
+for the full monorepo pattern.
+
 ## Changelog Policies
 
 Craft supports `simple` and `auto` changelog management modes.
--- a/docs/src/content/docs/targets/github.md
+++ b/docs/src/content/docs/targets/github.md
@@ -43,6 +43,44 @@ targets:
 
 This is useful for users who want to pin to a major version while automatically receiving updates.
 
+## Monorepo: independently-versioned products
+
+`tagPrefix` lets a single repository host several independently-versioned products by namespacing their git tags — for example `cli@1.2.3` and `mcp@2.0.0`. Craft honors the prefix on both the **write** side (the tag it creates) and the **read** side (latest-tag detection, changelog base, and CalVer scans are all scoped to the prefix), so the products don't cross-contaminate each other's version history.
+
+The intended layout is **one `.craft.yml` per product**, each with a single `github` target declaring its own `tagPrefix` and a matching `releaseBranchPrefix` (so release branches don't collide):
+
+```yaml
+# .craft.yml for the CLI product
+github:
+  owner: getsentry
+  repo: toolkit
+releaseBranchPrefix: release/cli
+targets:
+  - name: github
+    tagPrefix: "cli@"
+```
+
+```yaml
+# .craft.yml for the MCP product
+github:
+  owner: getsentry
+  repo: toolkit
+releaseBranchPrefix: release/mcp
+targets:
+  - name: github
+    tagPrefix: "mcp@"
+```
+
+Releasing `1.2.3` for each product then produces the tags `cli@1.2.3` / `mcp@1.2.3` on release branches `release/cli/1.2.3` / `release/mcp/1.2.3` — no collisions.
+
+:::caution
+Declaring **multiple** `github` targets with **different** `tagPrefix` values in a *single* config is ambiguous: Craft uses the first prefix for read-path operations and logs a warning. Use a separate `.craft.yml` per product instead.
+:::
+
+:::note[Known limitation: the GitHub "Latest" badge is repo-wide]
+GitHub tracks a single "Latest" release **per repository**, not per tag prefix. Craft decides whether to mark a release as latest by comparing its version against the repository's current latest release ([`isLatestRelease`](https://github.com/getsentry/craft/blob/master/src/targets/github.ts)), which is not prefix-scoped. In a monorepo this means the "Latest" badge can move between products (e.g. publishing `cli@9.0.0` may take the badge from `mcp@3.0.0`, and publishing `cli@1.2.3` while `mcp@3.0.0` is latest won't earn the badge). Tags, changelogs, release branches, and version detection remain correctly per-product — only GitHub's single "Latest" pointer is shared.
+:::
+
 ## Preview Releases
 
 If `previewReleases` is `true` (default), releases containing pre-release identifiers like `alpha`, `beta`, `rc`, etc. are marked as pre-releases on GitHub.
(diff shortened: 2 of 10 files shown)


docs/src/content/docs/targets/github.md
```diff
@@ -43,6 +43,44 @@ targets:
...
+releaseBranchPrefix: release/mcp
+targets:
+  - name: github
+    tagPrefix: "mcp@"
+```
+
+Releasing `1.2.3` for each product then produces the tags `cli@1.2.3` / `mcp@1.2.3` on release branches `release/cli/1.2.3` / `release/mcp/1.2.3` — no collisions.
+
+:::caution
+Declaring **multiple** `github` targets with **different** `tagPrefix` values in a *single* config is ambiguous: Craft uses the first prefix for read-path operations and logs a warning. Use a separate `.craft.yml` per product instead.
```
BYK, line 77: Isn't it better if we support this use case explicitly to manage everything from a single top-level file when it makes sense? Otherwise I fear we may have issues with the `publish` repo?

docs/src/content/docs/configuration.md
```diff
@@ -176,6 +176,22 @@ releaseBranchPrefix: publish
...
+The prefix may contain slashes, which is useful for monorepos that release
+several independently-versioned products from one repository. Pairing a slashed
+`releaseBranchPrefix` with a per-product `github.tagPrefix` keeps each product's
+release branches and tags separate:
+
+```yaml
+releaseBranchPrefix: release/cli
+targets:
+  - name: github
+    tagPrefix: "cli@"
```
BYK, line 188: I wonder if we we can find a way to unify these. I'd even go as far as saying we should support "workspaces" and associated patterns. Sounds like a natural next step of this evolution.

src/utils/git.ts
```diff
@@ -236,12 +247,21 @@ export async function findReleaseBranches(
...
-    if (!branchPrefix) {
+    // The prefix may itself contain slashes for monorepo per-product releases
+    // (e.g. "release/cli" → branches like "release/cli/1.2.3"). Compare the
+    // branch's leading segments against the prefix rather than only the first
+    // path segment, so slashed prefixes match exactly and fuzzily.
+    const prefixSegmentCount = prefix.split('/').length;
+    const branchSegments = withoutRemote.split('/');
+    const branchPrefix = branchSegments
+      .slice(0, prefixSegmentCount)
+      .join('/');
```
BYK, line 259: Why not just cut the first segment with `.indexOf()` and `.slice()` to avoid the full split/merge overhead. I'd say this is also more error prone? Slashes should not have any meaning for the prefix and we should treat that as an opaque string.

src/config.ts
```diff
@@ -328,11 +328,33 @@ export async function getGlobalGitHubConfig(
 
 /**
  * Gets git tag prefix from configuration
+ *
+ * Returns the `tagPrefix` of the first `github` target. In a monorepo where
```
BYK, line 332: Why limit this to the first target and not allow indexing at all?

BYK:
Parking this PR in favor of a broader, target-agnostic **workspaces** redesign (design doc landing shortly, tracked under #842).

## Why

The core feedback here — @BYK's "why limit to the first target / allow indexing" (`config.ts`) and "support managing everything from a single top-level file / workspaces" (`github.md`, `configuration.md`) — points at a bigger design than this PR's per-`.craft.yml`-per-product approach. After digging into the release + `publish_repo`/publish-issue pipeline, the conclusion is:

- The current model disambiguates products only by **version + subdirectory `path`**. `path` is a proxy for physical layout, not release identity — it breaks down for a single-file multi-product repo (issue-title collision at `action.yml`, shared publish-state key at `publishState.ts`, whole-config `craft targets` checklist).
- A first-class, **target-agnostic** top-level `workspaces:` concept + an explicit `--workspace` selector is the better long-term primitive. It also addresses #649 (dotnet/nuget) and #394 (crates resume).

## Plan

The design will be delivered as a sequence of small PRs off `master`:

- **PR A** — salvage the non-controversial fixes from this PR (they're correct regardless of the redesign):
  1. Wrap `getGitTagPrefix()` in a try/catch in `changelog.ts` so an invalid `.craft.yml` doesn't abort a standalone `craft changelog` run (Bugbot Medium).
  2. Treat the release-branch prefix as an **opaque string** in `findReleaseBranches` (cut at the last `/`), per @BYK's review + Bugbot Low.
  3. Keep the prefix-aware read paths (`getLatestTag --match`, tests, docs).
- **PR B–E** — `workspaces:` schema + resolver + selector → prepare/publish threading → action layer + (additive) `getsentry/publish` accept-side → docs.

Closing as parked (converting to draft) so the review threads aren't lost. The salvaged fixes will land in PR A.

BYK:
Closing as superseded. PR #847 already merged the useful prefix-aware tag work; PR #848 now provides the top-level release-workspaces model. The remaining #844 design/docs conflict with that model, and its divergent fallback/branch parsing paths are inferior to the fixes already on master.

PR #858, opened 2026-07-28
fix(crates): Ignore dev dependencies when ordering packages

### LINKED ISSUE #856: Crates target incorrectly detects cycles through versioned dev-dependencies
The Crates target reports `Circular dependency detected!` for valid Cargo workspaces that contain a cycle only through `dev-dependencies`.

For example, a workspace can contain `sentry -> sentry-core` as a normal dependency and `sentry-core -> sentry` as a dev-dependency used for doctests. Cargo permits this because dev-dependencies are not required to publish the crate. However, `CratesTarget.getPublishOrder` currently treats a dev-dependency as a workspace publish-order edge unless its requirement is `*` or the `noDevDeps` option is enabled. With workspace-inherited dependencies that include a version, Cargo metadata reports a requirement such as `^0.49.0`, so the path-only exception does not apply and Craft reports a false cycle.

This affects the default Crates target configuration (`noDevDeps: false`) and prevents publishing otherwise valid workspaces. The `url.parse()` deprecation warning in the output is unrelated.

<details>
<summary>Implementation notes</summary>

In [`src/targets/crates.ts`](https://github.com/getsentry/craft/blob/master/src/targets/crates.ts), `getPublishOrder` currently uses:

```ts
if (
  dep.kind === 'dev' &&
  (dep.req === '*' || this.cratesConfig.noDevDeps)
) {
  return false;
}
```

When calculating the publication order, ignore every dependency whose `kind === 'dev'`, regardless of `req` or `noDevDeps`:

```ts
if (dep.kind === 'dev') {
  return false;
}
```

`noDevDeps` controls the publishing command (`cargo hack publish --no-dev-deps`) and should not determine whether dev-dependencies participate in the publish-order graph. The existing `req === '*'` special case was introduced for path-only dev-dependencies, but versioned dev-dependencies are equally irrelevant to publication ordering.

Add a regression test in [`src/targets/__tests__/crates.test.ts`](https://github.com/getsentry/craft/blob/master/src/targets/__tests__/crates.test.ts) with `noDevDeps: false` (or the default configuration). Create two packages where `p1` has a normal dependency on `p2`, while `p2` has a versioned dev-dependency on `p1`, for example `kind: 'dev'` and `req: '^1.0.0'`. Assert that `getPublishOrder` returns `[p2, p1]` rather than throwing. Retain coverage for genuine cycles formed by normal dependencies.

The relevant metadata comes from `cargo metadata --no-deps --format-version=1`; Cargo distinguishes normal dependencies (`kind: null`) from dev-dependencies (`kind: "dev"`).
</details>

szokeasaurusrex:
Cargo dev-dependencies are not required to publish a crate, but the publish-order graph previously treated versioned dev-dependencies as workspace edges when `noDevDeps` was false. A valid graph such as sentry -> sentry-core with the reverse edge used only for doctests was therefore reported as circular. This failure [can be observed here](https://github.com/getsentry/publish/actions/runs/30360035005/job/90277265724).

Ignore all dev-dependencies while calculating publish order, independently of `noDevDeps`. Keep `noDevDeps` responsible only for selecting cargo-hack publish behavior, and add regression coverage for the default configuration with a versioned dev-dependency.

Closes #856

### DIFF
--- a/src/targets/__tests__/crates.test.ts
+++ b/src/targets/__tests__/crates.test.ts
@@ -76,4 +76,27 @@ describe('getPublishOrder', () => {
     const sortedPackages = [p2, p1];
     expect(target.getPublishOrder(packages)).toEqual(sortedPackages);
   });
+
+  test('excludes versioned dev dependencies when noDevDeps is false', () => {
+    const targetWithoutNoDevDeps = new CratesTarget(
+      {
+        name: 'crates',
+        noDevDeps: false,
+      },
+      new NoneArtifactProvider(),
+      { owner: 'getsentry', repo: 'craft' },
+    );
+    const packages = ['p1', 'p2'].map(cratePackageFactory);
+    const [p1, p2] = packages;
+
+    p1.dependencies = [cratePackageToDependency(p2)];
+    p2.dependencies = [
+      makeDev({
+        ...cratePackageToDependency(p1),
+        req: '^1.0.0',
+      }),
+    ];
+
+    expect(targetWithoutNoDevDeps.getPublishOrder(packages)).toEqual([p2, p1]);
+  });
 });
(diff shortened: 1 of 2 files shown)


Dav1dde (approved):
Reasoning makes sense to me 👍

Issue #860, opened 2026-07-29
(feat)(aws-lambda-layer): Add CompatibleArchitectures and region failure visibility


pabloDeputter:
### Problem Statement

1. `.craft.yml` cannot declare `compatibleArchitectures`, so layers never set AWS `CompatibleArchitectures` and the release registry never gets `compatible_architectures` for docs.
2. `publishToAllRegions` ignores per-region currently with a warning resulting in regions having mismatched versions.

### Solution BrainStorm

1. Add optional `compatibleArchitectures` to the aws-lambda-layer target.
2. Keep partial publishes non-blocking; report failures via logger and capture with Sentry.

### Related
- [sentry-python#6908](https://github.com/getsentry/sentry-python/issues/6908)
- [sentry-python#5231](https://github.com/getsentry/sentry-python/issues/5231)
- [sentry-docs#18858](https://github.com/getsentry/sentry-docs/issues/18858)

pabloDeputter:
For [#5231](https://github.com/getsentry/sentry-python/issues/5231), the problem is:
- Region A fails => doesn't get added to the release registry since it was dropped.
- when the next SDK is released, region A may succeed this time so the latest layer now supports the newest SDK version. However the earlier SDK version is never backfilled.

PR #861, opened 2026-07-29
feat(aws-lambda-layer): Add CompatibleArchitectures and per-region failure improvements

pabloDeputter:
- Introduce `compatibleArchitectures` parameter allowing specification of supported instruction set architectures (`x86_64`, `arm64`).
- Updated tests to cover scenarios involving new parameter.
- Make per-region publishing failure more prominent; perform 2 additional requests after initial one fails

Refs: #860

### DIFF
--- a/docs/src/content/docs/targets/aws-lambda-layer.md
+++ b/docs/src/content/docs/targets/aws-lambda-layer.md
@@ -11,6 +11,7 @@ Creates a new public Lambda layer in each available AWS region and updates the S
 |--------|-------------|
 | `layerName` | Name of the Lambda layer. Supports template variables (see below) |
 | `compatibleRuntimes` | List of runtime configurations |
+| `compatibleArchitectures` | Optional list of instruction set architectures (`x86_64`, `arm64`) |
 | `license` | Layer license |
 | `linkPrereleases` | Publish layers for preview/pre-release versions. Default: `false` |
 | `includeNames` | Must filter to exactly one artifact |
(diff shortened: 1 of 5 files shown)


docs/src/content/docs/targets/aws-lambda-layer.md
```diff
@@ -65,6 +69,9 @@ targets:
         versions:
           - nodejs10.x
           - nodejs12.x
```
BYK: Lol, node 10
```suggestion
         - nodejs22.x
         - nodejs24.x
```

src/targets/awsLambdaLayer.ts
```diff
@@ -351,6 +354,10 @@ export class AwsLambdaLayerTarget extends BaseTarget {
           account_number: getAccountFromArn(publishedLayers[0].arn),
           layer_name: resolvedLayerName,
           regions: regionsVersions,
+          compatible_runtimes: runtime.versions,
+          ...(compatibleArchitectures?.length
+            ? { compatible_architectures: compatibleArchitectures }
+            : {}),
```
BYK: This is hurting my brain and eyes, how about:

```ts
if (compatibleArchitectures?.length) {
  runtimeData.compatible_architectures = compatibleArchitectures;
}
```

Yes `runtimeData` is defined as `const` but that doesn't stop us from mutating the object itself. Ugly, but less ugly.

BYK (approved):

